PR #38, opened 2020-05-25
Flag for GTFS & GTFS-RT: Shorten Stop IDs

felixguendling:
In order to match the EVA-numbers from the SBB HRD dataset, only the GTFS trip id's fist part (until the first ":") can be used. Thus, the stops in the SBB GTFS dataset are joined together: all tracks of one stop are one station.

This behaviour is now optional since the DELFI dataset has another logic to use the ":" separator. Here, many stops in Germany start with "de:".


felixguendling:
Not a nice solution.

PR #124, opened 2021-05-18
Real-time updates for expanded trips

pablohoch:
`schedule.expanded_trips_` is now a `dynamic_fws_multimap` which is updated by the rt module.

Changes to existing code:

- `dynamic_fws_multimap` has a slightly different API than the previously used `fws_multimap`
- Buckets (routes) in `expanded_trips_` can now be empty
- `expanded_trips_.data_` can now contain `nullptr` trips

The csa, raptor and tripbased modules have been updated to support these changes, but do not yet support real-time updates, i.e. they should work as before as long as no real-time updates are applied.

For testing purposes, the following operations have been added:

- `/csa/update_timetable`
- `/raptor/update_timetable`
- `/tripbased/update_timetable`

They re-calculate the timetable for the specified module from scratch (which is **slow**, especially for tripbased) and can be called after real-time updates have been applied. As before, routing algorithms other than the routing module **may not be used** unless the global schedule and their module specific timetable are in sync (which is now the case after calling `update_timetable`).

Example for tripbased (this will apply real-time updates until the given `ris.init_time`, then re-create the tripbased timetable and run the routing queries using the updated real-time schedule):
```sh
motis \
  --mode batch \
  --modules ris rt tripbased \
  --init /tripbased/update_timetable \
  --ris.init_time=2022-06-02T07:00+0200 \
  --batch_input_file  ... \
  --batch_output_file ...
```

Other changes:

- Launcher changes:
  - `--init /operation` can now be used in all modes (the operation will be executed at the start)
  - Passing `--init` no longer implies `--mode init`. If init mode is desired, it must now be specified explicitly (`--mode init --init /operation`).
- csa: timetable initialization moved to import step
- tripbased: rest of initialization moved to import step
- rt: delays are now propagated along through edges
- `for_each_departure`/`for_each_arrival` now include events reachable via through edges
- `check_journey` (used in `print` and `intermodal_compare` commands) now accepts journeys where the scheduled departure time at a stop is before the scheduled arrival time at that stop as valid, as long as the current departure time isn't before the current arrival time
- fix cista serialization of `dynamic_fws_multimap` (previously unused)
- `/rt/update`: Delay updates (`RtDelayUpdate`) for rule service trips are now sent for all merged trips
- New option `dataset.debug_broken_trips`: Set to 1 to enable debug output for broken trips (in $data/log/schedule.txt - be sure to turn off dataset.cache_graph)
- `loader_options.expand_trips_` removed (now always enabled, but was already set to true in all cases anyway)

Expanded Trips excalidraw


modules/ris/src/ris.cc
```diff
@@ -259,6 +262,7 @@ struct ris::impl {
       offsets_.clear();
 
       ctx::await_all(motis_publish(msg));
+      in_flush_ = false;
```
felixguendling, line 265: skip_flush_?

modules/rt/include/motis/rt/reroute.h
```diff
@@ -303,8 +304,28 @@ inline void update_delay_infos(
   }
 }
 
+inline void update_expanded_trip(
+    schedule& sched, trip* trp,
+    mcd::vector<trip::route_edge> const& trip_edges) {
+  if (!trp->edges_->empty()) {
```
felixguendling, line 310: if (trp->edges_->empty()) {
  return;
}

modules/rt/include/motis/rt/reroute.h
```diff
@@ -303,8 +304,28 @@ inline void update_delay_infos(
   }
 }
 
+inline void update_expanded_trip(
+    schedule& sched, trip* trp,
+    mcd::vector<trip::route_edge> const& trip_edges) {
```
felixguendling, line 309:
```suggestion
    mcd::vector<trip::route_edge> const& new_trip_edges) {
```

felixguendling: bzw. nur die route übergeben?

modules/rt/include/motis/rt/reroute.h
```diff
@@ -303,8 +304,28 @@ inline void update_delay_infos(
   }
 }
 
+inline void update_expanded_trip(
```
felixguendling, line 307: remove_trip_from_old_expanded_route_add_to_new_expanded_route?! :D

modules/rt/include/motis/rt/reroute.h
```diff
@@ -323,8 +326,47 @@ inline void update_delay_infos(
...
+        if (exp_route.empty()) {
+          utl::erase(sched.route_to_expanded_routes_.at(old_route_id),
+                     old_exp_route_id);
+        }
+        break;
+      }
+    }
+  }
+  if (new_route_id.has_value()) {
+    auto new_exp_route = sched.expanded_trips_.emplace_back();
```
felixguendling, line 354: extract function (cf. modules/rt/include/motis/rt/additional_service_builder.h:253-159

PR #130, opened 2021-06-01
Add CMake target to build the Elm Web UI

pablohoch:
Download Elm binaries using CMake (instead of GitHub Actions) and add a new target `motis-web-ui` that builds the web ui using the downloaded Elm tools.

### DIFF
--- a/.github/workflows/unix.yml
+++ b/.github/workflows/unix.yml
@@ -47,6 +47,7 @@ jobs:
             os: macos-latest
             artifact: macos
             avx: On
+            webui: On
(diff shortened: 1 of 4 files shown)


felixguendling:
maybe make it optional?

`-DMOTIS_WITH_WEBUI=On`

pablohoch:
> maybe make it optional?
> 
> `-DMOTIS_WITH_WEBUI=On`

Like this (default=Off)?

felixguendling:
Yes

PR #147, opened 2021-08-17
Chat

felixguendling:


base/launcher/src/web_server.cc
```diff
@@ -24,6 +25,108 @@ using namespace motis::module;
...
+          err = build_reply(req_id, nullptr, e.code());
+        } catch (std::exception const& e) {
+          err = make_unknown_error_msg(e.what());
+        } catch (...) {
+          err = make_unknown_error_msg("unknown");
+        }
+        err->get()->mutate_id(req_id);
+
+        if (auto const lock = session_.lock(); lock) {
+          lock->send(err->to_json(), type,
```
pablohoch, line 93:
```suggestion
          lock->send(encode_msg(err, type == net::ws_msg_type::BINARY), type,
```

(Or using `binary_` like in `send` (but `type` should be consistent))

base/launcher/src/web_server.cc
```diff
@@ -24,6 +25,108 @@ using namespace motis::module;
...
+    if (lock) {
+      lock->on_close([s = shared_from_this(), cb = std::move(cb)]() { cb(); });
+    }
+  }
+
+  void closed() {
+    if (on_close_) {
+      on_close_();
+    }
+  }
```
pablohoch, line 111: Is this called anywhere?
Whats's the difference between `on_close_` and `set_on_close_cb`?

felixguendling:
```
#include "motis/chat/chat.h"

#include "motis/core/common/logging.h"

namespace motis::chat {

chat::~chat() {
  // This is important!
  // Otherwise, the client callbacks access module member variables
  // _after_ module destruction.
  for (auto const& c : clients_) {
    if (auto const lock = c.lock(); lock) {
      lock->set_on_close_cb({});
      lock->set_on_msg_cb({});
    }
  }
}

void chat::init(motis::module::registry& reg) {
  reg.client_handlers_.emplace(
      "/chat", [&](motis::module::client_hdl const& c) {
        if (auto const lock = c.lock(); lock) {
          lock->set_on_close_cb([this, c]() {
            auto const lock = c.lock();
            auto const it =
                std::find_if(begin(clients_), end(clients_),
                             [&lock](auto&& o) { return o.lock() == lock; });
            if (it != end(clients_)) {
              LOG(motis::logging::info) << " removing client";
              clients_.erase(it);
            }
          });
          lock->set_on_msg_cb([this, c](motis::module::msg_ptr const& msg) {
            auto msgs_sent = 0U;

            auto const lock = c.lock();
            for (auto const& o : clients_) {
              if (auto const o_lock = o.lock(); o_lock != lock) {
                o_lock->send(motis::module::make_success_msg(
                    msg->get()->destination()->target()->str()));
                ++msgs_sent;
              }
            }

            LOG(motis::logging::info)
                << " broadcasted to " << msgs_sent << " / " << clients_.size();
          });
          clients_.emplace_back(c);
          LOG(motis::logging::error) << " client added";
        } else {
          LOG(motis::logging::error) << " client obsolete";
        }
      });
}

}  // namespace motis::chat
```

base/launcher/src/web_server.cc
felixguendling: Good catch!

Issue #181, opened 2021-12-14
initialization error: schedule not loaded - when trying to run ./motis


wkulesza:
Hi,
great project, I love the demo.
Tried recreating all steps mentioned here https://motis-project.de/docs/install
using Switzerland data.
I grabbed the code and extracted (directory structure same as suggested)
Created data folder
Within data folder, I created hrd folder, where i added 4_tu_ch_oev_sammlung_ch__2021_20211201_203455.zip file ,
within data folder, there's also switzerland-latest.osm.pbf file and log folder, whre txt empty files have been created.

When tring to run motis from root folder i get:

```
root@vpsxxxxx:/home/user/motis# ./motis
Used Options:

  server.host: 0.0.0.0
  server.port: 8080
  server.cert_path: ::dev::
  server.priv_key_path: ::dev::
  server.dh_path: ::dev::
  server.api_key:
  server.log_path:
  server.static_path:


  import.paths:
  import.data_dir: data
  import.require_successful: 1


  dataset.path: rohdaten
  dataset.prefix:
  dataset.graph_path: default
  dataset.write_serialized: 1
  dataset.write_graph: 0
  dataset.read_graph: 0
  dataset.read_graph_mmap: 0
  dataset.cache_graph: 0
  dataset.apply_rules: 1
  dataset.adjust_footpaths: 1
  dataset.expand_footpaths: 1
  dataset.begin: TODAY
  dataset.num_days: 2
  dataset.planned_transfer_delta: 30
  dataset.wzr_classes_path:
  dataset.wzr_matrix_path:
  dataset.no_local_transport: 0


  modules: address, cc, csa, guesser, intermodal, lookup, osrm, parking, path, paxforecast, paxmon, ppr, railviz, revise, ris, routing, rt, tiles, tripbased
  exclude_modules:


  remotes:


  mode: server
  batch_input_file: queries.txt
  batch_output_file: responses.txt
  init:
  num_threads: 2
  direct: 0


  csa.bridge: 0
  csa.expand_footpaths: 0


  intermodal.router: routing
  intermodal.revise: 0


  osrm.profiles:


  parking.db_max_size: 549755813888
  parking.max_walk_duration: 10
  parking.import.edge_rtree_max_size: 3221225472
  parking.import.area_rtree_max_size: 1073741824
  parking.import.lock_rtrees: 0


  path.use_cache:
  path.max_size: 34359738368


  paxforecast.forecast_results:
  paxforecast.behavior_stats:
[...]

PR #182, opened 2021-12-16
Improve GTFS-RT Support

felixguendling:


derhuerst:
I'm curious: Specifically, how does this PR improve MOTIS' GTFS-RT support?

felixguendling:
Thank you for your interest.

We implemented the MOTIS GTFS-RT code with focus on the Switzerland GTFS from SBB. When trying the GTFS-RT stream from Poznan (mentioned in #181), I noticed that we do not cover some parts of the GTFS-RT specification (specifically how trips and stops can be identified). So these changes should support the Poznan GTFS-RT data - and therefore all GTFS-RT feeds that work similarly.

The GTFS-RT functionality in MOTIS is rather new and may be a little bit rough around the edges. We are trying hard to improve support - at least for all features that are used in practice.

After that works, I'm also planning a major refactoring of the code handling GTFS-RT to improve compactness and maintainability.

felixguendling:
Seems like delays seem to be working now for this feed!! :rocket: :grinning: 

@wkulesza thank you for providing the feed URL!

Screenshot from 2021-12-18 12-14-11
Screenshot from 2021-12-18 12-13-53


However, from the GTFS-static documentation about `stop_sequence` in `stops.txt`

> The values must increase along the trip but do not need to be consecutive.

For the GTFS-RT-feed to work properly, these values have to be consecutive and start with zero. This is the case for most feeds. We're planning to add support for non-zero-based and non-consecutive `stop_sequence` numbers nonetheless.

PR #189, opened 2022-01-04
Full Year Timetable

felixguendling:


derhuerst:
I'm curious, why is this closed?

felixguendling:
TL;DR: The topic full-year-timetable didn't get dropped, just moved to https://github.com/motis-project/nigiri to implement a RAPTOR-based routing from scratch to fulfill all new requirements we have from different projects - reusable as library for scripting languages, later to be also used by MOTIS to replace our core data model + routing algorithms.

----

Long version with some reasoning:

We have a lot of new requirements from different projects we're working on. Some of these requirements are incompatible with the current algorithm and core data model in MOTIS, the time-dependent graph. This requires a graph-based algorithm (Dijkstra / A* / Multi Criteria Dijkstra (MCD) [0]). Since we use some speed-up techniques like goal-direction and domination by early results, the MCD is quite fast. In the `modules` directory, we also have algorithms like the  RAPTOR algorithm [1] Connection Scanning Algorithm (CSA) [2], or TripBased [3] implemented which are faster in most cases.

Downsides of the current approach:

  - Those newer and faster algorithms are currently "second class citizens" because they retrieve their data from the core data model which is still the time-dependent graph from [0]. Basically, this means you cannot use MOTIS with *only* one of those new data models (RAPTOR, CSA - which would be more lightweight) - you always need the "heavy" graph at the core.
  - For real-time updates (like delays, re-routings, etc.), the time-dependent graph gets updated first and then updates are distributed to the other modules via the message system. This comes with additional serialization overhead.
  - The time-dependent graph as we implemented it now is not fast to copy. We can make a copy of the whole real-time timetable in ~2-3 seconds (for the Germany timetable of 2 days).

Additionally, we have new requirements from different projects:

  - Efficient support for full year timetable using bitfields to represent traffic days
  - Copying a timetable in the matter of milliseconds instead of seconds (200-300ms - 10x faster than before)
  - Faster routing
  - Routing from one start to many destinations (1 to N)
  - Computation of accessible routes (from door to door - this includes transfers)

These requirements would basically require a rewrite of the core of MOTIS. Also since the core data structures and algorithms were implemented (~2012), we learned a lot. Additionally, the RAPTOR algorithm has certain properties that make it best-suited for the listed requirements:

  - It should be possible to adapt the RAPTOR to use bitfields for the traffic days.
  - RAPTOR (and also CSA) use indices instead of pointers to reference data. Those indices can be copied very quickly using `memcpy` whereas copying the current implementation of the time-dependent graph in MOTIS requires recomputing all pointers (and there are a lot!). Of course the graph could also be changed to use indices. But still I would guess that the RAPTOR representation of the data is more compact in memory.
  - The RAPTOR is proven to be faster than the MCD for most cases and even allows for parallelization. We also have a GPU-based variant that is one order of magnitude faster than a comparable CPU version.
  - CSA and RAPTOR algorithm (and also the TripBased algorithm with some modifications) inherently search from one start to *all* other stops in the timetable. Our MCD in MOTIS basically requires the goal-direction to be fast which excludes the MCD as a solution that could cover all requirements.
  - CSA and RAPTOR algorithms (and also the TripBased algorithm with some modifications) inherently search from one start to *all* other stops in the timetable. Our MCD in MOTIS basically requires the goal-direction to be fast which excludes it as a solution that could cover all requirements. It's also possible to speedup the RAPTOR algorithm using goal-direction (as shown in the paper). However, it is sufficiently fast even without (which is important for the 1:N case).
  - MCD and RAPTOR could both be adapted to support the computation of accessible routes. However, the data is expected to become bigger and more complex (adding a detailed indoor-routing graph). Therefore, having a faster algorithm as a baseline would help here too.

So all in all, we decided to replace the MOTIS core with the RAPTOR data model and use the RAPTOR algorithm as our new core routing algorithm. We currently see a lot of advantages with no downside to the time-dependent graph + MCD combination what we have now.

One goal (certainly not the main goal, but .. why not?) is to make the routing algorithm modular and reusable as a stand-alone library outside MOTIS (e.g. in scripting languages like R, Julia, Python, JS/TS, Lua, Ruby, etc.), so I'm currently implementing it in a separate repository: https://github.com/motis-project/nigiri

Currently, there's not much to see there, since I'm still working on the part to fill the data model with data from the HAFAS Rohdaten timetable. I'm using this opportunity to get writ of the `schedule.raw` Flatbuffer timetable format which required us to create our own Flatbuffers variant with 64bit offsets (which became a bit messy and hard to update). Skipping this step should be more memory efficient. At the moment, MOTIS requires ~10GB extra memory during the initialization just to load the fbs file. So basically, I'm trying to also apply some improvements from "lessons learned".

----

Bibliography:
  - [0] *Disser, Yann, Matthias Müller–Hannemann, and Mathias Schnee. "Multi-criteria shortest paths in time-dependent train networks." International Workshop on Experimental and Efficient Algorithms. Springer, Berlin, Heidelberg, 2008.* [PDF](https://www2.mathematik.tu-darmstadt.de/~disser/pdfs/DisserMullerHannemannSchnee08.pdf)
  - [1] *Delling, Daniel, Thomas Pajor, and Renato F. Werneck. "Round-Based Public Transit Routing." Society for Industrial and Applied Mathematics. Proceedings of the Workshop on Algorithm Engineering and Experiments (ALENEX). Society for Industrial and Applied Mathematics, 2012.* [PDF](https://i11www.iti.uni-karlsruhe.de/extra/publications/dpw-rbptr-12.pdf)
  - [2] *Dibbelt, Julian, et al. "Intriguingly simple and fast transit routing." International Symposium on Experimental Algorithms. Springer, Berlin, Heidelberg, 2013.* [PDF](https://i11www.iti.kit.edu/extra/publications/dpsw-isftr-13.pdf)
  - [3] *Witt, Sascha. "Trip-based public transit routing." Algorithms-ESA 2015. Springer, Berlin, Heidelberg, 2015. 1025-1036.* [PDF](https://arxiv.org/pdf/1504.07149)

PR #195, opened 2022-01-19
Isochrone

felixguendling:


lucargir:
Is there any plan to have this integrated into MOTIS or not? I'd find it quite useful to use MOTIS with isochrone maps :D

felixguendling:
I agree, it would be nice to have this functionality in MOTIS. The show effect is good and it's helpful for some use cases.
So there are definitely plans to get this into MOTIS. However, there's no concrete schedule or road map.

1Maxnet1:
> I agree, it would be nice to have this functionality in MOTIS. The show effect is good and it's helpful for some use cases. So there are definitely plans to get this into MOTIS. However, there's no concrete schedule or road map. Currently, there's nobody actively working on this (no capacity).

Would it be possible to help with getting this into MOTIS? How would the process look like? I would assume something like:
1) Try it out
2) Rebase it onto the current master
3) Iron issues out
4) Create a new PR with the rebased state.

felixguendling:
That would be possible. But since it depends

  - on the old core (schedule data structure), which is deprecated
  - on the MOTIS web user interface, which is deprecated

it would only be a short-term solution.

Maybe some shaders could be reused.

kalon33:
Is this still planned? That would be really useful, notably to determine where to look for a home around your work depending on transit time, or finding an optimum between multiple work places. Thanks!

felixguendling:
It might happen in the future (no concrete timeline yet). Although not based on this PR as it is based on the old core, not nigiri.

dalbani:
A year or so later, I'm curious if isochrones are still on your radar? Thanks!

felixguendling:
Yep, it's still something we want but don't have capacity for.

PR #204, opened 2022-02-11
database_query_test: disable optimizations for msvc

pablohoch:
Workaround for failed tests when using Visual Studio 2022


pablohoch:
See #205

PR #224, opened 2022-03-09
RSL SMD Debug

pablohoch:
Additional debug output for testing


felixguendling:
not suitable for master?

pablohoch:
The changes are included in #220 (except for the "remove service class filter" which is now obsolete)

PR #225, opened 2022-03-13
add docker build

felixguendling:


Dockerfile.build
```diff
@@ -0,0 +1,55 @@
...
+ENV ASAN_SYMBOLIZER_PATH="/usr/lib/llvm-14/bin/llvm-symbolizer"
+ENV ASAN_OPTIONS="alloc_dealloc_mismatch=0"
+ENV UBSAN_OPTIONS="halt_on_error=1:abort_on_error=1"
+ENV DEBIAN_FRONTEND="noninteractive"
+ENV BUILDCACHE_COMPRESS="true"
+ENV BUILDCACHE_DIRECT_MODE="true"
+ENV BUILDCACHE_ACCURACY="SLOPPY"
+ENV PATH="/opt:/opt/cmake-3.23.0-rc3-linux-x86_64/bin:/opt/buildcache/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN apt-get update && \
```
derhuerst, line 12: As [the Docker build works layer-based](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache), concatenating shell commands via `&&` into run `RUN` directive makes sense, but only to a certain degree.

For example, if you change the CMake download URL (or any other) below, the whole command will be re-run (and then cached), even though the `apt-get install …` command is still exactly the same.

So when optimising Dockerfiles for fast builds, it makes sense to think about which commands change how often, moving the rarely changing command up and the often-changing commands down.

felixguendling: Thank you for your input! I tried to improve it. It's now located in [this repo](https://github.com/motis-project/docker-cpp-build). This is to be able to use the docker image as GitHub Actions build image as well as to be able to use it to build our other binaries like [pkg](https://github.com/motis-project/pkg), [ppr](https://github.com/motis-project/ppr), [tiles-server](https://github.com/motis-project/tiles), etc.

PR #250, opened 2022-05-25
GBFS Multi Provider Map

felixguendling:


felixguendling:
Added: color coding for number of available vehicles at stations.

Screenshot 2022-05-29 at 23 20 04

PR #259, opened 2022-07-18
Add french translation

kalon33:
Add french translation.


felixguendling:
Thank you very much for translating the UI! :smiley:

PR #274, opened 2022-08-26
start nigiri integration (wip)

felixguendling:
- Create nigiri module to wrap functionality from https://github.com/motis-project/nigiri
- Meta stations - remove code that made meta stations recursive (station A, B, C; A->B is meta station, B->C is meta station but A->C not, then the previous version of the code would create A->C, now we stick to the data)
- Meta stations - don't create meta station relationships for stations connected by footpath (same as above: stick to what the data says)
- CI: switch from GCC-10 to GCC-11 (to support C++20 and C++23 better, required by https://github.com/motis-project/nigiri)
- `--nigiri.no_cache true` is required to not write files; files contain padding bytes of structs that are filled with random data which triggers valgrind/asan - so just don't do it in tests. For production, this is fine.
- change `hashFiles('**/pnpm-lock.yaml')` to `hashFiles('ui/rsl/pnpm-lock.yaml')` to make CI build faster or Mac OS runners
- important for new cista version in combination with mimalloc: `CISTA_USE_MIMALLOC` CMake flag
- MOTIS API: remove category information:
  - `family` attribute is irrelevant, was just the internal ID of the category
  - `train_nr` is was specific to HAFAS Rohdaten, drop it
  - actually, only the `clasz` and `name` (which contains category + train number/line info) of a transport is communicated now via API
  - for more information, the `TripId` can be used, which still contains the train number, etc. (if available, otherwise 0)
- split validate_graph check for incoming edges to prevent MSVC internal compiler error, can be changed back if the issue is resolved in MSVC
- `wstring_convert` is deprecated, replace usage in `print_journey` and `print_trip` with libfmt
- print more information in `print_journey`: provider, direction, line, clasz
- batch mode: print which files cannot be found
- disable some modules by default, can still be activated by explicitly specifying them in `--modules` configuration parameter
- init operation: allow JSON files to be used as input for init operations, detected if the `init_op` starts withs `.` (like `./my_request_file.json`)
- Intermodal module takes router argument via query, which overrides the `.router` commandline flag, allowing the intermodal query generated to generate queries for different routers to compare the results
- Intermodal Comparator: add query file command line paramater, write failed queries to `fail` folder
- Intermodal Generator: can generate any query type now, with any start/dest mumo type (foot, ppr, bike, car, etc.) with any range
- Intermodal Test (`intermodal_itest.cc`): test for every router (`/routing`, `/tripbased`, `/nigiri`)
- Routing module, nigiri: compute `fastest_direct` connection that uses only `mumo` edges (start edge + dest edge) or uses `mumo` edges (start + dest edges) AND foot edges from the timetable (`START` - `mumo` - `STATION` - `mumo - `DEST`) or (`START` - `mumo` - `STATION` - `footpath` - `STATION` - `mumo` - `DEST`): routing module changes in label are that the `Filter` takes the `fastest_direct` (initialized to 1440 minutes = 1 day by default to cut off search) and most filter implementations cut off search after `travel_time > fastest_direct`.
- routing module: search debugging can now be enabled with `kTracing=true` compile time variable
- `label_chain_parser`: printable state, make state `enum class` (previously `int` / `enum`)
- Label generator: increase depth for generating start labels from 8 to 16) there are cases that are not covered with 8. Proper solution would be to use a dijkstra on the footpath network first
- add debug tools `try_reproduce.py` and `nigiri-debug.py` that help reproducing issues on a minimal timetable and debugging issues (reduce output to interesting parts)


base/loader/src/build_footpaths.cc
```diff
@@ -82,7 +82,7 @@ struct footpath_builder {
         duration = *adjusted_duration;
       }
 
-      from_station->equivalent_.emplace_back(to_station.get());
+      //      from_station->equivalent_.emplace_back(to_station.get());
```
felixguendling, line 85: TODO

base/loader/src/graph_builder.cc
```diff
@@ -110,7 +110,7 @@ trip* graph_builder::register_service(Service const* s, int day_idx) {
               s->trip_id() == nullptr ? "" : s->trip_id()->str(), nullptr, 0U,
               static_cast<trip_idx_t>(sched_.trip_mem_.size()),
               s->debug() == nullptr
-                  ? trip_debug{}
+                  ? trip_debug{nullptr, 0, 0}
```
felixguendling, line 113: TODO

base/loader/src/hrd/builder/meta_station_builder.cc
```diff
@@ -23,20 +23,19 @@ Offset<Vector<Offset<MetaStation>>> create_meta_stations(
...
-          return std::nullopt;
-        }
-      })  //
+      | transform([&](station_meta_data::meta_station const& m)
+                      -> std::optional<Offset<MetaStation>> {
+          try {
+            return std::make_optional(CreateMetaStation(
+                fbb, sb.get_or_create_station(m.eva_, fbb),
+                fbb.CreateVector(utl::to_vec(m.equivalent_, [&](auto&& e) {
+                  return sb.get_or_create_station(e, fbb);
```
felixguendling, line 32: TODO

base/loader/src/rule_service_graph_builder.cc
```diff
@@ -624,7 +624,7 @@ struct rule_service_route_builder {
       full_trip_id ftid;
       push_mem(gb_.sched_.trip_mem_, ftid, "", edges_ptr, lcon_idx,
                static_cast<trip_idx_t>(gb_.sched_.trip_mem_.size()),
-               trip_debug{}, mcd::vector<uint32_t>{});
+               trip_debug{nullptr, 0, 0}, mcd::vector<uint32_t>{});
```
felixguendling, line 627: TODO

base/module/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ compile_flatbuffers_to_resource_header(
 file(GLOB_RECURSE motis-module-files src/*.cc)
 add_library(motis-module STATIC ${motis-module-files} ${resource-header})
 target_include_directories(motis-module PUBLIC include)
-target_compile_features(motis-module PUBLIC cxx_std_17)
+target_compile_features(motis-module PRIVATE cxx_std_17)
```
felixguendling, line 30: TODO

modules/routing/include/motis/routing/pareto_dijkstra.h
```diff
@@ -43,6 +45,10 @@ struct pareto_dijkstra {
 
   void add_start_labels(std::vector<Label*> const& start_labels) {
     for (auto const& l : start_labels) {
+      //      std::cout << "START: ";
```
felixguendling, line 48: TODO

modules/routing/include/motis/routing/search.h
```diff
@@ -76,11 +78,16 @@ struct search {
     }
 
     auto const& meta_goals = q.sched_->stations_[q.to_->id_]->equivalent_;
+    //    std::cout << "GOAL: " << q.sched_->stations_[q.to_->id_]->name_ << "
```
felixguendling, line 81: TODO

PR #305, opened 2023-03-30
intermodal: stop counting change times at first/last station

felixguendling:


pablohoch:
Trip Based:
https://github.com/motis-project/motis/blob/master/modules/tripbased/src/tripbased.cc#L385
https://github.com/motis-project/motis/blob/master/modules/tripbased/src/tripbased.cc#L302

`q.intermodal_start_` -> `false`
`q.intermodal_destination_` -> `false`

PR #401, opened 2023-10-19
Added spanish translation + datetime format

MrElizondo:
- Added spanish translation following commits for polish and french.
- Added spanish datetime format

### DIFF
--- a/ui/web/src/Localization/Es.elm
+++ b/ui/web/src/Localization/Es.elm
@@ -1,4 +1,4 @@
-module Localization.De exposing (esLocalization, esTranslations)
+module Localization.Es exposing (esLocalization, esTranslations)
 
 import Localization.Base exposing (..)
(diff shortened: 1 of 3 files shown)


ui/web/src/Localization/Es.elm
```diff
@@ -0,0 +1,150 @@
+module Localization.De exposing (esLocalization, esTranslations)
```
felixguendling, line 1:
```suggestion
module Localization.Es exposing (esLocalization, esTranslations)
```

felixguendling (changes requested):

ui/web/src/Localization/Es.elm
MrElizondo: Yeah, I was already making the change. Thanks!

felixguendling:
Thank you! :sunglasses:

Issue #392, opened 2023-09-01
Save and Share button in the Web UI not working


1Maxnet1:
In the public Demo https://europe.motis-project.de as well as in our private hosted instance the icons for saving and sharing a connection do nothing

![grafik](https://github.com/motis-project/motis/assets/24827631/331cb081-bf26-48eb-b3b3-4b1fd1e39d14)

I would have expected this to allow me to somehow share this connection or save it. I tested it on Linux with Firefox and Google Chrome, but the behavior is the same.

felixguendling:
I quote from my reply to #391:

> Another problem in general with the frontends (Android, web) is, that currently we do not have enough developer capacity to maintain backend and frontends in all areas on a high level. And if we give up something, it's obviously the frontends. So currently, I consider the frontends (app and web) in "maintainance mode". We still need them to showcase MOTIS but we cannot put a lot of effort into them, because our focus should be on the backend.

You are right, those buttons have no functionality. The functionality was planned but never implemented. If someone makes a PR removing them, I merge it. But we (TU Darmstadt, algorithm group) consider the web frontend in its current form more or less as "discontinued" so we can focus to improve the algorithms in the backend (where we see our strengths more than in web development).

1Maxnet1:
Resolved by #414

PR #414, opened 2023-11-28
Remove not working icons from Web UI

1Maxnet1:
As discussed in #392 the icons have no effect and therefore can be removed. This PR does just that.


felixguendling:
Thank you!

PR #436, opened 2024-02-10
Fix station name in nigiri geo lookup responses

vkrause:
This was previously returning the id in the name field as well.


felixguendling:
Thank you very much! :)

PR #437, opened 2024-02-12
Deal with invalid GBFS station status objects more gracefully

vkrause:
The national Swiss feed has individual items with a missing or invalid station_id field, skip just those rather than aborting reading the entire feed.


felixguendling:
Thank you very much! :-)

PR #444, opened 2024-02-18
Provide GTFS colors in the routing and stop event APIs when available

vkrause:
This depends on https://github.com/motis-project/nigiri/pull/67 going in and the nigiri dependency being bumped accordingly.


base/core/src/message_to_journeys.cc
```diff
@@ -47,6 +47,14 @@ journey::transport to_transport(Walk const& walk, uint16_t duration) {
   return t;
 }
 
+static uint32_t to_color(std::string_view const color_str) {
+  if (color_str.size() != 6) {
+    return 0;
+  }
+  uint32_t const c = 0xFF000000U | static_cast<uint32_t>(std::strtol(color_str.data(), nullptr, 16));
```
felixguendling, line 54: Can you please convert this to [AAA style](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)? I'm not sure if that's a cast. If it is, `static_cast` on the right side (and just `auto const` on the left) would make it more obvious.

base/core/src/message_to_journeys.cc
```diff
@@ -47,6 +47,14 @@ journey::transport to_transport(Walk const& walk, uint16_t duration) {
   return t;
 }
 
+static uint32_t to_color(std::string_view const color_str) {
+  if (color_str.size() != 6) {
```
felixguendling, line 51: Does it make sense to also check `std::all(..., isdigit)`?

vkrause: The input is a hex number, so that alone wont work, but I can certainly make this more robust (and same in nigiri, where this is probably even more important).

felixguendling:
Regarding the clang-tidy warning about `static uint32_t to_color(std::string_view const color_str)`: I think it's fine to just remove `static` and not move it to an anonymous namespace. The linker will figure it out. At least we didn't try to hide stuff from the linker anywhere else. So if this would be a goal, we would need to add `static` or anonymous namespaces at a lot of places.

vkrause:
Right, that's just habit from otherwise mostly working on shared library code.

Issue #443, opened 2024-02-17
Routing in areas with little data is very slow


jbruechert:
Queries in some areas are very slow (as in, multiple minutes).
As far as I can tell, this happens when there is a lot less data at the destination than at the start.

Reproducible examples on europe.motis-project.org include

* Warsaw, Metro Marymont -> Riga Bus Station (They are on the same bus route)
* München Hbf -> Verona Porta Nuova (Again, there is a nightjet covering the whole route)

Is there possibly some threshold when motis stops searching for other connections, that is too high in such cases?

felixguendling:
TL;DR: it's not trivial.

To understand the problem (and fix it), it's good to know how MOTIS executes PreTrip queries (also called *range query* in the literature).

A range query asks for all optimal connections within a departure interval [^1]. In contrast to the literature, we additionally require results to be not dominated by all journeys that start outside the interval (e.g. to not display a journey 10:00-17:00 if there's also 11:00-17:00 starting outside the interval). This is achieved by having an additional OnTrip (aka *earliest arrival*) search at "end of the interval" + 1. The results of this additional OnTrip request are only used to dominate non-optimal results from inside the interval but otherwise ignored. So the range query how we define it in MOTIS is "Give me all globally Pareto-optimal journeys that depart within interval [t1, t2]".

Now, there's trick how to make range-queries fast. From the initial RAPTOR paper [^0] Section *4.2 Range Queries: rRAPTOR*:

> [...] we order Ψ from latest to earliest, and then run RAPTOR for every τ ∈ Ψ in order. However, we keep the labels τk(p) between rounds instead of reinitializing them. To see why this is correct, note that this value of τk(p) corresponds to an intermediate journey departing from ps no earlier than journeys computed in the current run (recall that Ψ is ordered). Thus, if τk(p) is smaller, we also know how to reach p earlier. Hence, we can safely prune the current journey.

This means, we can reuse all labels from later departure times if we iterate departures in the interval from late to early, starting with the first earliest arrival query at t2+1 as discussed before to eliminate journeys that are only locally Pareto-optimal but are dominated by journeys that depart outside the interval.

[...]

PR #445, opened 2024-02-18
Allow to set a timeout for searches

jbruechert:
If supported by the chosen router, it will stop expanding its search interval after the given timeout.

Depends on https://github.com/motis-project/nigiri/pull/68

Workarounds https://github.com/motis-project/motis/issues/443 enough to not be a blocker for public hosting.

Let me know if this is generally something you want, and if I should change anything about the implementation.

### DIFF
--- a/modules/intermodal/include/motis/intermodal/intermodal.h
+++ b/modules/intermodal/include/motis/intermodal/intermodal.h
@@ -26,6 +26,7 @@ private:
 
   std::string router_{"routing"};
   bool revise_{false};
+  unsigned timeout_{0};
   ppr_profiles ppr_profiles_;
 };
 
--- a/modules/intermodal/src/intermodal.cc
+++ b/modules/intermodal/src/intermodal.cc
@@ -38,6 +38,7 @@ namespace motis::intermodal {
 intermodal::intermodal() : module("Intermodal Options", "intermodal") {
   param(router_, "router", "routing module");
   param(revise_, "revise", "revise connections");
+  param(timeout_, "timeout", "routing timeout in seconds (0 = no timeout)");
 }
 
 intermodal::~intermodal() = default;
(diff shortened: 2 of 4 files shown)


modules/intermodal/include/motis/intermodal/intermodal.h
```diff
@@ -26,6 +26,7 @@ struct intermodal : public motis::module::module {
 
   std::string router_{"routing"};
   bool revise_{false};
+  int timeout_ = 0;
```
felixguendling, line 29:
```suggestion
  unsigned timeout_{0};
```

As negative values don't make sense here, maybe `unsigned` and please use the `var_{..}` style (just for consistency - both styles work here).

modules/intermodal/src/intermodal.cc
```diff
@@ -38,6 +38,7 @@ namespace motis::intermodal {
 intermodal::intermodal() : module("Intermodal Options", "intermodal") {
   param(router_, "router", "routing module");
   param(revise_, "revise", "revise connections");
+  param(timeout_, "timeout", "timeout in seconds for searching for routes (0 = no timeout)");
```
felixguendling, line 41: maybe more compact comment "routing timeout in seconds (0 = no timeout)"
```suggestion
  param(timeout_, "timeout", "routing timeout in seconds (0 = no timeout)");
```

modules/nigiri/src/routing.cc
```diff
@@ -155,6 +157,13 @@ motis::module::msg_ptr route(tag_lookup const& tags, n::timetable const& tt,
   auto extend_interval_later = false;
   auto start_time = n::routing::start_time_t{};
   auto start_station = n::location_idx_t::invalid();
+  auto timeout = [&]() -> std::optional<std::chrono::seconds> {
+    if (req->timeout() == 0) {
+      return {};
```
felixguendling, line 162: `std::nullopt` to make it more obvious, what's happening here.

```suggestion
      return std::nullopt;
```

felixguendling:
If you want to prevent a malicious user from using CPU resources, this won't be sufficient, because this only "protects" the intermodal API, not the routing API. Make sure to have a whitelist of API endpoints (that doesn't contain the `/nigiri` or `/routing` endpoint) in some kind of reverse proxy / load balancer in front of MOTIS. You probably also need to have rate limits in your reverse proxy.

felixguendling:
The CI build is all red, so there seem to be errors in the code.

PR #453, opened 2024-02-26
nigiri clasz filter

felixguendling:


felixguendling:
Based on https://github.com/motis-project/nigiri/pull/72

PR #460, opened 2024-03-01
Handle relative HTTP redirects as well

vkrause:
Use boost::url::resolve to handle all possible redirect variants. The previous approach only handled redirects to absolute URLs. Some of the US GTFS-RT feeds use relative redirects though which caused an exception and aborted the entire GTFS-RT update then.

### DIFF
--- a/base/module/src/context/motis_http_req.cc
+++ b/base/module/src/context/motis_http_req.cc
@@ -1,6 +1,8 @@
 #include "motis/module/context/motis_http_req.h"
 
 #include "boost/algorithm/string/predicate.hpp"
+#include "boost/url/url.hpp"
+#include "boost/url/url_view.hpp"
 
 #include "net/http/client/http_client.h"
 #include "net/http/client/https_client.h"
@@ -27,6 +29,7 @@ struct http_request_executor
     l(debug, "http request {} {}",
       net::http::client::method_to_str(req.req_method),
       fmt::streamed(req.address));
+    request_url_ = req.address;
 
     auto cb = [self = shared_from_this()](auto&& a,
                                           net::http::client::response&& res,
@@ -76,13 +79,17 @@ struct http_request_executor
       }
     }
 
+    auto resolved_url = boost::urls::url{};
+    boost::urls::resolve(boost::urls::url_view(request_url_.str()),
+                         boost::urls::url_view(target), resolved_url);
     ++redirect_count_;
-    make_request(url(target));
+    make_request(url(resolved_url.c_str()));
   }
 
   unsigned redirect_count_{0U};
   boost::asio::io_service& ios_;
   http_future_t f_;
+  url request_url_;
 };
 
 std::shared_ptr<ctx::future<ctx_data, net::http::client::response>>
--- a/.pkg
+++ b/.pkg
@@ -89,7 +89,7 @@
 [boost]
   url=git@github.com:motis-project/boost.git
   branch=master
-  commit=be5235eb2258d2ec19e32546ab767a62311d9b46
+  commit=2db0acc02ea3f3775d1f8b48ba6c10f58dd6fcc9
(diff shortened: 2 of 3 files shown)


base/module/src/context/motis_http_req.cc
```diff
@@ -76,13 +77,24 @@ struct http_request_executor
...
     }
 
+    url target_url;
+    try {
+      target_url = url(target);
+    } catch (std::invalid_argument const&) {
+      // not an absolute redirect, try relative redirect with absolute path
+      // (technically we'd also need to handle relative paths here)
+      target_url = url(fmt::format("{}://{}{}", request_url_.prot(),
+                                   request_url_.host(), target));
```
derhuerst, line 87: Does this work if the redirect is e.g. to `foo`, or `./foo`, or `../foo/bar`?

vkrause: No, as mentioned redirects to relative paths are still not supported, that will continue to produce an exception as before. This is just patching up the bare minimum needed to make all GTFS-RT feeds currently in Transitous work.

Fixing this properly is more effort, but before doing that I'd suggest to explore using an existing HTTP/URL/etc implementation instead of continuing to do all this manually.

derhuerst: Yes, 💯 for using an off-the-shelf HTTP client. (libcurl?)

pablohoch: Using [Boost.URL](https://www.boost.org/doc/libs/1_82_0/libs/url/doc/html/url/ref/boost__urls__resolve.html) would probably be more robust and should handle these cases.

felixguendling: Both, using libcurl as well as switching to Boost.URL would need to be done in the [`net` library](https://github.com/motis-project/net). Long term, switching to something off-the-shelf makes sense. But to enable MOTIS today to work with relative redirects, I'll merge it as it is.

vkrause: Oh, indeed, that is much better! Seeing a handwritten URL class here I didn't even think about that we do have Boost available already.

felixguendling:
In general, we try to follow the rule that exceptions should only be thrown for exceptional stuff (that cannot be expected and are definitely an error). Here, the exception is used for control flow for a completely (non-exceptional) normal case. Can those cases be detected by checking `target.starts_with("/")` or a similar check to not run into the `throw` in the first place? Otherwise, I would try to expose a `is_url(std::string_view)` in the `net` library. Making sure exceptions are thrown only for real errors makes it easier to debug MOTIS with `catch throw` in the debugger.

vkrause:
There is no exception handling anymore after the switch to boost::url::resolve, I'm just not getting Github to update PR...
(see https://github.com/motis-project/motis/compare/master...vkrause:motis:work/vkrause/fix-http-relative-redirect meanwhile).

felixguendling:
Looks like Mac OS doesn't like Boost.URL. Maybe changing `target_compile_features(motis-module PUBLIC cxx_std_17)` to C++20 helps?

vkrause:
The errors all refer to C++11 features and happen while building Boost.URL itself IIUC, so my guess would be https://github.com/motis-project/boost/pull/3.

base/module/src/context/motis_http_req.cc
```diff
@@ -76,13 +79,17 @@ struct http_request_executor
       }
     }
 
+    boost::urls::url resolvedUrl;
+    boost::urls::resolve(boost::urls::url_view(request_url_.str()),
+                         boost::urls::url_view(target), resolvedUrl);
```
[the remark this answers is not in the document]
felixguendling: AAA style + snake case

felixguendling (approved):

PR #486, opened 2024-03-31
enable gbfs+osr

felixguendling:


felixguendling:
Enables it only in theory - in practice it's very slow as `osr` doesn't have real table-routing. It can be improved by setting `max_dist` parameter in `osr` appropriately (requires protocol adjustment) but the clean approach would be to do it all in Dijkstra's algorithm directly.

Issue #452, opened 2024-02-24
MOTIS Potential


felixguendling:
This GitHub issue exists to give an overview of features that would extend MOTIS in a valuable way. If you want to discuss a specific feature in more detail, please open a separate issue. I will link those more detailed discussion issues here for reference to keep an overview. Most of the features would need to be implemented in the routing core [nigiri](https://github.com/motis-project/nigiri).

- [ ] Via-Routing: give the routing a list of via via-(stations/coordinates) with a minimum required stay time for each station/coordinate. The routing should produce Pareto-optimal results (as always) for all journeys that go through this list of stops, staying at each stop at least the given time. Note: this is not a traveling salesman problem as the routing will not optimize the ordering of the via-stops.
- [ ] Mode of (public) transport restriction: for any combination of the [transport classes](https://github.com/motis-project/nigiri/blob/4adc2c34fd09b13a7b3cad3a4b4a756b5d04c6b7/include/nigiri/types.h#L250-L265): allow to compute Pareto-optimal journeys that only involve the selected modes of transportation.
- [ ] Limiting the number of transfers. This can also be done external to MOTIS by filtering the returned routing results. However, it's possible to save some computation time if the maximum number of transfers is known beforehand. This option is already [supported](https://github.com/motis-project/nigiri/blob/4adc2c34fd09b13a7b3cad3a4b4a756b5d04c6b7/include/nigiri/routing/query.h#L43) by the routing core - it's just not available via the API.
- [ ] Override the minimum transfer time for every station when routing or multiplying each transfer time with a constant factor.
- [ ] Limit the routing to connections where bicycle carriage is allowed. To stay algorithmically correct, this needs to be incorporated into the [route definition](https://github.com/motis-project/nigiri/blob/master/include/nigiri/loader/gtfs/route_key.h#L7) (transports on a route segment need to either all allow or all not allow bicycle carriage).
- [ ] Support for NeTEx and SIRI. The main reason is that GTFS-RT currently does not support important use cases such as reroutings or additional trains.
- [ ] Support Service Alerts (e.g. for now: in GTFS-RT)
- [ ] During [journey reconstruction](https://github.com/motis-project/nigiri/blob/master/src/routing/raptor/reconstruct.cc), if there are several stations where it's possible to switch between two trips: chose the "best" transfer station regarding some criteria (e.g. station size).
- [ ] Read `shapes.txt` and output compressed polylines for the public transport part in API routing results.
[...]

PR #492, opened 2024-04-23
Allow to disable importing

jbruechert:
I haven't thoroughly tested this yet


felixguendling:
I guess another option would be to split the import functionality into preprocessing the data and actually loading the data. My plan is to address this and many other issues mentioned in #452 (such as a stable API + versioning, running MOTIS without the `input` folder, etc.) in one go by creating something like "MOTIS 2.0". My plan is to start working on this big refactoring in one month.

PR #522, opened 2024-06-19
Add transfer time settings

pablohoch:
Adds two new options to `RoutingRequest` and `IntermodalRoutingRequest`:
- `min_transfer_time`: Minimum transfer time in minutes
- `transfer_time_factor`: Multiply all transfer times by this factor

Needs https://github.com/motis-project/nigiri/pull/113 or https://github.com/motis-project/nigiri/pull/114


pablohoch:
> Hi @pablohoch, is there some documentation somewhere that explains the interest of this feature ? Usecases for example ? I'm intrigued :)

You can use these new options to adjust the transfer times used by the routing algorithm. Using `min_transfer_time` you can set a minimum time (in minutes) for all transfers and by using `transfer_time_factor` you can increase the time of all transfers, e.g. by 2.0x. For example, if a transfer takes 5 minutes according to the schedule, by setting `min_transfer_time = 10` or `transfer_time_factor = 2.0` you can ensure that you have at least 10 minutes for that transfer (the options are applied to all transfers).

There are various use cases: For example, if you know that you need more time for transfers (e.g. travelling with lots of luggage, mobility impairments etc.) or you expect delays and want connections where you are more likely to catch the connecting trains.

laem:
I'm seing the opportunity of a first integration of these features in the form of a "turtoise option", that triples transfer times. Exposing the whole complexity of these two parameters combined would overwhelm the user (but could be made available in more advanced options).  

image

PR #537, opened 2024-09-04
Show routing between stations

MichaelKutzner:
The server will now show the intended routing between stations, if available. This requires GTFS feeds to contain a valid `shapes.txt` file. Trips without a specific shape will not change and still show an interpolation between stops.
For performance reasons, this feature is used for selected trips only.


MichaelKutzner:
I opened this Pull Request to check for issues, already. As the required changes to `nigiri` aren't merged, yet, this cannot be merged likewise.

derhuerst:
> As the required changes to `nigiri` aren't merged, yet, this cannot be merged likewise.

I assume you're talking about https://github.com/motis-project/nigiri/pull/127 ?

MichaelKutzner:
Yes, you're right. I forgot about the reference. Sorry about that.

MichaelKutzner:
I just noticed this error:

```
/__w/motis/motis/modules/nigiri/src/nigiri.cc:479:52: error: implicit conversion 'unsigned int' -> bool [readability-implicit-bool-conversion,-warnings-as-errors]
  479 |                                 merge_duplicates_, max_footpath_length_);
      |                                                    ^                   
      |                                                                         != 0u
```

Given the recent [change to `finalize()`](https://github.com/motis-project/nigiri/pull/128/files#diff-5d9321a1cafac0590569e1d1e9c034266b6b3f7df1fd8cebd9d5077ee4ba1ec8) split a parameter, I assume it's save to use `merge_duplicates_` for both arguments without changing the behavior?

I guess, I will create a new Pull Request fixing this first?

MichaelKutzner:
Thank you for explanation. I didn't see the impact or difference after running a quick test locally.

PR #539, opened 2024-09-06
Update for latest nigiri changes

MichaelKutzner:
Fix compile error caused by changed function signature.

Call to `nigiri::loader::finalize()` now requires an additional argument: https://github.com/motis-project/nigiri/pull/128/files#diff-5d9321a1cafac0590569e1d1e9c034266b6b3f7df1fd8cebd9d5077ee4ba1ec8
For build log see #537


felixguendling:
This would not be backward compatible.

@mority already did a version that lets you configure both options.
https://github.com/motis-project/motis/pull/538/files

MichaelKutzner:
Totally missed #538, which should fix it as well.

PR #552, opened 2024-10-11
Update nigiri to include motis-project/nigiri#144

MichaelKutzner:
This updates nigiri version to include a fix for incomplete segments.


MichaelKutzner:
The change was tested for datasets with and without `shapes.txt` provided.
Notice that the fix applied with motis-project/nigiri#144 can lead to bad shapes, if a trip passes a stop multiple times without providing `shapes_dist_travled`. This is due to the algorithm currently used to match stops to shape points.

PR #565, opened 2024-10-15
add stops endpoint, rename railviz to trips

felixguendling:


exe/server.cc
```diff
@@ -71,7 +72,8 @@ int server(data d, config const& c) {
   GET<ep::routing>(qr, "/api/v1/plan", d);
   GET<ep::stop_times>(qr, "/api/v1/stoptimes", d);
   GET<ep::trip>(qr, "/api/v1/trip", d);
-  GET<ep::railviz>(qr, "/api/v1/railviz", d);
+  GET<ep::trips>(qr, "/api/v1/railviz", d);
```
anbraten, line 75: Probably needs to be `/api/v1/trips` 🤔 or a separate line so both are supported for the moment.

felixguendling: #566 should fix, thank you!

PR #569, opened 2024-10-16
ui: Add URL parameters to pass start and destination

jbruechert:


felixguendling:
> `Cannot access url.searchParams on a page with prerendering enabled`

I'm not a Svelte expert but this worked okay for GET params:
https://github.com/motis-project/motis/blob/ad5060b1d7b5866dbbc1a2941801706981ce6267/ui/src/routes/%2Blayout.ts#L6-L7

jbruechert:
That doesn't work with server side rendering, and `npm run dev` is configured to do that.

felixguendling:
`dark` and `debug` parameters seem to also work in development mode?

jbruechert:
Ah right, it was guarded by browser &&. I hope this works now, I fought with typescript for way too long.

PR #570, opened 2024-10-16
railviz rt coloring

felixguendling:


felixguendling:
included in https://github.com/motis-project/motis/pull/583

PR #576, opened 2024-10-16
replace http server

felixguendling:


felixguendling:
uWebSocket doesn't have pipelining

PR #603, opened 2024-10-22
Improve bounding boxes used by railviz

MichaelKutzner:
Fixes an issue for trips with shapes not completely covered by bounding boxes created by stop locations. This caused trips to vanish for curved shaped, if the visible view didn't overlap with the direct route between stops.
For further improvements this calculation is now performed during import. Trivial bounding boxes for segments might not be stored to save disk space.


src/railviz.cc
```diff
@@ -93,19 +97,27 @@ struct route_geo_index {
...
                   n::clasz const clasz,
                   n::vector_map<n::route_idx_t, float>& distances) {
+    // Fallback, if no route bounding box can be loaded
+    auto const get_box = [&](n::route_idx_t const route_idx) {
+      auto bounding_box = geo::box{};
+      for (auto const l : tt.route_location_seq_[route_idx]) {
+        bounding_box.extend(
+            tt.locations_.coordinates_.at(n::stop{l}.location_idx()));
+      }
+
```
felixguendling, line 110:
```suggestion
```

PR #608, opened 2024-10-23
Validate OpenAPI specification in CI

Schmiddiii:


Schmiddiii:
Works as expected, except failing docker publish due to fork which is expected (as mentioned in the Matrix chat).
Should be ready.

felixguendling:
Thank you! :partying_face:

PR #648, opened 2024-11-05
car parking

felixguendling:


felixguendling:
warning: doesn't really work because most parking lots (where I checked) are not connected to the street network in OSM

PR #667, opened 2024-11-21
Add French translation.

kalon33:
Add french translation for Motis.


felixguendling:
Thank you very much! :)

PR #668, opened 2024-11-21
Use `maxTravelTime` to filter results

MichaelKutzner:
Rename query parameter `maxHours` to `maxTravelTime`.
Setting `maxTravelTime` will prevent calculating journeys with a travel time larger than `maxTravelTime`.


openapi.yaml
```diff
@@ -359,7 +359,7 @@ paths:
           schema:
             type: integer
 
-        - name: maxHours
+        - name: maxTravelTime
```
felixguendling, line 362: maybe change unit to minutes to give users more fine grained control?

MichaelKutzner: Sure, I can change that. Using floats works too, but that might not be intuitive for everyone.

PR #681, opened 2024-12-06
osr matches

felixguendling:


src/match_platforms.cc
```diff
@@ -232,4 +235,62 @@ osr::platform_idx_t get_match(n::timetable const& tt,
...
+
+void way_matches_storage::preprocess_osr_matches(
+    nigiri::timetable const& tt,
+    osr::platforms const& pl,
+    osr::ways const& w,
+    osr::lookup const& l,
+    platform_matches_t const& platform_matches) {
+
+  for (auto i = n::location_idx_t{0U}; i != tt.n_locations(); ++i) {
+    auto const loc = n::location_idx_t{i};
```
mority: `i` is already a `location_idx_t`

src/match_platforms.cc
```diff
@@ -232,4 +235,62 @@ osr::platform_idx_t get_match(n::timetable const& tt,
...
+    auto const p = platform_matches[loc];
```
mority: `p` seems to only be used once?

src/match_platforms.cc
```diff
@@ -232,4 +235,62 @@ osr::platform_idx_t get_match(n::timetable const& tt,
...
+    matches_.emplace_back(l.get_raw_match(
+        osr::location{tt.locations_.coordinates_[loc], pl.get_level(w, p)},
+        max_matching_distance_));
```
mority:
```
matches_.emplace_back(l.get_raw_match(
        osr::location{tt.locations_.coordinates_[i], pl.get_level(w, platform_matches[i])},
        max_matching_distance_));
```

src/match_platforms.cc
```diff
@@ -232,4 +235,62 @@ osr::platform_idx_t get_match(n::timetable const& tt,
...
+    osr::lookup const& lookup,
+    way_matches_storage const* way_matches,
+    osr::search_profile const p,
+    std::span<nigiri::location_idx_t const> const locations,
+    std::span<osr::location const> const osr_locations,
+    osr::direction const dir,
+    double const max_matching_distance) {
+  auto const use_raw_matches =
+      way_matches && !way_matches->matches_.empty() &&
+      way_matches->max_matching_distance_ >= max_matching_distance;
```
mority, line 282: Not sure about this one.
I presume, the `way_matches` were found with a max. distance of `way_matches->max_matching_distance_`.
So, the distance of each individual way match is between 0 and `way_matches->max_matching_distance_`. 
But then even for cases where `way_matches->max_matching_distance_` < `max_matching_distance` there might be relevant way matches since they all satisfy the condition of having a distance smaller than `max_matching_distance`

include/motis/match_platforms.h
```diff
@@ -26,4 +51,13 @@ platform_matches_t get_matches(nigiri::timetable const&,
 
 std::optional<std::string_view> get_track(std::string_view);
 
+std::vector<osr::match_t> get_many_platform_way_matches(
```
mority: Why is it called "many"?

src/endpoints/routing.cc
```diff
@@ -233,10 +243,15 @@ std::vector<n::routing::offset> get_offsets(
...
-              max_matching_distance, nullptr, &sharing, elevations);
+
+          auto const osr_profile = gbfs::get_osr_profile(prod);
+          auto const pos_match = r.l_->match(
+              pos, false, dir, max_matching_distance, nullptr, profile);
+          auto const paths =
+              osr::route(*r.w_, *r.l_, osr_profile, pos, near_stop_locations,
+                         pos_match, near_stop_matches(osr_profile),
+                         static_cast<osr::cost_t>(max.count()), dir, nullptr,
+                         &sharing, elevations);
```
mority: seems pretty similar to lines 268 - 273, maybe define a lambda and call it twice?

src/endpoints/routing.cc
```diff
@@ -250,10 +265,12 @@ std::vector<n::routing::offset> get_offsets(
...
-      auto const paths =
-          osr::route(*r.w_, *r.l_, profile, pos, near_stop_locations,
-                     static_cast<osr::cost_t>(max.count()), dir,
-                     max_matching_distance, nullptr, nullptr, elevations);
+      auto const pos_match =
+          r.l_->match(pos, false, dir, max_matching_distance, nullptr, profile);
+      auto const paths = osr::route(
+          *r.w_, *r.l_, profile, pos, near_stop_locations, pos_match,
+          near_stop_matches(profile), static_cast<osr::cost_t>(max.count()),
+          dir, nullptr, nullptr, elevations);
```
mority: see line 254

src/endpoints/routing.cc
```diff
@@ -233,10 +243,15 @@ std::vector<n::routing::offset> get_offsets(
               gbfs_rd.get_products_routing_data(*provider, prod.idx_);
           auto const sharing = prod_rd->get_sharing_data(
               r.w_->n_nodes(), ignore_rental_return_constraints);
-          auto const paths = osr::route(
-              *r.w_, *r.l_, gbfs::get_osr_profile(prod), pos,
-              near_stop_locations, static_cast<osr::cost_t>(max.count()), dir,
-              max_matching_distance, nullptr, &sharing, elevations);
+
+          auto const osr_profile = gbfs::get_osr_profile(prod);
```
mority: `profile` is also an `osr::search_profile`. Maybe rename them to something like `mode_profile` and `rental_profile` to better distinguish them?

mority (changes requested):

src/match_platforms.cc
traines-source: Yes, but in that case for simplicity the matches are just recomputed entirely as before without preprocessed raw matches. I think the effort of merging the two is not worth it because ideally this happens rarely because a sensible `way_matches->max_matching_distance_` was chosen and if it happens, the larger `max_matching_distance` is probably so much higher that the preprocessed matches only represent a minority of the needed matches.

include/motis/match_platforms.h
traines-source: Because due to `reverse: true` in there it can currently only be used for the `many` of one-to-many. But I'll try to find a clearer name.

test/matching_test.cc
```diff
@@ -123,6 +123,9 @@ TEST(motis, get_way_candidates) {
   for (auto const [with, without, l] :
        utl::zip(with_preprocessing, without_preprocessing, locs)) {
     ASSERT_EQ(with.size(), without.size());
+    for (auto [a, b] : utl::zip(with, without)) {
+      std::cout << a.dist_to_way_ << " dist " << b.dist_to_way_ << std::endl;
+    }
```
mority, line 136: avoid `std::endl`
suggestion: `fmt::println("{} dist {}", a.dist_to_way_, b.dist_to_way_)`

traines-source: oops committed by accident :D

src/match_platforms.cc
traines-source: I've switched to using a `std::optional` now to make this clearer (also has the advantage that stops for which there are actually no matches are not reprocessed)

src/import.cc
```diff
@@ -446,20 +446,38 @@ data import(config const& c, fs::path const& data_path, bool const write) {
...
+        if (write) {
+          cista::write(data_path / "matches.bin", *d.matches_);
+        }
+        if (c.timetable_.value().preprocess_max_matching_distance_ > 0.0) {
+          progress_tracker->status("Prepare Platform Way Matches")
+              .out_bounds(50, 100);
+          d.way_matches_ = std::make_unique<way_matches_storage>(
+              data_path, cista::mmap::protection::WRITE,
+              c.timetable_.value().preprocess_max_matching_distance_);
+          d.way_matches_->preprocess_osr_matches(*d.tt_, *d.pl_, *d.w_, *d.l_,
```
traines-source, line 469: multithreading?

mority (approved):

PR #689, opened 2024-12-21
cmake: Add option to pass flags not used for host tools

jbruechert:
This allows to target machines that support more instructions than the build machine.

Not sure if this is a good way to achieve this, but it at least doesn't require significant changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,3 +198,9 @@ add_custom_target(motis-web-ui
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/ui"
         VERBATIM
 )
+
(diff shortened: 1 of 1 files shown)


felixguendling:
Compiling `mimalloc` with `-Weverything [...] -Werror` has a good chance to cause trouble because we don't want to fix random warnings in mimalloc when we update. So maybe it would be better to have a separate set of flags (`MOTIS_ARCH_FLAGS`?) that's independent of the aggressive `-Weverything [...] -Werror` flags that are only relevant for our own code.

jbruechert:
That's pretty mich what this does except for different variable naming, isn't it?

felixguendling:
Ahh. Mistook it for `motis-compile-options`.

PR #700, opened 2025-01-17
Fix missing closing tag

MichaelKutzner:
I also wasn't sure, if `_idx` was intended. But I didn't change it, as it's often used as suffix.

Also for the _Tooling_ section. Would it be useful to mention, that some checks can cause false positives and may be disabled, e.g. `ASAN_OPTIONS=alloc_dealloc_mismatch=0`?

### DIFF
--- a/docs/STYLE.md
+++ b/docs/STYLE.md
@@ -128,7 +128,7 @@ Only use shortened version of the variable name if it's still obvious what the v
 - Output = `out`
 - Request = `req`
 - Response = `res`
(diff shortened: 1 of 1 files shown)


felixguendling:
Yes, I agree to both. `_idx` was because we use it often as postfix. But maybe that's not relevant here. A section about `strong` and something explaining `vecvec` and `vector_map` would also make sense.

felixguendling:
Thank you! :-)

PR #704, opened 2025-01-18
Fix index out of range for stop_times last arrival of multi-section run

traines-source:
I stumbled upon a motis crash `libc++abi: terminating due to uncaught exception of type cista::cista_exception: bucket::at: index out of range` with the 06.01.2025 DELFI GTFS and this query (didn't dare to try it out on one of the public instances...):
`/api/v1/stoptimes?stopId=gtfs_de%3A03159%3A33817%3A%3A8&time=2025-01-18T13%3A08%3A00.000Z&arriveBy=true&n=10` 

Occurs in [nigiri/src/rt/frun.cc#L105](https://github.com/motis-project/nigiri/blob/0861ed983e0b84062f49e82c416b2e4aa000e972/src/rt/frun.cc#L105) for the last stop of multi-section trips, because [it would always look for a `event_type::kDep`](https://github.com/motis-project/motis/commit/9902350695aca90516d6b8f25ee0861408b5e74f#diff-ffee06f9f27f250abca7b76072cbdd73f6f7768946170f1b5d2817004081897dL55).
I.e. an alternative fix with less (interface) changes would be to make [`get_trip_index(ev_type)` in nigiri](https://github.com/motis-project/nigiri/blob/0861ed983e0b84062f49e82c416b2e4aa000e972/src/rt/frun.cc#L105) more robust, but would then still lead to wrong results for the `trip_id` at stops where the `trip_id` switches (if I understand correctly).

### DIFF
--- a/include/motis/tag_lookup.h
+++ b/include/motis/tag_lookup.h
@@ -20,7 +20,7 @@ struct tag_lookup {
   std::string id(nigiri::timetable const&, nigiri::location_idx_t) const;
   std::string id(nigiri::timetable const&,
                  nigiri::rt::run_stop,
-                 nigiri::event_type const) const;
+                 nigiri::event_type) const;
   nigiri::location_idx_t get_location(nigiri::timetable const&,
                                       std::string_view) const;
   std::pair<nigiri::rt::run, nigiri::trip_idx_t> get_trip(
--- a/include/motis/tag_lookup.h
+++ b/include/motis/tag_lookup.h
@@ -18,7 +18,9 @@ struct tag_lookup {
   nigiri::source_idx_t get_src(std::string_view tag) const;
   std::string_view get_tag(nigiri::source_idx_t) const;
   std::string id(nigiri::timetable const&, nigiri::location_idx_t) const;
-  std::string id(nigiri::timetable const&, nigiri::rt::run_stop) const;
+  std::string id(nigiri::timetable const&,
+                 nigiri::rt::run_stop,
+                 nigiri::event_type const) const;
   nigiri::location_idx_t get_location(nigiri::timetable const&,
                                       std::string_view) const;
   std::pair<nigiri::rt::run, nigiri::trip_idx_t> get_trip(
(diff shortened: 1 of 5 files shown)


include/motis/tag_lookup.h
```diff
@@ -18,7 +18,9 @@ struct tag_lookup {
   nigiri::source_idx_t get_src(std::string_view tag) const;
   std::string_view get_tag(nigiri::source_idx_t) const;
   std::string id(nigiri::timetable const&, nigiri::location_idx_t) const;
-  std::string id(nigiri::timetable const&, nigiri::rt::run_stop) const;
+  std::string id(nigiri::timetable const&,
+                 nigiri::rt::run_stop,
+                 nigiri::event_type const) const;
```
felixguendling, line 23:
```suggestion
                 nigiri::event_type) const;
```

`const` says that the implementation doesn't modify but is irrelevant for the caller in case of value parameters -> should be omitted in the header (only relevant int the source so the compiler can check)

felixguendling (changes requested):

felixguendling:
Thank you! :stars: 

Do you think it makes sense to add a `utl::verify` somewhere to prevent this from happening in case the wrong event type would be supplied in the future?

felixguendling:
Hmm.. it already fired an exception. I would've hoped that this gets caught somewhere in the HTTP handler.

include/motis/tag_lookup.h
traines-source: ah yes should have remembered the style guide...

traines-source:
Yes rather than `utl::verify` (which would basically have the same result, just with a nicer error message, right?) I would maybe make `get_trip_index` more robust. And yes I was surprised that this took down motis entirely, not only the thread or sth.

felixguendling:
My hope was that at least the "catch all" here would've caught it:
https://github.com/motis-project/net/blob/30f8b05a435af6f8327376d6e95089b4341ac4f7/src/web_server/query_router.cc#L98-L112

felixguendling (approved):

felixguendling:
Ah. I guess it's because those methods in `frun` are marked `noexcept`. Maybe it would be better to just get writ of `noexcept` in MOTIS code completely except if it's proven to improve performance.

traines-source:
Ah yes. And maybe we should, because a dying motis is a bad thing :)

Issue #706, opened 2025-01-20
Improve nigiri generated footpaths durations


Pycpik:
Hi,

We would like to improve the footpaths suggested by MOTIS. 

The pedestrian times calculated by nigiri (GTFS or generated) are regularly too shorts compared to reality. The use of OSR footpaths seems to have better durations but does not always work.

As discussed with @felixguendling on Matrix, two fast improvments could be made:

- the generated times are currently very simple. 
  https://github.com/motis-project/nigiri/blob/0861ed983e0b84062f49e82c416b2e4aa000e972/src/loader/gtfs/stop.cc#L306-L320
  Something better than the hardcoded `2_minutes` could be found.

- We could try to route with osr and if it fails use GTFS/generated ones.

I create this issue to track possible improvements.

Thank you

felixguendling:
I created a first version: https://github.com/motis-project/motis/releases/tag/v2.0.14

New configuration options - those are only active with `osr_footpath=true`:

- `extend_missing_footpaths` - adds footpaths from the default footpaths but calculates footpath length based on 0.7m/s (use `true` to try it)
- `use_osm_stop_coordinates` - updates stop coordinates according to the stop object matched in OSM (probably you want `false`)

Then when routing use:

- `useRoutedTransfers=true` to enable the `osr_footpaths`
- `pedestrianProfile=FOOT` (which is the default value, so you can also omit it)

Looking forward to your feedback!

Pycpik:
Hi,

We've tested the 2.0.14 with `osr_footpaths=true`and `useRoutedTransfers=true`. 

Functionally, the routes are different but relevant. However I could not find any case with no itineraries in the MOTIS response although we found some the first time we tried `useRoutedTransfers`.

In load testing, we saw an increase in average response time of 45% (330 ms to 480ms) but we have to do more tests to find out where it comes from.

We still have to test `timetable.extend_missing_footpaths=true`.

felixguendling:
I would recommend to turn on `timetable.extend_missing_footpaths=true` unless you know that all stops are modeled perfectly and in a `osr` compatible way. Having less transfers due to modeling problems in OSM can explain differences in routing times.

PR #707, opened 2025-01-21
osr_footpaths: add missing footpaths

felixguendling:


felixguendling:
fixes #706

(work in progress)

Issue #708, opened 2025-01-21
WebUI contrast issue: white text on white line badge not visible


Altonss:
In dark mode, this can happen:

![Image](https://github.com/user-attachments/assets/3dcacc59-36b8-4de0-bb8f-86f177c51ae1)

felixguendling:
If that's not the `route_text_color`, then we have a bug.
If it's the `route_text_color` there's not much we can do except ask the agency to change their coloring choice.

traines-source:
I think this must come from `route_text_color` (but haven't yet found the train that you screenshotted), since there is no white  default background color [and so the text color can't be default either](https://github.com/motis-project/motis/blob/945ebef7e18cd110f872bb1c1d1055ce87873200/ui/src/lib/modeStyle.ts#L93). But in this case it should look the same in normal mode.

traines-source:
Ah found it, the `route_text_color` is `undefined`, and so it defaults back to the document default color (i.e. only in dark mode to white). Fixing it now.

traines-source:
Yes I thought about doing that, but then I found out that it is actually [specified](https://gtfs.org/documentation/schedule/reference/#routestxt) that `route_text_color` should fall back to black.
So not sure rn if the contrast detection is worth the effort...

felixguendling:
If it's coming from the dataset and we're doing what the standard says, I would rather contact the data provider to fix this. I think it's out-of-scope for MOTIS to improve contrast. It's better to do this earlier in the pipeline.

felixguendling:
Another approach would to fix it with gtfs-tidy before it comes into MOTIS. Would that make sense?

felixguendling:
At least, the code would only be implemented once and not every UI has to do it again and again. If it's bad data that affects all clients, it would probably make sense to fix it once for everyone.

PR #705, opened 2025-01-18
(Mobile) UI improvements

traines-source:
Some improvements for the UI, without changing too much (see https://github.com/public-transport/transitous/issues/789). This is not perfect yet, consider it a WIP.
* many small CSS, some JS tweaks
* some backports from [prima-mobile](https://github.com/motis-project/prima-mobile)
* improve error handling, resolves #647 
* add URL state management for stops and trips (while I think one could have used [SvelteKit Routing](https://svelte.dev/docs/kit/routing), I've opted for `pushState` like in prima-mobile, which means less refactoring and easier web server configuration)

TODO:
* annoying bug in the AdressTypeahead Combobox on mobile, where if you select an entry, you accidentally click other buttons (Arrival/Departure/Wheelchair beneath...) Didn't yet manage to fix that...
* showing map only on demand on mobile
* URL state mgmt for the itineraries (could be very useful, but also very confusing if returned itineraries change, because we do not have connection ids)

PR #710, opened 2025-01-21
UI fixes

traines-source:
* this resolves, if we agree that checking GTFS-provided colors for sufficient contrast is not in scope rn, #708
* this does very ugly things to at least temporarily fix the effect of accidentally tapping on the Departure/Arrival buttons or the date field when selecting a result from the From/To autocompletes. This is probably sth that needs fixing upstream in [bits-ui](https://github.com/huntabyte/bits-ui/tree/next), but I didn't yet manage to isolate the issue and spent far too much time on this already...

Lmk if you think that it makes sense to backport the changes from #705 and this PR to `prima-mobile` (since I still don't see any nice, straightforward way to share the components...)

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -57,6 +57,7 @@
 			"resolved": "https://registry.npmjs.org/@alloc/quick-lru/-/quick-lru-5.2.0.tgz",
 			"integrity": "sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==",
 			"dev": true,
+			"license": "MIT",
 			"engines": {
 				"node": ">=10"
 			},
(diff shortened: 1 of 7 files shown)


felixguendling:
> checking GTFS-provided colors for sufficient contrast

Yes, that's out of scope. I would prefer if checks for this were added to GTFS validator tools. This is nothing we should try to fix in MOTIS or the UI as we're too late in the pipeline.


> upstream in [bits-ui](https://github.com/huntabyte/bits-ui/tree/next), but I didn't yet manage to isolate the issue and spent far too much time on this already.

If it's easier to just roll our own components in some cases, it's fine to kick out bits-ui. It just comes in handy if it's exactly matching the use case. If it's not, it can become a burden instead of helping us.


> Lmk if you think that it makes sense to backport the changes from #705 and this PR to `prima-mobile` (since I still don't see any nice, straightforward way to share the components...)

That would be amazing! :)

PR #713, opened 2025-01-22
Improve formatDuration

### LINKED ISSUE #712: Leg duration display 0 min
When the leg duration is 0 min, motis displays simply empty parentheses. Wouldn't it be better to make it clear and display "(0 min)" ?
See the following screenshot:

Image

See:
https://github.com/motis-project/motis/blob/6a00abe08f674c2c40dc21fa2d123b8b0fa2eaf5/ui/src/lib/formatDuration.ts#L1-L8

Altonss:
If the duration is 0 min, display "0 min" instead of nothing.
Should fix #712


felixguendling:
Thank you! :)

PR #732, opened 2025-02-10
Prolong default `http_timeout`

traines-source:
As discussed [here](https://github.com/public-transport/transitous/pull/871) and [here](https://github.com/public-transport/transitous/pull/857), the default timeout for RT downloads of 10 seconds may be too short for large feeds, because it seems to encompass the entire transfer and not only until the first packet. As far as I understand, it can be increased without risking exceeding the 60 seconds interval because the downloads run in parallel. Increasing the timeout to 30 seconds has helped for Transitous.

For some reason, it does not seem to throw a timeout error, but returns some garbled protobuf out of which it even [succeeds to read a very small number of trip updates](https://github.com/public-transport/transitous/pull/857#issuecomment-2645831606). When testing locally, I do get a timeout error as expected.

Btw, I'm seeing in the logs of my DELFI RT feed mirror that some instance running at TU Darmstadt is also frequently hitting the 10s timeout. Would be interesting to know whether it also doesn't throw timeout errors.

(... and a typo fix)


felixguendling:
maybe the config to string test can be deleted?

traines-source:
I think it suffices for me to notice that such a test exists :) Ah and I see that the msvc failure was probably fixed with https://github.com/felixguendling/cista/commit/e03a1ff0a84d3f638bf4bff7357d19e542640288.

PR #734, opened 2025-02-12
ODM Interval Extension

mority:
Addresses the issue of different time intervals being searched by the subqueries of the ODM `meta_router` by employing a two-step routing process.
1. Run all subqueries (previous behavior)
2. Determine the largest interval searched by the subqueries in step 1. Re-run the ones that searched a smaller interval. Use the max. interval as new search interval for these subqueries and forbid interval extension.


src/odm/prima.cc
```diff
@@ -11,6 +11,20 @@
 #include "nigiri/common/parse_time.h"
 #include "nigiri/timetable.h"
 
+#include "motis/odm/odm.h"
+#include "motis/odm/prima.h"
+
+#include <ranges>
```
felixguendling, line 17: include order

src/odm/meta_router.cc
```diff
@@ -17,6 +17,7 @@
 #include "boost/thread/tss.hpp"
 
 #include "utl/erase_duplicates.h"
+#include "utl/for_each_bit_set.h"
```
felixguendling, line 20: maybe I missed it but it doesn't seem used?

felixguendling:
> Determine the largest interval searched by the subqueries in step 1. Re-run the ones that searched a smaller interval. Use the max. interval as new search interval for these subqueries and forbid interval extension.

Seems quite complex. Why not just search 24h and that's it? Empty? Fine. Not empty? Fine.

mority:
Starting with the same, fixed 24-h search interval for all subqueries outperforms this approach. #736

Issue #703, opened 2025-01-18
plan: START and END place seems to be reversed when searching with arriveBy=true


mlundblad:
When doing a coordinate-based query setting arriveBy=true, it looks like the start coordinate place is labelled END and the arrival is labelled START:

Sample query:

https://api.transitous.org/api/v1/plan?fromPlace=50.849516050000005%2C4.367312985278861&toPlace=50.8140554%2C4.3841294&arriveBy=true&time=2025-02-01T07%3A30%3A00Z

![Image](https://github.com/user-attachments/assets/72a1e9f1-19c0-4c06-af4f-ed97776e8e0f)

felixguendling:
True. The stop name here is more an implementation detail.
It's labeled by where from where to where the search goes.
The search starts at the destination (`START`) and ends at the origin (`END`).
In the end, I would expect the client to replace the name of the first+last stop with whatever the user entered, no matter what the name has been before (calling some kind of reverse-geocoder in case of coordinates).

mlundblad:
> True. The stop name here is more an implementation detail. It's labeled by where from where to where the search goes. The search starts at the destination (`START`) and ends at the origin (`END`). In the end, I would expect the client to replace the name of the first+last stop with whatever the user entered, no matter what the name has been before (calling some kind of reverse-geocoder in case of coordinates).

Yeah, I noticed this when checking some "arrive by" routes in GNOME Maps. In the normal case we substitute with the original query name (so it would be the searched name, or the raw coordinates if the user clicked on the context menu "Route from here/to here".

I think probably we should just change the logic so it replaces either START or END for "from" in the first leg (when non-transit) and for "to" in the last leg (when non-transit).

PR #738, opened 2025-02-14
ODM direct rides that start or end at a station

mority:
fixes ODM direct rides when to or from is a station, i.e., tt_location

### DIFF
--- a/include/motis/odm/meta_router.h
+++ b/include/motis/odm/meta_router.h
@@ -67,6 +67,7 @@ private:
       nigiri::interval<nigiri::unixtime_t> const&) const;
   std::vector<routing_result> search_interval(
       std::vector<nigiri::routing::query> const&) const;
+  void add_direct() const;
 
   ep::routing const& r_;
   api::plan_params const& query_;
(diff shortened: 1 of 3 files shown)


felixguendling:
this seems to do it as proposed here but it doesn't apply to the normal routing endpoint?
https://github.com/motis-project/motis/issues/703

mority:
> this seems to do it as proposed here but it doesn't apply to the normal routing endpoint? #703

No, it only changes the way how direct rides are represented as nigiri journeys. Before this PR, both start and destination of the direct-ride journey were set to `nigiri::get_special_station(nigiri::special_station::kStart)` and `nigiri::get_special_station(nigiri::special_station::kEnd)`, respectively. But this led to a fatal bug in
https://github.com/motis-project/motis/blob/43840a2d3f1a8d609122d74be59199c54a020821/src/place.cc#L101
When the direct-ride journeys were translated to API responses, the start of the journey leg being `nigiri::special_station::kStart` would cause `std::get<osr::location>(start)` to get called. However, the `start` of the journey would not be an `osr::location`.

PR #740, opened 2025-02-14
Resolve intermodal dummies

mority:
replaces `nigiri`'s intermodal dummy stations with coordinates or reverse geocoding if available
#703

### DIFF
--- a/include/motis/endpoints/routing.h
+++ b/include/motis/endpoints/routing.h
@@ -84,6 +84,8 @@ struct routing {
       std::chrono::seconds max) const;
 
   config const& config_;
+  adr::typeahead const& t_;
+  adr::reverse const* r_;
   osr::ways const* w_;
   osr::lookup const* l_;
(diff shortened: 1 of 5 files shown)


felixguendling:
I'm not sure if the reverse-geocoding is really up to this task in the current status. It contains stuff like ATMs (or basically anything that has a `name` tag in OSM) which is not really useful.

I would expect the client-side UI to fill in something that's meaningful to the user. For example the content of the search fields from and to. But since those are not transmitted to MOTIS, I think it's fine to fill in `START` and `END`. The problem with #703 is more that they are reversed in case `arriveBy` is set to `true`. But fixing this should be easy to add as a last step in `journey_to_response`.

mority:
Oh, okay. Reverse geocoding only seemed to produce street addresses but I only tested a dozen queries or so. I guess, we close this for now and leave this responsibility to the UI as you suggested.

Issue #719, opened 2025-01-23
Route display bug


Altonss:
Brussel-Zuid to VUB on api.transitous.org:

![Image](https://github.com/user-attachments/assets/9f9ee7bc-04b4-4af4-a2a1-8a48f182f3a4)

![Image](https://github.com/user-attachments/assets/affe057f-3ba5-494c-b833-39fe869bf39b)

felixguendling:
https://github.com/motis-project/motis/issues/706#issuecomment-2608464611

This is what needs to be done to activate the feature in #707 which is currently not activated in the UI nor in the transitous.org configuration. So I doubt that it's fixed by #707. Maybe we can have an eye on it and reopen if it happens again.

PR #741, opened 2025-02-15
UI fixes

### LINKED ISSUE #737: ui mobile view: exit connection details = new search
Desktop = no new search
Mobile = new search

traines-source:
* fixes #737 
* fixes `NaN` display part of #719 (in cases where [dummy_itinerary](https://github.com/motis-project/motis/blob/34e33b9534101f956c18b5e118d265b1bf5d5692/src/street_routing.cc#L248) is used for the initial/final walking legs – happening often when starting from coordinates)
* and other things

### DIFF
--- a/ui/src/app.css
+++ b/ui/src/app.css
@@ -109,3 +109,7 @@
 .maplibregl-ctrl-group .maplibregl-ctrl-geolocate {
 	display: none;
 }
+
+.hide {
+	display: none;
+}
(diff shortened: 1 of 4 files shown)


felixguendling:
Nice! :)

ui/src/app.css
```diff
@@ -109,3 +109,7 @@
 .maplibregl-ctrl-group .maplibregl-ctrl-geolocate {
 	display: none;
 }
+
+.hide {
+	display: none;
```
felixguendling, line 114: tailwind has already `hidden` which should do the same?

traines-source: Ah yes, would have to verify whether it works with dynamically added classes, will change it in some future UI PR then.

PR #746, opened 2025-02-17
Update to osr with elevation profile support

MichaelKutzner:
This updates to the latest `osr` version. No elevation data will be calculated or used, yet.


felixguendling (commented):
I usually only use [[fallthrough]] once to mark the point until it falls through
https://github.com/motis-project/nigiri/blob/4c925185fcabca0b3a154d5f5d2b212b536013ec/src/clasz.cc#L28

MichaelKutzner:
> I usually only use [[fallthrough]] once to mark the point until it falls through https://github.com/motis-project/nigiri/blob/4c925185fcabca0b3a154d5f5d2b212b536013ec/src/clasz.cc#L28

I didn't know about this convention. I have updated it.

PR #747, opened 2025-02-17
swap START/END when arriveBy is set

### LINKED ISSUE #703: plan: START and END place seems to be reversed when searching with arriveBy=true

mority:
closes #703

### DIFF
--- a/src/journey_to_response.cc
+++ b/src/journey_to_response.cc
@@ -43,6 +43,15 @@ api::ModeEnum to_mode(osr::search_profile const m) {
   std::unreachable();
 }
 
+void cleanup_intermodal(api::Itinerary& i) {
+  if (i.legs_.front().from_.name_ == "END") {
+    i.legs_.front().from_.name_ = "START";
+  }
+  if (i.legs_.back().to_.name_ == "START") {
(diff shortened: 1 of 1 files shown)


vkrause:
Are we sure this is correct, the MOTIS web ui now shows journeys in arrival mode starting at "END" at arriving at "START"?

felixguendling:
True. Not sure why. The code looks ok. @mority maybe you can have a look what's happening?

mority:
I'll look into it.

felixguendling:
We have the latest version deployed [here](https://prima-motis.motis-project.de/) and the issue seems to be solved here. Maybe api.transitous.org doesn't run the latest version?

vkrause:
Indeed, seems fixed on the Transitous instance as well now.

PR #750, opened 2025-02-18
Add handling for invalid commands in main.cc

fbarbe00:
**This is a minor change**
### Description
This PR improves the user experience by adding handling for invalid commands in  `exe/main.cc`. When a user enters an unrecognized command, the program will now print "Invalid command. Type motis --help for a list of commands."

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -131,6 +131,10 @@ int main(int ac, char** av) {
         return 1;
       }
     }
+
+    default:
+      fmt::println("Invalid command. Type motis --help for a list of commands.");
+      return 1;
   }
 
   google::protobuf::ShutdownProtobufLibrary();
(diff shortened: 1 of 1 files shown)


felixguendling:
The protobuf shutdown was placed wrong it seems. Maybe remove it for now. Or save the return value in a variable so you can return it after protobuf shutdown.

fbarbe00:
Hi @felixguendling , I have added a commit that introduces a new variable to return after the protobuf shutdown

exe/main.cc
```diff
@@ -55,12 +55,14 @@ int main(int ac, char** av) {
   --ac;
   ++av;
 
+  int return_value = 0;
```
felixguendling, line 58:
```suggestion
  auto return_value = 0;
```

AAA style (see https://github.com/motis-project/motis/blob/master/docs/STYLE.md)

fbarbe00:
My bad! Fixed.

felixguendling:
thank you! :partying_face:

PR #755, opened 2025-02-20
Check sizes of intervals in blacklisting/whitelisting responses

mority:


src/odm/prima.cc
```diff
@@ -108,7 +108,14 @@ std::size_t prima::n_events() const {
   return from_rides_.size() + to_rides_.size() + direct_rides_.size();
 }
 
+std::size_t n_pt_updates(json::array const& update) {
+  return std::accumulate(
+      update.begin(), update.end(), 0U,
```
felixguendling, line 113:
```suggestion
      update.begin(), update.end(), std::size_t{0U},
```

vielleicht so? :man_shrugging:

PR #769, opened 2025-03-06
(Reverse) Geocoding: Filter out unwanted types

### LINKED ISSUE #577: Reverse Geocoding: Filter by Type
e.g. only stop places

HerrLevin:
Fixes #577 

Needs: https://github.com/triptix-tech/adr/pull/12


openapi.yaml
```diff
@@ -89,6 +89,16 @@ paths:
           description: latitude, longitude in degrees
           schema:
             type: string
+        - name: filterType
```
CommanderStorm, line 92: (not a maintainer)

I think for both of these usages, something like `type` would be more in-line with the other API-parameters

openapi.yaml
```diff
@@ -89,6 +89,16 @@ paths:
...
+          in: query
+          required: false
+          description: |
+            Optional. Default is all types.
+            
+            Only return results of the given types.
```
CommanderStorm, line 98: (not a maintainer)

Looking at the code, this is likely a filter which can only be applied once => can only return one type

There is an occurance of this below too.

```suggestion
            Only return results of the given type.
```

PR #770, opened 2025-03-07
Fix reading GBFSv3 feeds with a null global_rules key

vkrause:
We only checked for the presence of the key here, not for its type.

The Finnish national aggregate feed has the key set, but with value null, which caused the entire feed to be discarded.


felixguendling:
Thanks! :-)

PR #771, opened 2025-03-08
UI URL, START/END, dates, no WebGL

### LINKED ISSUE #725: [UI] No indication of date change when querying earlier trips
On a relation without a lot of daily trips, querying earlier trips might display connections from the day before without any warning/indication.
This can be highly misleading and lead to errors for the user.

traines-source:
* prima backports, closes #725, START/END filling as per #740 
* if we agree that we restrict ourselves to enable linking to a search and not to specific results/itineraries, closes https://github.com/public-transport/transitous/issues/899 (with a very ugly/long URL based on the existing JSON.stringified from/to params)
* preserves debug/dark/motis url params
* support for viewing UI without webgl (also used in mobile mode)

### DIFF
--- a/ui/src/app.css
+++ b/ui/src/app.css
@@ -74,6 +74,14 @@
 	}
 }
 
+body {
+	font:
+		12px / 20px Helvetica Neue,
+		Arial,
+		Helvetica,
+		sans-serif;
+}
+
 .maplibregl-popup-anchor-top .maplibregl-popup-tip,
 .maplibregl-popup-anchor-top-left .maplibregl-popup-tip,
 .maplibregl-popup-anchor-top-right .maplibregl-popup-tip {
--- a/ui/src/lib/ConnectionDetail.svelte
+++ b/ui/src/lib/ConnectionDetail.svelte
@@ -32,6 +32,7 @@
 	<Time
 		variant="schedule"
 		class="font-semibold w-16"
+		queriedTime={timestamp}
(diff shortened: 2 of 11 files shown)


ui/src/app.css
```diff
@@ -74,6 +74,14 @@
 	}
 }
 
+body {
```
felixguendling, line 77: The default with tailwind seems to be:

```css
html, :host {
font-family: ui-sans-serif,system-ui,sans-serif,"Apple Color Emoji","Segoe UI Emoji",Segoe UI Symbol,"Noto Color Emoji";
 //...
}
```

What's wrong with this?

traines-source: Since the entire UI is/was a maplibregl control, [styles from there are applied](https://github.com/maplibre/maplibre-gl-js/blob/d42e8dbfd50a49b00d7fe4768472d33ebf8df1f7/src/css/maplibre-gl.css#L2). So in order to make it look the same without/outside maplibregl, I needed to apply this manually. Another option would be to override the maplibregl styles with the tailwind styles (i.e. looks different than before) or applying the `maplibregl-map` class at the right places outside of the actual map.

felixguendling: The tailwind font family list looks like they put more effort into this. So maybe we can use this?

felixguendling:
I merge it. I guess the font family thing is a minor nitpick that can be fixed later on if necessary.

PR #778, opened 2025-03-14
Docs: Clone via SSH without passphrase

mority:


jbruechert:
I wonder if we should change pkg or at least the cmake integration here.
The ssh cloning is only useful if you want to push to the repository again, right? 

Apparently you can set a different url for pushing like so: `git remote set-url --push origin git@github.com:USERNAME/REPOSITORY`, I feel like using that by default instead of full ssh in pkg would make this a lot easier.

felixguendling:
In case we clone with HTTPS and push with SSH (which makes sense), we should always make sure that there are no Git Submodules that require SSH.

Currently, this seems to be fine:

```bash
➜  motis git:(master) find . -name .gitmodules | xargs -I {} grep "url" {}
	url = https://github.com/google/boringssl
	url = https://github.com/litespeedtech/lsquic
	url = https://github.com/litespeedtech/ls-qpack
	url = https://github.com/litespeedtech/ls-hpack
	url = https://github.com/uNetworking/uSockets.git
url = https://github.com/jupp0r/prometheus-cpp
url = https://github.com/Microsoft/vcpkg
url = https://github.com/microsoft/GSL
url = https://github.com/google/googletest
url = https://github.com/google/benchmark
url = https://github.com/open-telemetry/opentelemetry-proto
url = https://github.com/nlohmann/json
url = https://github.com/opentracing/opentracing-cpp.git
	url = https://github.com/google/googletest.git
	url = https://github.com/civetweb/civetweb.git
	url = https://github.com/abseil/googletest.git
	url = https://github.com/microsoft/vcpkg.git
	url = https://github.com/google/googletest.git
	url = https://github.com/open-source-parsers/jsoncpp.git
```

felixguendling:
Merging this for now because this disclaimer is correct at the moment.
We can remove it once pkg uses HTTP for pull and SSH for push as described by @jbruechert .

PR #780, opened 2025-03-16
osr: switch to sequential import for now

felixguendling:


felixguendling:
Unfortunately we had a bug in the parallel implementation using TBB. For now to fix the bug immediately, all parallel parsing has been removed probably making the import extremely slow. However, a slow import is still better than having bugs.

PR #781, opened 2025-03-17
docs: one-to-many does route, not geocode

CommanderStorm:
Hi,
First of all, thanks for the awsome work done here.

I think I spotted a typo in the docs.
This API should likely be in the routing tab, not the geocoding tab.

image


felixguendling:
Thank you! :)

PR #783, opened 2025-03-19
Add One-to-All API

MichaelKutzner:
This adds an API endpoint for One-to-All search.

Notice that the code currently uses a development branch and therefore cannot be merged yet.


openapi.yaml
```diff
@@ -1278,6 +1420,34 @@ components:
...
+      description: Place reachable by One-to-All
+      type: object
+      properties:
+        place:
+          $ref: "#/components/schemas/Place"
+          description: Place reached by One-to-All
+        duration:
+          type: integer
+          description: Total travel duration
+        connection_count:
```
felixguendling, line 1433: alles camelCase in der API

src/endpoints/one_to_all.cc
```diff
@@ -0,0 +1,138 @@
...
+                             query.additionalTransferTime_ == 0 &&
+                             query.transferTimeFactor_ == 1.0),
+                .min_transfer_time_ = n::duration_t{query.minTransferTime_},
+                .additional_time_ =
+                    n::duration_t{query.additionalTransferTime_},
+                .factor_ = static_cast<float>(query.transferTimeFactor_)},
+    };
+  };
+
+  auto const make_place = [&](n::location_idx_t const l, n::unixtime_t const t,
```
felixguendling, line 70: make use of place.h to_place so it's only one code location to change in case we want extra infos

src/endpoints/one_to_all.cc
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  auto const unreachable = query.arriveBy_
+                               ? n::kInvalidDelta<n::direction::kBackward>
+                               : n::kInvalidDelta<n::direction::kForward>;
+  auto const rtt = rt_->rtt_.get();
+
+  auto const make_query = [&](n::location_idx_t const l,
+                              n::unixtime_t const time) {
+    return n::routing::query{
```
felixguendling, line 42: only used once -> no function

openapi.yaml
```diff
@@ -14,6 +14,148 @@ externalDocs:
 servers:
   - url: https://europe.motis-project.de
 paths:
+  /api/v1/one-to-all:
```
felixguendling, line 17: we should not version the first implementation so changes are possible without supporting all development versions at the same time

MichaelKutzner: Sounds reasonable, I will change it.
Would `/api/one-to-all` work? Or should it's experimental state be mentioned, e.g. `/api/experimental/one-to-all`? Or something else?

felixguendling: Everything with out a version `v*` is experimental / might change any time without prior notice. So no version is enough. Once we see that this API works for data consumer software, we can declare it stable and version it.

felixguendling: Not sure. Until now the only idea for ordering API endpoints was to put APIs with the same tags together (which is the case here I guess).

MichaelKutzner: The current endpoints with tag:

```
/api/v1/one-to-many: routing
/api/v1/reverse-geocode: geocode
/api/v1/geocode: geocode
/api/v1/trip: timetable
/api/v1/stoptimes: timetable
/api/v1/plan: routing
/api/experimental/one-to-all: routing
/api/v1/map/trips: map
/api/v1/map/initial: map
/api/v1/map/stops: map
/api/v1/map/levels: map
/api/debug/footpaths: debug
```

I would probably prefer all `routing` endpoints at top, but a different order could be reasonable as well.

src/endpoints/one_to_all.cc
MichaelKutzner: Thank you for mentioning it. That also solved the issue I had with some names.

src/endpoints/one_to_all.cc
MichaelKutzner: Changed it. I also made `make_all()` a block, instead of a function.

openapi.yaml
felixguendling: Please press "resolve conversation" if you addressed the issue someone else brought up.

include/motis/endpoints/one_to_all.h
```diff
@@ -0,0 +1,24 @@
...
+#include "boost/url/url_view.hpp"
+
+#include "motis-api/motis-api.h"
+#include "motis/data.h"
+#include "motis/fwd.h"
+
+namespace motis::ep {
+
+struct one_to_all {
+  enum class dir_t { kArrival, kDeparture };
```
felixguendling, line 12: always reuse, don't redefine in each file
nigiri::event_type
or nigiri::direction

openapi.yaml
```diff
@@ -1278,6 +1420,34 @@ components:
...
+          description: Place reached by One-to-All
+        duration:
+          type: integer
+          description: Total travel duration
+        connectionCount:
+          type: integer
+          description: Number of connections used
+
+    Reachable:
+      description: Object all reachable places by One-to-All search
```
felixguendling, line 1438: to object = ablehnen?

openapi.yaml
```diff
@@ -1278,6 +1420,34 @@ components:
...
+      type: object
+      properties:
+        one:
+          $ref: "#/components/schemas/Place"
+          description: One location used in One-to-All search
+        all:
+          description: |
+            List of locations reachable by One-to-All
```
felixguendling, line 1446: for short one line descriptions don't use '|' notation

src/endpoints/one_to_all.cc
```diff
@@ -0,0 +1,117 @@
...
+
+  auto const unreachable = query.arriveBy_
+                               ? n::kInvalidDelta<n::direction::kBackward>
+                               : n::kInvalidDelta<n::direction::kForward>;
+  auto const rtt = rt_->rtt_.get();
+
+  auto const make_place = [&](n::location_idx_t const l, n::unixtime_t const t,
+                              dir_t const dir) {
+    auto place = to_place(&tt_, &tags_, w_, pl_, matches_, tt_location{l},
+                          place_t{}, place_t{});
```
felixguendling, line 43: omit -> use default parameter values

src/endpoints/one_to_all.cc
```diff
@@ -0,0 +1,117 @@
...
+              .factor_ = static_cast<float>(query.transferTimeFactor_)},
+  };
+
+  auto const state =
+      query.arriveBy_
+          ? n::routing::one_to_all<n::direction::kBackward>(tt_, rtt, q)
+          : n::routing::one_to_all<n::direction::kForward>(tt_, rtt, q);
+
+  auto all = std::vector<api::ReachablePlace>{};
+  {
```
felixguendling, line 88: why block?

src/endpoints/one_to_all.cc
```diff
@@ -0,0 +1,117 @@
...
+    auto const dir = query.arriveBy_ ? dir_t::kDeparture : dir_t::kArrival;
+    for (auto i = n::location_idx_t{0U}; i < tt_.n_locations(); ++i) {
+      if (state.get_best<0>()[to_idx(i)][0] == unreachable) {
+        continue;
+      }
+
+      auto const fastest = query.arriveBy_
+                               ? n::routing::get_fastest_one_to_all_offsets<
+                                     n::direction::kBackward>(
```
felixguendling, line 97: maybe this should be a normal function parameter, not a template parameter?

template parameter if we need to code-gen two versions for performance reasons
since this contains a conditional jump anyway, there's no benefit, just complicated API

include/motis/endpoints/one_to_all.h
MichaelKutzner: `nigiri::event_type` solves the problem, thank you.

src/endpoints/one_to_all.cc
MichaelKutzner: Sorry, I somehow missed the default values.

openapi.yaml
MichaelKutzner: Sorry about that, should have been `object containing`.

src/endpoints/one_to_all.cc
MichaelKutzner: Sounds fine to me. I will create a Pull Request for it.

MichaelKutzner: Code has been updated

openapi.yaml
```diff
@@ -1278,6 +1420,34 @@ components:
...
+      description: Place reachable by One-to-All
+      type: object
+      properties:
+        place:
+          $ref: "#/components/schemas/Place"
+          description: Place reached by One-to-All
+        duration:
+          type: integer
+          description: Total travel duration
+        connectionCount:
```
felixguendling, line 1433: what's a "connection"? Does this require more explanation?
Even the code has more explanation. Why not have the same name?
(either both times `k` or both times `connectionCount` but not 2 names for the same concept if we can avoid it)

PR #786, opened 2025-03-20
Add hard coded One-to-All limits

MichaelKutzner:
This will add some hard coded limits for One-to-All searches, as requested in #783.

If preferred, I can also make them configurable, for example:

```yaml
limits:
  one_to_all:
    max_travel_time: 90
    max_results: 65535
```

In this case, there could also be a setting to disable this endpoint.

The current limits will probably need some adjustments, but I don't think, we can use a much lower results limit, if we want to support requests with up to 1 hour travel time.


felixguendling:
if limits and enabled/disabled endpoints are part of config.yml then all limits and all endpoints enabled/disabled should become configurable

PR #792, opened 2025-03-20
Focus on shorter ODM rides

mority:


mority:
Results in shorter ODM rides being used, as connections with these are no longer dominated by slightly shorter ones that use longer ODM rides.

image

The results take less taxi minutes and therefore are not removed by the current filter settings.

PR #807, opened 2025-04-01
Support for timetable attribution link

traines-source:
To finally properly attribute timetable sources...

Link header is somewhat of a standard according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link and https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#license. Alternative would be a custom header.

Sends header on all API requests and exposes it in the UI regardless if map is shown or not.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -62,9 +62,9 @@ int server(data d, config const& c, std::string_view const motis_version) {
   auto s = net::web_server{ioc};
   auto r = runner{server_config.n_threads_, 1024U};
   auto qr = net::query_router{net::fiber_exec{ioc, r.ch_}};
-  qr.add_header("Server", std::format("MOTIS {}", motis_version));
+  qr.add_header("Server", fmt::format("MOTIS {}", motis_version));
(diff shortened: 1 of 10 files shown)


src/server.cc
```diff
@@ -63,6 +63,10 @@ int server(data d, config const& c, std::string_view const motis_version) {
   auto r = runner{server_config.n_threads_, 1024U};
   auto qr = net::query_router{net::fiber_exec{ioc, r.ch_}};
   qr.add_header("Server", std::format("MOTIS {}", motis_version));
+  if (c.server_->data_attribution_link_) {
+    qr.add_header("Link", std::format("<{}>; rel=\"license\"",
```
felixguendling, line 67:
```suggestion
    qr.add_header("Link", fmt::format("<{}>; rel=\"license\"",
```

`#include <format>` is missing. But we're usually only using `fmt::` because it gets continuous improvements (unlike the `std::` version). So for consistency please switch to `fmt`.

PR #809, opened 2025-04-04
expose cancelled/in/out_allowed flags

traines-source:
In some sense this is a breaking change to the API. Alternatively one could only expose cancelled trips/stops when a flag is passed in the request.
(depends on https://github.com/motis-project/nigiri/pull/185)


ui/src/lib/StopTimes.svelte
```diff
@@ -87,24 +88,39 @@
...
+					<div class="flex items-center text-muted-foreground min-w-0">
+						<div><ArrowRight class="stroke-muted-foreground h-4 w-4" /></div>
+						<span class="ml-1 leading-tight text-ellipsis overflow-hidden">{stopTime.headsign}</span
+						>
+					</div>
+					{#if stopTime.pickupDropoffType == 'NOT_ALLOWED'}
+						<div class="flex items-center text-destructive text-sm">
+							<CircleX class="stroke-destructive h-4 w-4" />
+							<span class="ml-1 leading-none"
+								>{stopTime.cancelled
```
felixguendling, line 115: formatting

PR #810, opened 2025-04-08
Implement One-to-All routing for arbitrary locations

MichaelKutzner:
This adds support for arbitrary start or stop locations in One-to-All search.

I also added several test cases, to ensure nothing broke.
For the first leg, it works (mostly) as expected. On the final leg I encountered some strange behavior, that needs to be checked. Therefore some tests are currently disabled, as fixing this will not be part of this pull request.


MichaelKutzner:
I removed all inactive tests, which currently fail because of different reasons.
For now, these are added to a new branch: https://github.com/MichaelKutzner/motis/blob/one2all_tests/test/routing_test.cc
But it might be reasonable to merge them again in the future.
I also did some investigation and found an issue, that would be fixed with motis-project/nigiri#189.

PR #813, opened 2025-04-10
gtfs rt metrics prometheus endpoint

felixguendling:


traines-source:
image
Looking good!

felixguendling:
Thanks for checking! :)

derhuerst:
Wouldn't it make more sense if the "number of (un)successful updates in the feed" metrics were gauge ones?

traines-source:
I get your point, I think both have their advantages and disadvantages here, not really sure what fits better. But I would argue that knowing the polling interval, it's easier to get from a counter to a "gauge" with `increase()` than the other way round.

PR #827, opened 2025-04-23
Add elevation profiles to the API

MichaelKutzner:
This will add an option to select elevation profiles for routing. For now only `BIKE` mode will be supported. Other street modes will not use this option.

This change is for the API only. I intend to extend the web interface afterward.


openapi.yaml
```diff
@@ -180,6 +180,23 @@ paths:
...
             default: FOOT
 
+        - name: elevationCosts
+          in: query
+          required: false
+          description: |
+            Optional. Default is `NONE`.
+
+            Set an elevation cost profile, to penalize routes with incline.
+            This can cause locations to become unreachable, if an elevation is too high.
```
felixguendling, line 190: Explanation missing. I guess either remove this or explain what's happening in depth.

openapi.yaml
```diff
@@ -180,6 +180,23 @@ paths:
...
+            The profile is used for routing in transfers between transit connections,
```
felixguendling, line 191: > in transfers

That's not the case at the moment?

openapi.yaml
```diff
@@ -675,6 +692,20 @@ paths:
...
           schema:
             type: number
+        - name: elevationCosts
+          in: query
+          required: false
+          description: |
+            Optional. Default is `NONE`.
+
+            Set an elevation cost profile, to penalize routes with incline.
+            This can cause locations to become unreachable, if an elevation is too high.
```
felixguendling, line 702: see above

openapi.yaml
```diff
@@ -822,6 +853,24 @@ paths:
...
             default: FOOT
+
+        - name: elevationCosts
+          in: query
+          required: false
+          description: |
+            Optional. Default is `NONE`.
+
+            Set an elevation cost profile, to penalize routes with incline.
+            This can cause locations to become unreachable, if an elevation is too high.
```
felixguendling, line 864: see above

openapi.yaml
```diff
@@ -1439,6 +1488,14 @@ components:
           description: score according to the internal scoring system (the scoring algorithm might change in the future)
           type: number
 
+    ElevationCosts:
+      description: Different elevation cost profiles for street routing
```
felixguendling, line 1492: Maybe explain more about the costing in osr including an example?

src/import.cc
```diff
@@ -160,6 +160,27 @@ data import(config const& c, fs::path const& data_path, bool const write) {
     osm_hash.second = hash_file(*c.osm_);
   }
 
+  auto const elevation_dir =
+      c.get_street_routing()
+          .transform([](config::street_routing const& sr) {
+            return sr.elevation_data_dir_.value_or(fs::path{});
```
felixguendling, line 166: why 2x value_or?

docs/setup.md
```diff
@@ -77,7 +78,8 @@ gbfs:
   feeds:
     montreal:
       url: https://gbfs.velobixi.com/gbfs/gbfs.json
-street_routing: true              # enable street routing (default = false)
+street_routing:                   # enable street routing (default = false; Using boolean values true/false is supported for backward compatibility)
+  elevation_data_dir: srtm/       # folder which contains elevation data, e.g. SRTMGL1 data tiles in HGT format
```
felixguendling, line 82: Maybe provide your scripts for downloading SRTM data?

felixguendling: Maybe an extra HowTo for setting up elevation routing would help?

vkrause:
Looking at the API this only seems to deal with configuring the cost profile. Are there plans to also return elevation information for the paths in the result, so we can show this in client UI as well? At least "N meters elevation gained", but better would be data for a full elevation plot along the path.

openapi.yaml
MichaelKutzner: I added a more detailed explanation. Hopefully I didn't miss anything.
Same for the other locations.

openapi.yaml
MichaelKutzner: That's right, I didn't remove this after copying.

openapi.yaml
MichaelKutzner: Some explanation has been added. But I'm not sure, if that's enough for understanding. And if not, what a good example would be.

src/import.cc
MichaelKutzner: Changed it. I didn't notice the better version before.

docs/setup.md
MichaelKutzner: I added a brief HowTo about setting the data up.
Providing the scripts wouldn't help, as the download will require an account, which can be created for free. But I added the script to extract the downloaded files.

docs/elevation-setup.md
```diff
@@ -0,0 +1,76 @@
+# Setting up elevation tiles
+
+This page explains how to set up elevation tiles, that are required for elevation profiles.
+
+For performance reasons, all tile data must be stored uncompressed.
```
felixguendling, line 5: maybe add an estimate how much storage space is needed?

MichaelKutzner: Estimations have been added. I couldn't find my data about the additional disk space used after a world width import though. But I think it was about the same as `osr/way_osm_nodes_index.bin`. So not too bad. But it should be updated, once this has been checked again.

Sorry for the late reply. I didn't notice, the change was still pending.

PR #828, opened 2025-04-23
Adaptations for RT ADDED/NEW

traines-source:
TODO: 
* ~~update nigiri dep~~
* ~~change `nigiri_bin_ver`~~ done in alert PR


openapi.yaml
```diff
@@ -2171,6 +2171,11 @@ components:
         realTime:
           description: Whether there is real-time data about this leg
           type: boolean
+        scheduled:
```
felixguendling, line 2174: maybe an additional field for replacement would make sense to reference the original trip?

openapi.yaml
```diff
@@ -2171,6 +2171,11 @@ components:
...
+          description: |
+            Whether this leg was originally scheduled to run or is an additional service.
+            Scheduled times will equal realtime times in this case.
```
felixguendling, line 2177: for replacement, we would probably want to match stops to the original trip and give delays according to the scheduled times of this original trip?

src/tag_lookup.cc
```diff
@@ -80,13 +80,18 @@ std::string tag_lookup::id(nigiri::timetable const& tt,
     return fmt::format("{:%Y%m%d}_{:02}:{:02}_{}_{}", day, start_hours.count(),
                        start_minutes.count(), get_tag(src), id);
   } else {
-    // TODO support added trips
-    return "";
+    auto const id = s.fr_->id();
+    auto const time = std::chrono::time_point_cast<std::chrono::minutes>(
+        (*s.fr_)[0].time(n::event_type::kDep));
+    return fmt::format("{:%Y%m%d}_{:%H}:{:%M}_{}_{}", time, time, time,
```
felixguendling, line 86: we have this code now 3x (here, above, and in tag_lookup.cc)
maybe we can unify this to one location (maybe in tag_lookup.cc)?

openapi.yaml
felixguendling: as long as we don't have to change the API for this, that's fine :)

src/tag_lookup.cc
traines-source: Not sure what exactly you mean, the two formats? I think they're so different that I don't really see how to unify them.

openapi.yaml
traines-source: Same as above, and when supporting Replacements, I probably wouldn't do that either for simplicity (since in theory, replacements can have completely different stop sequences)

PR #831, opened 2025-04-24
Add Tab for direct departures search

HerrLevin:
I've added a simple layout to search for departures at a single station instead of clicking through the planner.

https://github.com/user-attachments/assets/02ec8fc4-f187-424f-8491-9422da3da3d8

### DIFF
--- a/ui/src/lib/StopTimes.svelte
+++ b/ui/src/lib/StopTimes.svelte
@@ -52,7 +52,7 @@
 			class="font-bold"
 			variant="outline"
 			onclick={() => {
-				onClickStop(stopId, stopName, queryTime, !arriveBy);
+				onClickStop(stopName, stopId, queryTime, !arriveBy);
 			}}
 		>
 			{#if arriveBy}
--- a/ui/src/lib/StopTimes.svelte
+++ b/ui/src/lib/StopTimes.svelte
@@ -16,7 +16,7 @@
 		stopName,
 		time: queryTime,
 		stopNameFromResponse = $bindable(),
-		arriveBy,
+		arriveBy
 	}: {
 		stopId: string;
 		stopName: string;
--- a/ui/src/lib/StopTimes.svelte
+++ b/ui/src/lib/StopTimes.svelte
@@ -9,21 +9,20 @@
 	import { Button } from '$lib/components/ui/button';
 	import { t } from '$lib/i18n/translation';
 	import type { RequestResult } from '@hey-api/client-fetch';
+	import { onClickStop, onClickTrip } from '$lib/utils';
 
 	let {
 		stopId,
+		stopName,
 		time: queryTime,
 		stopNameFromResponse = $bindable(),
 		arriveBy,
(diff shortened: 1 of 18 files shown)


felixguendling:
Thank you! This looks like a useful feature!
The CI isn't green. Can you please check with `npm run check` and `npm run lint` locally?
Regarding readability, I would prefer to have more contrast than dark blue on black.
Is there a particular reason you didn't use the [shadcn tabs component](https://next.shadcn-svelte.com/docs/components/tabs)? Until now it has saved us some lines of code and styling work to just use shadcn.

HerrLevin:
Fixed the ci issues. :)

I'll look into the shadcn tabs!

ui/src/lib/StopTimes.svelte
```diff
@@ -53,7 +52,7 @@
 			class="font-bold"
 			variant="outline"
 			onclick={() => {
-				setArriveBy(!arriveBy);
+				onClickStop(stopId, stopName, queryTime, !arriveBy);
```
traines-source, line 55: I think stopId and stopName arguments are in the wrong order, this also explains that I'm getting an error when switching to arrivals.

traines-source (changes requested):
Looking good in general, but
* I can't switch to arrivals anymore (error 500, see comment)
* mobile view is slightly destroyed due to fixed width
* on mobile, this further decreases the remaining space for the itinerary results, but I think this is something out of scope of this PR that needs to be improved/changed regardless
* I'm not 100% convinced of the tab name "Fahrplan/Timetable/Horaire". Maybe rather "Verbindungen/Connections/Itinéraires"? But that's just a suggestion.

HerrLevin:
Thanks for pointing this out! I thought I fixed the departures/arrivals bug already but apparently not. :D

- [x] fixed will only be applied after md breakpoint (Tried it with fractions but tailwind 3 is a bit limited here & this will give weird results. The fixed width seems to be the best solution above md. Below that `w-full` seems to do the trick)
- [x] fixed the departures/arrival thing
- [x] Changed the text to Connections (I like it!)

traines-source (approved):
Not sure why responsiveness wouldn't work similar to before, but it's good like that as well. Thanks :raised_hands:

ui/src/routes/+page.svelte
```diff
@@ -338,28 +291,41 @@
 	<LevelSelect {bounds} {zoom} bind:level />
 
 	<div class="maplibregl-control-container">
-		<div class="maplibregl-ctrl-top-left">
+		<div class="maplibregl-ctrl-top-left w-full">
```
networkException, line 294: Unfortunately this change causes the map to become non-interactive for the whole width of the page for any height range elements are visible on the left:

image

image

Hovering anywhere over the highlighted area changes the cursor to `default` instead of `grab` and the map cannot be interacted with.

traines-source: Thanks for noticing, I've fixed it in #837

PR #835, opened 2025-04-25
Remove dead api client code

HerrLevin:
In #831 I came across some code that's apparently not used anywhere. I don't know if this is left here intentionally or a remainder of the old api. I was able to run `npm run build` without any error so I'm assuming that this is really not needed. (I hope the CI doesn't say otherwise)


felixguendling:
thanks! :)

PR #837, opened 2025-04-27
direct departures ui fix

traines-source:
fix for #831


felixguendling:
@traines-source you should be able to create a release (the release should trigger the CI to upload build assets).

PR #838, opened 2025-04-28
Support elevation profile selection on the web interface

MichaelKutzner:
This will add a drop down menu to select an elevation profiles.

select options
example routing

Any opinions about the menu labels? Showing both intend and option seems redundant, but showing the option text only might be too brief.
Maybe I should add something like a mouseover or tooltip, to explain these options? Using a slider with some explanation might work as well, if it fits the style.
I'm also uncertain, if the icon is fitting. I couldn't find a better one, but I don't mind replacing or removing it.


felixguendling:
- Always use spacing. UI elements should never be "glued" to each other without spacing. Use the same spacing everywhere to make it look consistent.
- It's not clear that elevation cost only applies to bike. There's no way to select any profiles for first and last mile. So does this change even help the user?
- "Cost" means price for almost everyone (except maybe for people who write routers). Please use terminology that *everyone* can understand.
- Translation seems to be missing (mixed UI German + English)?
- "High costs for incline" doesn't make sense if no profile is selected

felixguendling:
30min for 4km seems a bit off. I can't remember Aachen to have mountains that high. Google Maps says 15min for 4km. So maybe 15-20min would be realistic?

MichaelKutzner:
> * Always use spacing. UI elements should never be "glued" to each other without spacing. Use the same spacing everywhere to make it look consistent.
> * It's not clear that elevation cost only applies to bike. There's no way to select any profiles for first and last mile. So does this change even help the user?
> * "Cost" means price for almost everyone (except maybe for people who write routers). Please use terminology that _everyone_ can understand.
> * Translation seems to be missing (mixed UI German + English)?
> * "High costs for incline" doesn't make sense if no profile is selected

Thank you for the feedback.
1. I will fix the spacing.
2. That was my point. The label itself seems too short. Having a hover text could add more details, without wasting space when not shown. I will check, if this is feasible.
3. That's a good point. I will update.
4. I haven't translated these texts yet, as they probably still need to change.
5. That would be the selected profile. But I will update the text, to make it more clear. Maybe I can also deactivate or hide the menu, if bike routing is disabled.

Regarding the example, it's not that bad when compared to the fastest route without elevation profile selected. Maybe something like 27 minutes would be more reasonable, but it cannot be faster than 24 minutes, without changing the velocity.

example with no profile

https://github.com/motis-project/osr/blob/21523b107d95191f5e705de2beec6ec47ea4b35d/include/osr/routing/profiles/bike.h#L189

MichaelKutzner:
Designs suggestions

MichaelKutzner:
I discarded the slider, as there are currently only 3 profiles available. As this would lead to large gaps between each option, using a drop down menu seems more suitable.

Furthermore, following changes to the layout have been made:

* There now should be always a margin of at least `2` between elements
* The menu will now only show up, if _Bike Carriage_ has been enabled
* There will be a tooltip, explaining the options intend
* Translations have been added.  Please check the French translation in particular, as I had to use a translator for this.

## Menu closed
Layout example (closed)

## Menu opened, with tooltip shown
Layout example (opened)

felixguendling (commented):
I think this UI has several problems.

- popup and dropdown doesn't work well together, IMO popups are a bad way to communicate stuff anyway because on mobile there's no hover and in general it's a hidden feature - not everyone hovers to see if there's a popover
- this option refers to bike carriage (meaning taking the bike from start to destination including trains) but the option is also relevant for bike (first mile, not restricting public transport to `bikes_allowed=true`)
- the UI in general is not really extensible. I would imagine maybe a drop down / radio button list for first mile and one for list mile. However, incline is probably an option that applies to both, so it should show up one a bike is selected in either of those
- I would prefer the use of shadcn. It seems to work and reduces the code a lot (see branch ui-test)

MichaelKutzner:
Thank you for the feedback, especially about the dropdown menu. I haven't tested it on a mobile device, so I wasn't aware of this problem.

ad 2) + 3) You are right, the incline option shouldn't be limited to bike carriage. But as I didn't see an option to select street modes for first and last mile, it's currently limited to that. Extending the logic, to also show this option, if bike routing is selected for the first or last mile, should be fairly easy though.
Depending on the the complexity about adding these inputs as well, should it be part of this Pull Request? Or a new one?

ad 4) I will have a look at that branch. Thank you for pointing that out.

MichaelKutzner:
About the menu for first and last mile, did you intend something like this?
Mode select for first and last mile

I added a menu for direct routing as well, although this causes the dialog to be a bit crowded. I also checked on mobile and it seems usable to me, both in portrait and landscape mode.

For comparison, I attempted to disable the elevation option, instead of hiding it. Additionally I tried using a row-wise layout, but I don't like, that the dropdown menu makes it difficult to notice, when the incline option becomes available. Additionally the alignment isn't right yet.
Example for row-wise layout

MichaelKutzner:
For comparison, these are some screenshots for both styles on mobile:

## Style 1
### Portrait mode
style 1 portrait mode
### Landscape mode
style 1 landscape mode

## Style 2
### Portrait mode
style 2 portrait mode
### Landscape mode
style 2 landscape mode

As the first style seems fine to me, I will go with that. Therefore the misaligned labels for style 2 haven't been fixed.

traines-source:
I feel both are fine, the one you chose first definitely has the advantage of being more compact, the other one better readability of state. Still, it probably makes it necessary to transform the Options panel into a popup, at least on mobile, or rework the scrolling behavior, because otherwise the routing results are barely visible when the Options panel is expanded (but maybe it's not as crucial that it has to be done in this PR).
Another thing that maybe needs to be checked is the interplay between Fahrradmitnahme (more precise label?) and the Dropdowns. I haven't looked at it more closely yet.

ui/src/routes/+page.svelte
```diff
@@ -162,12 +174,13 @@
...
-						preTransitModes: modes,
-						postTransitModes: modes,
-						directModes: modes,
+						preTransitModes: preTransitModes,
+						postTransitModes: postTransitModes,
+						directModes: requestDirectModes,
 						requireBikeTransport: bikeCarriage,
 						requireCarTransport: carCarriage,
 						transitModes: selectedTransitModes.length ? selectedTransitModes : undefined,
+						elevationCosts: elevationCosts,
```
felixguendling, line 183: use short hand for variable name == key name, don't repeat the value

i.e. `elevationCosts: elevationCosts` => `elevationCost`

ui/src/routes/+page.svelte
```diff
@@ -136,6 +137,10 @@
 			(urlParams?.get('selectedTransitModes')?.split(',') as Mode[])) ||
 			[]
 	);
+	let firstMileMode: Mode = $state((urlParams?.get('firstMileMode') ?? 'WALK') as Mode);
```
felixguendling, line 140: use `$state<Mode>` (also everywhere below)

ui/src/routes/+page.svelte
MichaelKutzner: I also changes some nearby occurrences.

ui/src/routes/+page.svelte
MichaelKutzner: Changes should have been applied at all occurrences.

traines-source:
A couple of things I noticed:
* Initial state of the mode dropdown is now empty instead of "all modes", but still behaves as all modes (probably because this is the backend default). I think the initial state should still be "all modes" with all modes also selected individually. I think Felix' intention was to also support selecting no modes (to just find direct walking etc.) by deselecting everything manually.
* when selecting "Fahrradmitnahme", BIKE is added to the pre/post/direct modes (which makes sense), but this is not reflected in the dropdowns below. I think this would be better to understand? Same for car carriage.
* intentional that the car option is missing for last mile?
* intentional that the inital state of the direct dropdown is empty, but still direct WALK is allowed by default as before (which I think is good)?
* intentional that only one option can be selected for first and last mile, but all for direct?
* in theory, for "Sharing-Fzg berücksichtigen", elevation preference could also be relevant – maybe out of scope since we don't want to start worrying about `pre/postTransitRentalFormFactors` for now
* at this point, even on desktop an Options popup would probably be better

Sorry if you discussed some of this already before or if I missed sth. Also maybe not all of this needs to be adressed immediately. And apart from that, it's looking really good, it will be great to have all these options finally in the UI :tada:

felixguendling:
> I think Felix' intention was to also support selecting no modes

Correct.


> when selecting "Fahrradmitnahme"

Maybe this has to disable everything else as it's incompatible with all kinds of other settings such as sharing mobility, car, walking (without bike), etc.


> intentional that the car option is missing for last mile?

I think getting picked up or emulating a taxi to the destination is something we should support.


> intentional that the inital state of the direct dropdown is empty

You're right. Probably it should reflect what the search will do by default (`WALK`).


> intentional that only one option can be selected for first and last mile, but all for direct?

That was my proposal. The idea was that

- usually the user already knows the preferred mode (I want to do park&ride / sharing mobility /etc.)
- most modes do not make sense to combine because for example car is always faster than bike is always faster than walk (except for special scenarios like one way streets, turn restrictions, etc.). So in the end, a search with all options enabled will still only deliver the fasted (as it's optimized in MOTIS now). This would only make sense if we would've a McRAPTOR that penalizes car vs other options.


> "Sharing-Fzg berücksichtigen", elevation preference could also be relevant – maybe out of scope

Yes, I would say this PR already took way longer than I expected. Let's try to merge it asap and move this to another discussion. I think this is more complex (e.g. with electric sharing vehicles you don't care much about incline, but without electric propulsion you do).


> Options popup would probably be better

As long as it only has "irrelevant" information that's fine. Everything important should not be hidden behind a popup where we have to hope that the user finds it by hovering.


Thank you for your feedback @traines-source - this was very helpful to improve the usability!

traines-source:
> Maybe this has to disable everything else as it's incompatible with all kinds of other settings such as sharing mobility, car, walking (without bike), etc.

I mean in theory there are scenarios, such as walking to the train station where you have parked your bike, taking your bike with your/a rental car, or other weird things like that. So I would just have reflected in the dropdowns the fact that bike is set for first/last mile as well. But yeah not that important.

> Options popup would probably be better

I meant for the entire Options panel, upon the "Options" button click, instead of expanding below, it should open a dialog. Because even on desktop you now have to manually close it again anyways in order to properly see any routing results. But yeah we can do that later.

MichaelKutzner:
Thank you for the feedback.

> > Maybe this has to disable everything else as it's incompatible with all kinds of other settings such as sharing mobility, car, walking (without bike), etc.
> 
> I mean in theory there are scenarios, such as walking to the train station where you have parked your bike, taking your bike with your/a rental car, or other weird things like that. So I would just have reflected in the dropdowns the fact that bike is set for first/last mile as well. But yeah not that important.
> 

That's something we just discussed internally. Disabling all options would prevent such use cases. I think, it would be more reasonable to set all modes to `BIKE`, if bike carriage is selected. Same for car.
However, if both are selected, it gets tricky. I also don't know, if people will like their settings being overridden, if they select bike carriage last. So maybe change these only, if the user hasn't changed them before?


> * Initial state of the mode dropdown is now empty instead of "all modes", but still behaves as all modes (probably because this is the backend default). I think the initial state should still be "all modes" with all modes also selected individually. I think Felix' intention was to also support selecting no modes (to just find direct walking etc.) by deselecting everything manually.

I will have a look at the default value.


> * intentional that the car option is missing for last mile?

I will change it to allow cars there as well.

> * intentional that the inital state of the direct dropdown is empty, but still direct WALK is allowed by default as before (which I think is good)?

So changing the text to show the default, if none is selected, should be fine? Or would you prefer a different behavior?

traines-source:
> So maybe change these only, if the user hasn't changed them before?

I would keep it as simple as possible, whatever that is (I would think it would be hard-updating the dropdown list to bike/car) I guess it shouldn't really be possible to select both bike and car carriage, but we also don't have to exclude all the whims a user might have :D

> So changing the text to show the default, if none is selected, should be fine?

Yes, just reflect the exact list that is sent to the backend as well?

felixguendling:
Looks like adding inputs for max*Time would also make sense:
https://github.com/motis-project/motis/issues/844

(best would be to get writ of them completely by having "smarter" routing logic that figures this out automatically - but that's where we are at the moment)

MichaelKutzner:
I changed to text to indicate when additional modes are used, because of bike carriage or car carriage have been enabled.
For me it feels too long, but I don't have any ideas how to better indicate why these modes are enabled. Suggestions how to improve this are therefore highly appreciated.

I haven't made any changes for the max*Time inputs, yet. For me, using `<input type="number"` seems fitting, but I cannot find any adequate bits-ui component.
If you have any suggestions what to use, I can add this. Otherwise we might make it a new pull request, so we can take some time to decide about the layout.

Some screenshots of the current state, all made on mobile:
Example for fallback WALK
Bike and car carriage enabled
Bike carriage enabled

ui/src/routes/+page.svelte
```diff
@@ -146,12 +151,19 @@
...
 		...(bikeCarriage ? ['BIKE'] : []),
 		...(carCarriage ? ['CAR'] : [])
 	] as Mode[]);
+	let preTransitModes = $derived([firstMileMode, ...additionalModes]);
+	let postTransitModes = $derived([lastMileMode, ...additionalModes]);
+	let requestDirectModes = $derived([
+		...directModes,
+		...additionalModes,
+		...(directModes.length == 0 && additionalModes.length == 0 ? ['WALK'] : [])
+	]);
```
traines-source, line 165: Frankly I would just have moved this logic from here to the selection of the carriage options, i.e. when bike carriage is selected, `firstMileMode`, `lastMileMode`, `directMode` are updated to `BIKE`, and by that, the dropdowns automatically display it and the exact same state is sent to the backend. That's what I meant.
Similarly, I'm not sure why there needs to be a distinction between "Zu Fuß" and "Vorgabe (Zu Fuß)", is there a good reason to not just set `directModes` to contain `WALK` by default, also enabling the user to completely deselect it?

But since you've now already implemented a more sophisticated solution, I personally am also fine with that. And yes, let's bother with `max*Time` in this PR.

(Deselecting all transit modes still doesn't work, but that can also be left to a future PR, albeit it's inconsistent with the empty state in the UI)

i.e. LGTM :+1:

traines-source (approved):

ui/src/routes/+page.svelte
MichaelKutzner: > Frankly I would just have moved this logic from here to the selection of the carriage options, i.e. when bike carriage is selected, `firstMileMode`, `lastMileMode`, `directMode` are updated to `BIKE`, and by that, the dropdowns automatically display it and the exact same state is sent to the backend. That's what I meant. Similarly, I'm not sure why there needs to be a distinction between "Zu Fuß" and "Vorgabe (Zu Fuß)", is there a good reason to not just set `directModes` to contain `WALK` by default, also enabling the user to completely deselect it?

Updating the modes when bike or car carriage is selected was my first idea. But I wasn't sure, if overwriting the user's settings is a good idea. Assume there is a use case, where first and last mile modes are selected first, before bike carriage is selected. But I'm fine with changing it accordingly, as this currently can result in setting multiple modes for the first and last mile.

About the "default" text: Maybe I misread your suggestion, but I kept the behavior unchanged and intended to make the actual request parameters obvious. If you prefer to allow queries without any direct connections allowed, it's easy to change.
For the currently long notation, e.g. "Vorgabe (Zu Fuß)": Maybe it would be better readable, if if would be something like "WALK*" instead. With a footnote like "* Automatically enabled by some settings".
This obviously isn't necessary, if bike and car carriage don't enable additional modes but change the selected items instead.

> (Deselecting all transit modes still doesn't work, but that can also be left to a future PR, albeit it's inconsistent with the empty state in the UI)

I haven't tested it much, as it wasn't intended as part of this pull request. I will make a note to check and fix it.

traines-source: > Assume there is a use case, where first and last mile modes are selected first, before bike carriage is selected.

Since the bike carriage toggle is above the dropdown, I think this would be fine/a problem of the user.

> For the currently long notation, e.g. "Vorgabe (Zu Fuß)"

I like the notation as it is, if we want to distinguish it. Of course I still think it would be easiest/most obvious to just reflect the exact list of request params in the dropdown (i.e. not distinguish it/"change the selected items instead").

felixguendling:
I will merge it now because it's already an improvement.

But maybe we have a future PR:

- to simplify as @traines-source described ("just reflect the exact list of request params in the dropdown)
- to disable first mile, last mile, direct for bike carriage and car carriage
- to prevent bike carriage to be combined with car carriage

Maybe we could use the [`Accordion`](https://next.shadcn-svelte.com/docs/components/accordion) widget with a radio button or something similar to have only one option enabled/visible at the same time:
- bike carriage
- car carriage
- or separate first mile, last mile, and direct mode

I assume that if you take your bike / car into public transport, that means you also want to go to the destination with it.

PR #841, opened 2025-04-29
Various edge cases

traines-source:
* the conditions in ConnectionDetail are getting completely out of hand...
* For polyline string encoding, coordinates with |lon| > 107.3741824 and our chosen precision 7 yield a 32-bit int overflow ([due to the additional left shift](https://developers.google.com/maps/documentation/utilities/polylinealgorithm)). Maybe due to that reason, we use an int64 there, but I have not found any (JS) implementation that correctly decodes that. From my perspective this case is not really defined. The only decoder https://polylinedecoder.online/ I've found that allegedly supports precision 7 wasn't able to decode it either. So either we make a breaking change to the API and lower the precision to 6 (which is still about 10cm precision) which probably avoids many errors in client implementations, or we keep precision 7 and write our own JS decoder.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1962,7 +1962,7 @@ components:
           description: Whether there is real-time data about this leg
           type: boolean
         polyline:
-          description: Google polyline encoded coordinate sequence (with precision 7) where the trip travels on this segment.
+          description: Google polyline encoded coordinate sequence (with precision 6) where the trip travels on this segment.
           type: string
 
     Direction:
@@ -1990,7 +1990,7 @@ components:
         - length
       properties:
         points:
-          description: The encoded points of the polyline using the Google polyline encoding with precision 7.
+          description: The encoded points of the polyline using the Google polyline encoding with precision 6.
           type: string
         length:
           description: The number of points in the string
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -467,8 +467,9 @@ api::plan_response routing::operator()(boost::urls::url_view const& url) const {
   auto const rtt = rt->rtt_.get();
   auto const e = rt_->e_.get();
   auto gbfs_rd = gbfs::gbfs_routing_data{w_, l_, gbfs_};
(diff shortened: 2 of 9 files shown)


felixguendling:
Would it make sense to increment the API version from v1 to v2 for all endpoints that use precision=7? Otherwise, clients will struggle with the switch?

traines-source:
Yes, if we decide to go that path. Although then we have to maintain two versions... One could argue that it's a minor enough change and that maybe not that many clients have implemented polylines. And considering that half the world has been wrong anyways, afterwards the entire world will be wrong (when continuing to assume 7 instead of 6).

jbruechert:
Also @mlundblad

traines-source:
AFAICT what will happen is that all coordinates are divided by ten (i.e. everything will end up somewhere around Africa), at least no crashes or sth. But yes maintaining two versions should be doable as well, probably the cleanest way.

Schmiddiii:
> would your apps / integrations break if we change this?

Railway would not break, we currently don't support polylines at all.

jbruechert:
I'm pretty sure this change would break GNOME maps and KDE Itinerary in some way or the other, they both use the polylines.

vkrause:
Itinerary would break with this, yes.

To give us more flexibility in the future, maybe add explicit properties for the precision/dimension parameters to the response? IIRC the Hafas API does that for example.

felixguendling:
> To give us more flexibility in the future, maybe add explicit properties for the precision/dimension parameters to the response? IIRC the Hafas API does that for example.

Sounds reasonable. :+1:

include/motis/endpoints/routing.h
```diff
@@ -96,7 +96,8 @@ struct routing {
       api::ElevationCostsEnum,
       std::chrono::seconds max,
       double max_matching_distance,
-      double fastest_direct_factor) const;
+      double fastest_direct_factor,
+      int api_version) const;
```
felixguendling, line 100:
```suggestion
      unsigned api_version) const;
```

PR #842, opened 2025-05-05
Fix for very short ODM direct routes

traines-source:
* Fix negative overflow when direct route is < `kODMOffsetMinImprovement`, leading to a 65476s max distance, leading to Dijkstra discovering the whole of Germany (and while doing that, permanently allocating 6 GB in every thread this runs on, since Dijkstra reset only clears cost vectors for reuse, but doesn't free)
* Using `n::log` for timestamp prefixing


felixguendling:
Oopsie!

PR #845, opened 2025-05-07
Routing metrics

traines-source:


include/motis/endpoints/one_to_all.h
```diff
@@ -5,6 +5,7 @@
 #include "motis-api/motis-api.h"
 #include "motis/data.h"
 #include "motis/fwd.h"
+#include "motis/metrics_registry.h"
```
felixguendling, line 8: now that we have a separate struct could we please hide all the details and just have a forward declaration in `fwd.h`? =)

traines-source: You mean like that?

Issue #455, opened 2024-02-27
MOTIS potential: compressed polylines


GerdC:
RE issue #452 Motis potential

There are multiple ways to compress polylines.

Often Googles compressed polyline format is used. It is efficient, proven, and there are libraries in many languages supporting it.

There are reasons to extend this format, like HERE did with flexible-polyline: https://github.com/heremaps/flexible-polyline (MIT license). Standard application of this format is to compress elevation for each coordinate in the polyline.

(ORS has another polyline format with elevation. I find that not well documented, and there appears to be not many implementations in different languages)

Also, routers like GraphHopper and OpenRouteService have an additional way to add properties to a polyline (more properties than would fit into HEREs flexible-polyline)

Their idea is the following. Let's take steepness as an example:
They return a structure that says: from polyline index 7 to index 13, the steepness id is 1, meaning a steepness between 4% and 7%.

Different modes have different properties that are interesting for them.
While for walking and bike, the steepness is interesting, for cars it is interesting if it is a toll road.
Examples for such properties are in this ORS documentation: https://giscience.github.io/openrouteservice/api-reference/endpoints/directions/extra-info/

GraphHoppers/ORSs idea can be combined with polyline and with flexible-polyline.

These are the options I came across using real world routing APIs.

felixguendling:
This depends on software that is not really part of MOTIS (only linked as a library or in the future used as third-party API). Since we might evolve the MOTIS API protocol, we can still keep this issue here.

mlundblad:
> RE issue #452 Motis potential
> 
> There are multiple ways to compress polylines.
> 
> Often Googles compressed polyline format is used. It is efficient, proven, and there are libraries in many languages supporting it.
> 

Is this the same as EPAF, right? This is what we currently support in GNOME Maps (the format used in GraphHopper and OpenTripPlanner).

GerdC:
Never heard of EPAF before, but this Gnome code implements EPAF and a few lines later links to Googlees polyline documentation: https://mail.gnome.org/archives/commits-list/2022-August/msg02657.html
So yes, EPAF should be identical with Googles polyline format.

mlundblad:
> Never heard of EPAF before, but this Gnome code implements EPAF and a few lines later links to Googles polyline documentation: https://mail.gnome.org/archives/commits-list/2022-August/msg02657.html So yes, EPAF should be identical with Googles polyline format.

[...]

PR #856, opened 2025-05-12
Add elevation data for routing path and steps

MichaelKutzner:
As requested by https://github.com/motis-project/motis/pull/827#issuecomment-2824074797

This will extend itineraries to contain the the maximum elevation up and down.
The response will contain the total elevation, as well as elevation per step. If no elevation data is available, these fields will be `0`.
As itineraries are only available for One-to-One queries, the feature will not support One-to-Many and One-to-All.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2036,6 +2036,8 @@ components:
         - exit
         - stayOn
         - area
+        - elevationUp
+        - elevationDown
       properties:
         relativeDirection:
           $ref: '#/components/schemas/Direction'
@@ -2072,6 +2074,12 @@ components:
             This step is on an open area, such as a plaza or train platform,
             and thus the directions should say something like "cross"
           type: boolean
+        elevationUp:
+          type: integer
+          description: The maximum numbers of meters routed uphill
+        elevationDown:
+          type: integer
+          description: The maximum numbers of meters routed downhill
 
     RentalFormFactor:
       type: string
@@ -2385,6 +2393,8 @@ components:
         - startTime
         - endTime
         - transfers
+        - elevationUp
+        - elevationDown
         - legs
       properties:
         duration:
@@ -2401,6 +2411,12 @@ components:
         transfers:
           type: integer
           description: The number of transfers this trip has.
+        elevationUp:
+          type: integer
+          description: The maximum numbers of meters routed uphill
+        elevationDown:
+          type: integer
+          description: The maximum numbers of meters routed downhill
         legs:
           description: Journey legs
           type: array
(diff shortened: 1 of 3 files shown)


felixguendling:
Is it possible to plot something like this?

https://developers.google.com/maps/documentation/javascript/examples/elevation-paths

image

(image source: https://www.bikemap.net/)

MichaelKutzner:
I need to have a look at this.
The visualization would be nice. But I don't know how this works, yet. I also think, we might need to make two graphs, if transits are included. At least for now. Otherwise the corresponding segments would currently be empty, which probably isn't helpful.
There is also a different problem, that we only store, very compressed, elevation differences per way. We therefore cannot show the elevation height.

felixguendling:
> We therefore cannot show the elevation height

We could just look it up in the raw data if it's still there? Lookups along the routed path should be fast.

@GerdC proposed to use another polyline format for this: https://github.com/motis-project/motis/issues/455

MichaelKutzner:
Currently the raw tile data is not accessible by the server. Changing this would be more complicated for than just passing the values already computed in `osr`.
But even then it's a bit more complicated, if we want to show the total elevation changes for an itinerary, as we don't have that data for transits. So it might be better to discard these fields for now. However, this would make the feature less useful for me at the moment.

As I intended to continue working on a different feature, I will leave it in this state for now. If anyone else is interested in implementing this feature, feel free to do so.

vkrause:
* Using 0 as "not available" marker is ambiguous, as that's also a valid value, could we differentiate this from the "flat" case somehow?
* Being able to show the elevation profile would IMHO be the ultimate goal here, but there's intermediate steps that also provide a lot of value already and sound like those are easier to reach:
  * elevation changes on transit are nice to have, but nowhere near as important as elevation changes on walk/wheelchair/bike/scooter sections.
  * absolute elevation values are also nice to have, but relative changes are what matters the most when deciding whether you could/want to make a walk/bike/etc section.

GerdC:
Advantage of HERE's flexible polyline format: encoders/decoders in many languages under MIT license. But do what you want.

MichaelKutzner:
> * Using 0 as "not available" marker is ambiguous, as that's also a valid value, could we differentiate this from the "flat" case somehow?

I don't think, that this is currently possible. If a way is totally flat, i.e. 0 meters up and down, no elevation data will be stored for that way. Same, if the elevation tiles provide no valid data for a way.
I think, for now the best option would be to check, if some elevation data has been loaded from disk. If no data is available, the field should not be used (if possible) or contain an invalid value. But this still wouldn't differentiate, if a way is flat or not covered by the provided tiles.

PR #857, opened 2025-05-12
Update street modes when carriage is enabled

MichaelKutzner:
Implementing change suggested at https://github.com/motis-project/motis/pull/838#discussion_r2083283396

Enabling bike carriage or car carriage will now update the selected modes for first and last mile. Additionally, bike or car will be enabled for direct routing.
This change should make the code much easier to read and maintain.

As I'm not sure, if it would be a good idea to allow users to deselect all modes for direct routing, it's now also prohibited to disable the last enabled mode. If this shouldn't be an issue, I will remove this check.


MichaelKutzner:
It's now possible to disable all direct modes.
I tested setting `directModes=` in the URL,  which will not result in an leading comma, if modes are added later on. I also made this change for `selectedTransitModes`, which should now work the same.
Additionally, while not directly related, using `selectedTransitModes=` will no longer fall back to the server default, allowing queries without any transits.

traines-source (approved):

traines-source:
Nice, thanks :raised_hands:

PR #858, opened 2025-05-12
Configurable endpoint limits

traines-source:
plus adaptations for import progress tag, depends on https://github.com/motis-project/nigiri/pull/198


src/endpoints/routing.cc
```diff
@@ -544,6 +544,14 @@ api::plan_response routing::operator()(boost::urls::url_view const& url) const {
     utl::verify(tt_ != nullptr && tags_ != nullptr,
                 "mode=TRANSIT requires timetable to be loaded");
 
+    auto const kMaxResults = config_.timetable_
```
felixguendling, line 547:
```suggestion
    auto const kMaxResults = config_.timetable_
```

`k*` variables are reserved for compile-time constants

src/endpoints/routing.cc
```diff
@@ -544,6 +544,23 @@ api::plan_response routing::operator()(boost::urls::url_view const& url) const {
...
+    auto const max_results = config_.timetable_
+                                 .and_then([](config::timetable const& x) {
+                                   return std::optional{x.plan_max_results_};
+                                 })
+                                 .value_or(256U);
+    utl::verify(query.numItineraries_ <= max_results,
+                "maximum number of minimum itineraries is {}", max_results);
+    auto const max_timeout = std::chrono::seconds{
+        config_.timetable_
+            .and_then([](config::timetable const& x) {
```
felixguendling, line 556: `utl::verify(tt_ != nullptr, ...)` implies that config_.timetable_ has to be set too.
If in doubt just add it to the `utl::verify` statement above and assume that it's valid here.

src/endpoints/one_to_all.cc
```diff
@@ -20,14 +20,17 @@ namespace motis::ep {
 
 namespace n = nigiri;
 
-constexpr auto const kMaxResults = 65535U;
-constexpr auto const kMaxTravelMinutes = 90U;
-
 api::Reachable one_to_all::operator()(boost::urls::url_view const& url) const {
+  auto const max_travel_minutes =
+      config_.timetable_
+          .and_then([](config::timetable const& x) {
```
felixguendling, line 26: `one_to_all` endpoint can only be instantiated if `tt_` is available (as it's a reference, not a pointer in the member variable). So `config_.timetable_` has to be set. If in doubt add a `utl::verify`.

src/endpoints/map/stops.cc
```diff
@@ -18,9 +18,14 @@ api::stops_response stops::operator()(boost::urls::url_view const& url) const {
   utl::verify(min.has_value(), "min not a coordinate: {}", query.min_);
   utl::verify(max.has_value(), "max not a coordinate: {}", query.max_);
   auto res = api::stops_response{};
-  auto n_items = 0U;
+
+  auto const max_results = config_.timetable_
```
felixguendling, line 22: see the other two remarks, `tt_` is a ref <=> config_.timetable_ should be set

PR #859, opened 2025-05-12
api: Use last stop as headsign if not provided in data

jbruechert:


src/journey_to_response.cc
```diff
@@ -341,7 +341,14 @@ api::Itinerary journey_to_response(
...
                   .scheduled_ = fr.is_scheduled(),
-                  .headsign_ = std::string{enter_stop.direction()},
+                  .headsign_ =
+                      [&]() {
+                        if (!enter_stop.direction().empty()) {
+                          return std::string{enter_stop.direction()};
+                        }
+
+                        return to_place(fr[t.stop_range_.to_ - 1], false).name_;
+                      }(),
```
felixguendling, line 351:
```suggestion
                  .headsign_ = enter_stop.direction().empty()
                                 ? to_place(fr[t.stop_range_.to_ - 1], false).name_
                                 : std::string{enter_stop.direction()}
```

No need for a complicated lambda if `?:` does the job. I didn't format it though.

src/journey_to_response.cc
```diff
@@ -341,7 +341,10 @@ api::Itinerary journey_to_response(
                       exit_stop.scheduled_time(n::event_type::kArr),
                   .realTime_ = fr.is_rt(),
                   .scheduled_ = fr.is_scheduled(),
-                  .headsign_ = std::string{enter_stop.direction()},
+                  .headsign_ =
+                      enter_stop.direction().empty()
+                          ? to_place(fr[t.stop_range_.to_ - 1], false).name_
```
felixguendling, line 346: > `auto const exit_stop = fr[t.stop_range_.to_ - 1U];`

This is the exit stop. I guess you want the last stop of the trip, not the exit stop of this journey leg?

The last stop of the transport might not be the same as the last stop of the trip because we concatenate trips with the same `block_id` into one transport.

jbruechert: Oh ok, will need to have a closer look later.

jbruechert: Is there a simple way to reconstruct the original trip after the trips were concatenated?

felixguendling: `frun::for_each_trip` is an example where the transport is split into trip ranges.

So in this case, it would probably make sense to iterate until either the end is reached or the `trip_idx_t` changes (whatever comes first) and output the last seen stop as headsign.

felixguendling: Otherwise, it would probably also be fine to just put the last stop of the transport as that's either the same (e.g. for trips going in circles) or at least its also a stop where you could go with this vehicle.

traines-source: I built a similar thing for ADDED trips recently directly into [frun.cc](https://github.com/motis-project/nigiri/blame/master/src/rt/frun.cc#L174), which does have the advantage that it's also used for stop_times etc. Maybe it makes sense to extend it there. However, I opted for the final stop of the transport (and used `size()` instead of `stop_range` because e.g. in `stop_times.cc` the range is artificially shortened to only the corresponding stop).

felixguendling: Sounds good. Then it's at least consistent with the stop times endpoint.

Issue #861, opened 2025-05-16
Map not usable while transit or itinerary is selected


MichaelKutzner:
## Brief issue description
When a transit or itinerary is selected, the map view cannot be moved and the zoom cannot be changed. Additionally, the browser tab will consume a lot of resources, making the page hardly responsive.

## Detailed issue description
Changing the map view or zoom will trigger map updates repeatedly. This is indicated by many `updateRailviz: effect` messages in the console log. Often this will also result in an infinite loop, causing an error:
```
Uncaught Error: https://svelte.dev/e/effect_update_depth_exceeded
```
After this error happened, the map view and zoom might be changed, but this will no longer update the transits shown.

## Effected browsers
- ✔ Firefox browser
- ✘ Chromium based browsers

## Cause
After some research, I think, that this is caused by MapLibre GL JS's [`flyTo`](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#flyto) method, which will be triggered continuously by an effect:
https://github.com/motis-project/motis/blob/1f3368758390cb2027479b29fcf7984f422a4fd9/ui/src/routes/%2Bpage.svelte#L238-L261

## Discussion
- Is using an `$effect` necessary here?
- If the `$effect` cannot be removed for now, can it be limited? Or are there related issues, that I don't see, yet?

I already tried some changes, to run this code at most once: https://github.com/motis-project/motis/compare/master...MichaelKutzner:motis:flyTo_debug
At least for me, this would solve the performance issue. And I didn't notice any impact on Chromium.

@traines-source: I think, you were the last to make major changes to this function / effect. So you probably know best, if the `$effect` is needed here?

traines-source:
Yes we're definitely using `$effect` too much and I'm using Firefox too seldomly... But the only alternative I see is calling the `flyToSelectedItinerary` function everywhere the itinerary is updated, which are quite a few places... So your solution should be fine, it should really only run if the itinerary changes. Do you create a PR from your branch? I think an alternative also for other places could be [`untrack`](https://svelte.dev/docs/svelte/svelte#untrack).

MichaelKutzner:
Thank you for the explanation. I haven't looked up where this would have to be changed, but it probably would be more work to do.
I will update my code to make it a Pull Request.

Concerning `untrack`, I already attempted a fix using this a while ago ( #822 ). But is seems, that this caused a different issue, I didn't notice ( #846 ).

PR #862, opened 2025-05-16
Ensure itinerary is focused only once

MichaelKutzner:
Implements the fix discussed in #861


felixguendling (approved):

traines-source (approved):
Thanks for noticing and investigating!

PR #884, opened 2025-05-29
Fix stop_times radius parameter filtering

hannsadrian:
I noticed, that setting the radius in the /api/stoptimes endpoint did virtually have no effect. It showed only departures from the specified id if the radius was set to 0 or 1 (which one was a lucky shot) and above that for all surrounding stops belonging together. I've looked at the filtering logic and noticed that after the geo filtering, add(y) was called. However, the add function looks for similar stops or childs and adds these as well.

In my specific use case it is very relevant to be able to reliably view departures in a small radius. This pr enables this by adding only locations that are actually in the query radius. I think this is the intended way for the parameter to work.


felixguendling:
What about API stability? I would prefer to introduce an `exact` parameter that is optional and false by default so the API stays backward compatible.

openapi.yaml
```diff
@@ -1265,6 +1265,17 @@ paths:
...
+        - name: exactRadius
+          in: query
+          schema:
+            type: boolean
+          required: false
+          description: |
+            Optional. Default is `false`.
+            
+            If set to `true`, only stations that are phyiscally in the radius are considered.
+            If set to `false`, additionally to the stations in the radius, equivalences and children are considered.
```
```suggestion
            If set to `false`, additionally to the stations in the radius, equivalences with the same name and children are considered.
```

src/endpoints/stop_times.cc
```diff
@@ -299,7 +299,13 @@ api::stoptimes_response stop_times::operator()(
...
-                         [&](n::location_idx_t const y) { add(y); });
+                         [&](n::location_idx_t const y) {
+                           if (query.exactRadius_ == true) {
+                             locations.emplace_back(y);
+                           } else {
+                             add(y);
+                           }
+                         });
   } else {
     add(x);
```
felixguendling: the same should apply to `!query.radius_` (only return the exact events from the selected location)

src/endpoints/stop_times.cc
```diff
@@ -299,9 +299,19 @@ api::stoptimes_response stop_times::operator()(
   if (query.radius_) {
     loc_rtree_.in_radius(tt_.locations_.coordinates_[x],
                          static_cast<double>(*query.radius_),
-                         [&](n::location_idx_t const y) { add(y); });
+                         [&](n::location_idx_t const y) {
+                           if (query.exactRadius_ == true) {
```
felixguendling: - never use `== true` - that's not needed
- instead of doing the check 2x it can be moved to `add` (`if exact locations.emplace_back(l); return;`)

hannsadrian: thank you, that's totally right. I fixed it 😇

PR #889, opened 2025-06-03
add place description (as in motis-project/nigiri#204)

hannsadrian:
this pr shows the stop description implemented in https://github.com/motis-project/nigiri/pull/204


openapi.yaml
```diff
@@ -1882,6 +1882,9 @@ components:
             The current track/platform information, updated with real-time updates if available. 
             Can be missing if neither real-time updates nor the schedule timetable contains track information.
           type: string
+        description:
+          description: description of the location that provides useful, quality information
```
```suggestion
          description: description of the location that provides more detailed information
```

hannsadrian: the original description was taken from the gtfs spec 😅

felixguendling: True. But I would not want to be blamed if the information is neither useful, nor quality (as it's often the case with real-world data).

PR #890, opened 2025-06-03
ODM mixer requirements

mority:
- CSV format to specify I/O of ODM `mixer` in unit tests
- Printing capabilities to output I/O of ODM `mixer` in CSV format while `meta_router` responds to queries


src/odm/mixer/journeys.cc
```diff
@@ -0,0 +1,197 @@
...
+  T n;
+  auto [ptr, ec] = std::from_chars(first, last, n);
+  if (ec == std::errc::invalid_argument ||
+      ec == std::errc::result_out_of_range) {
+    return std::nullopt;
+  }
+  return n;
+}
+
+std::optional<nigiri::unixtime_t> read_hours_minutes(std::string_view t) {
```
traines-source: Is this still used somewhere or has it been replaced by `nigiri::parse_time`? Didn't find the usage.

traines-source:
Looks good, wdyt, should we add more fixtures (e.g. from https://github.com/motis-project/motis/pull/811) before or after reworking the "generosity"?

src/odm/mixer/journeys.cc
mority: Right, that's not used anymore. Initially, I used time of days for departures/arrivals but decided to use datetimes in the end.

traines-source (approved):

mority:
I think, it would be great to add more real-world cases. But maybe only ones where it is clear that we won't be revoking them in the near future.

traines-source:
Well I guess what we'll revoke is up for debate :) The last three examples over in #811 (Bad Muskau-Cottbus, Uhsmannsdorf, Bad Mukskau-Pechern, basically always the early ones) are the ones that were actively requested by customers or actually used (cf. journeys completed so far)

mority:
#1051

PR #891, opened 2025-06-04
Add visualization for isochrones

MichaelKutzner:
Add a first implementation to visualize isochrones.

image

Most settings will be shared with the "Connections" tab, which should make maintenance easy.
I also added support to share the query URL. But it currently doesn't get updated, when color or opacity are updated. I need to check the reason for that later.
A different, more irritating issue, occurs, when the map is zoomed out very far. This will cause the reachable stations to be displayed at the wrong location. I need to investigate, as the used projection is probably too simple.
image

About the implementation:
Isochrones at each reachable station are estimated by the time remaining and used street mode. At least for now. This might become more accurate, if supported by the server.
For rendering an additional layer is used, that uses a canvas element. This is done, to avoid long computations, that would otherwise have been caused by creating the union of all polygons. Other approaches, which might have reduced the required work, didn't work well for me.

Possible improvements:
- Render simplified shapes first, to faster show some results
- Remove shapes, that are completely contained in an other shape.
- Perform expensive calculations asynchronous
- Compute union asynchronous and update the layer, once the computation is completed

These will not be part of this Pull Request


ui/src/lib/i18n/de.ts
```diff
@@ -83,6 +85,8 @@ const translations: Translations = {
 	REGIONAL_RAIL: 'Regionalzug',
 	OTHER: 'Andere',
 	routingSegments: {
+		maxTransfers: 'Max. Transporte',
```
traines-source:
```suggestion
		maxTransfers: 'Max. Umstiege',
```

ui/src/lib/i18n/fr.ts
```diff
@@ -82,6 +84,8 @@ const translations: Translations = {
 	REGIONAL_RAIL: 'Train régional',
 	OTHER: 'Autres',
 	routingSegments: {
+		maxTransfers: 'Transport max.',
```
traines-source:
```suggestion
		maxTransfers: 'Max. de correspondances',
```

MichaelKutzner:
Thank you for the feedback.

> I wasn't able to run this locally (`Named export 'Map' not found. The requested module 'maplibre-gl' is a CommonJS module, which may not support all module.exports as named exports.`), but might be something wrong with my setup. Will need to look more closely into everything.

I need to check, where this might come from. Maybe I broke something, without realizing.

> Have you checked that maplibregl (being a canvas-based lib) doesn't support the stuff that you need to do (also the turf stuff) in a fast way? That would also resolve the projection issues I guess.

That's a good point. I will check this again. Turf.js works fine on small data sets, but having over 50.000 circles causes a lot of work. So this seemed like a reasonable tradeoff.
About the projection issue, I'm currently thinking about implementing this in multiple stages. A fast approximation first, to immediately render some results, even if they might be incorrect. And to replace these later with the more accurate results, once they have been computed.

> And do we really need to allow the user to select the opacity and in particular the color (esp. because if we go the full way similar to direkt.bahn.guru/chronotrains.com, we could at some point actually use colors for different levels of isochrones/per hour of travel time)?

I decided to add an option to select color and opacity for now, as some colors and combinations might be less optimal, if most transports of an area use a similar color. Or for visually impaired people. But I'm fine, if we use a fixed value. And maybe allow changing these via query parameters, if there is demand.
I already thought about different colors for different isochrones as well. I think, this would at least require a gradient, instead of a simple color picker. Or one color per isochrone, which might lead to a confusing UI.

> Also, similar to your findings, I had tested that even on the Transitous dataset, one can increase the maxTravelTime to something like 24h with the queries still only taking a few seconds, just that then the responses are maybe 100 MB in size for all train connections – which for backend usage isn't that much of a problem though. So I intend to propose increasing the limits over at Transitous. Maybe one should have a separate, lower limit for the UI then, because 100 MB of JSON and rendering the corresponding shapes will probably kill the browser.

I haven't tested queries that long yet. But I agree, that a 100 MB response will most likely cause some problems for the browser. Also the available options in the UI shouldn't be that high, however, I wouldn't deny long duration queries, if the server supports them. For example, if someone is interested in all reachable stations without any changeover. I don't think, that this would cause a problem.

ui/src/lib/i18n/de.ts
MichaelKutzner: That's a good point. I tried to reflect the internal naming convention, but that's something the user shouldn't care or know about. So selecting the number of changeovers (`0…maxTransfers`) instead might be a better fit.
I will change it.

MichaelKutzner: The UI now uses changeovers instead of transfers. I hope, this makes using more intuitive.

ui/src/lib/i18n/fr.ts
MichaelKutzner: Same as for `de.ts`

ui/src/routes/+page.svelte
```diff
@@ -351,10 +462,11 @@
 					? 'hide'
 					: ''}
 			>
-				<Tabs.Root value="connections" class="max-w-full w-[520px] overflow-y-auto">
-					<Tabs.List class="grid grid-cols-2">
+				<Tabs.Root bind:value={activeTab} class="max-w-full w-[520px] overflow-y-auto">
+					<Tabs.List class="grid grid-cols-3">
 						<Tabs.Trigger value="connections">{t.connections}</Tabs.Trigger>
 						<Tabs.Trigger value="departures">{t.departures}</Tabs.Trigger>
+						<Tabs.Trigger value="isochrones">Isochrones</Tabs.Trigger>
```
traines-source: i18n

MichaelKutzner:
> > Named export 'Map' not found
> 
> Your fix helped indeed, but I had to switch to a `type` import for `CanvasSource` as well. Not sure why this isn't a problem for you and the CI.

Thank you for the feedback. I'm also not sure why in behaves different, but I will update this as well.

> > I decided to add an option to select color and opacity for now
> 
> I do agree that it's difficult to find an easily distinguishable color (esp. for the bright mode map, but I guess having a greyscale bright map variant is overkill). At the same time I think too many options can be overwhelming. In any case, the color picker is really laggy for my, maybe only update the color when closing it?
> 

I haven't checked, how the browser handles the color picker. Maybe there is an option to detect, whether the dialog is open. In that case, it might be possible to limit the rendering.
About the colors, we recently had a brief discussion, whether vehicles should be visible, if isochrones are enabled, as this takes resources as well. If we render just the isochrones, the color can be picked easily.

> In general it is freezing on me quite often (of course more so with increasing travel time and transit modes), but admittedly I'm on a slow computer. Maybe we would at least need a spinner like for the normal queries, maybe one can put some stuff off the render thread? And maybe one could have a look how direkt.bahn.guru/[chronotrains.com](http://chronotrains.com/) have been doing it, because the latter is working well for me even for European queries (but maybe less stations still since it's only trains – and presumably they are doing much more preprocessing). Then again I'm wondering whether it's worth the effort to try and copy everything they already have...
> 

I agree, that the current solution can be quite laggy, especially when rendering many datapoints. I already took some time to get familiar with worker threads, to offload some computations and the rendering. But it's still work in progress.
I'm not happy with the current state yet, as the first actions can still cause noticeable delays. However, it's now possible to get the isochrones as a GeoJSON object, which fixes the projection error.
I will have another look at these as well.

> Nice idea with the size of the circles representing the postTransitTime!

MichaelKutzner:
I think, the improvements made for rendering are now ready for review.

The most noticeable things that have changed:
* Render boxes to show results faster and on less powerful hardware
* For best results, which also aren't effected by the projection issue, a proper union can be computed
* The rendering and maximum computation level can now be selected
* Computation and rendering will be performed by workers
* Vehicle positions will no longer be visible on the isochrones view
* The URL is now updated properly, if opacity or color are changed
    * As vehicle positions are no longer shown, we can also fix them now, if you prefer

To indicate to the user, if the selected and shown levels match, non matching levels will render the isochrones in a different color. It's currently hardcoded to be magenta.

Please don't forget to check the performance. While it works reasonable fine for me, your experience might be different. For example, maybe I'm using the workers in a way, that could easily be improved, as I think, I'm still not familiar enough with them.

traines-source:
This is feeling much snappier, so much so that it's actually fun to use :+1: Works nicely even on my slow computer, even for the geometry circles (although they are taking quite a bit longer, as expected).
The magenta "loading indicator" is difficult to understand I think. People (at least I before reading) are clicking around wondering why it's not the color they selected (if we keep it selectable – I'm always a fan of less cluttered UIs but if there are good reasons to be able to choose the color, then why not). Any reason to not just display a spinning circle (while also previewing the approximate isochrones in the selected color?)
Also I think making both the rendering and computation level selectable is overwhelming. At least only show the computation level (rendering level could still be selectable via the query params), maybe even none of the two since you have built the nice "progressive enhancement" behavior?

When selecting a maxTransitTime greater than the server side maximum (where the default is still 90 min), it fails silently. Maybe at least display an error (ofc ideally only possible values would be selectable but I guess this is unnecessarily complicated) – otherwise not distinguishable from waiting for the first render or a place where no isochrones were found (e.g. somewhere on a mountain).

MichaelKutzner:
Thank you for the feedback

> This is feeling much snappier, so much so that it's actually fun to use 👍 Works nicely even on my slow computer, even for the geometry circles (although they are taking quite a bit longer, as expected). The magenta "loading indicator" is difficult to understand I think. People (at least I before reading) are clicking around wondering why it's not the color they selected (if we keep it selectable – I'm always a fan of less cluttered UIs but if there are good reasons to be able to choose the color, then why not). Any reason to not just display a spinning circle (while also previewing the approximate isochrones in the selected color?) Also I think making both the rendering and computation level selectable is overwhelming. At least only show the computation level (rendering level could still be selectable via the query params), maybe even none of the two since you have built the nice "progressive enhancement" behavior?

That's a good suggestion, I didn't think about. Showing a loading symbol (spinning circle), while still rendering in the selected color should work fine. I just wanted something to indicate, that it's still not finished.
Removing one option should probably work. Making the remaining an URL only parameter might be fine as well. I need to implement and test this.

> When selecting a maxTransitTime greater than the server side maximum (where the default is still 90 min), it fails silently. Maybe at least display an error (ofc ideally only possible values would be selectable but I guess this is unnecessarily complicated) – otherwise not distinguishable from waiting for the first render or a place where no isochrones were found (e.g. somewhere on a mountain).

Unless there is an endpoint, that can report the server limits, it's probably very difficult to only show possible values. Having at least an error message is therefore helpful indeed.

I will try to add these changes.

MichaelKutzner:
I have updated the code, according to your suggestions.
There will now be an error message or warning, if the server request failed or there are no isochrones to render. If computation is still in progress, the loading symbol will shown instead.
As I managed to combine both options for selecting the isochrones level, the layout is a bit simpler now.

_Edit_: Add an example image of the UI changes
image
Do you think, this is too simple? I didn't want to fill the space with more text. But I also haven't deleted the now unused translations.

MichaelKutzner:
> * Is it intentional that before doing anything, it's already showing the error "No data"?

That's something, I previously had no idea how to fix. But I don't think, it would have been wrong, just unintuitive.
It should be handled now.

> * Instead of "Anfrage fehlgeschlagen", couldn't we display the actual server error message as we do in the other places? Because then at least there is a hint what the problem is ("maxTravelTime too large: 150 > 90") even if it is always in English.

The error message should now be visible.
Took me a while to get it working and it's probably not solved best.

> * If one has executed a normal plan query before switching to the Isochrones tab, the results card of that persists and the errors are shown below that results card. There is currently no way of closing that results card. In the "Abfahrten" tab, it at least disappears when doing a stoptimes query there.

I extended the condition to hide these, if the `isochrones` tab is selected.

> * The loading symbol starts showing only after the request has returned from the server I think. But usually that part is fast so it's not that noticeable :)

I updated the behavior. There is a small gap, where the icon will vanish, after the response was received and before the more expensive calculations start. But this seems fine to me.

> But I guess all of these are details, we can also fix these in a later PR. Apart from that, I like how it is now 👍

traines-source:
Nice, thank you! :raised_hands: Curious to see it on https://api.transitous.org !

PR #910, opened 2025-06-29
Allow to set the maximum number of transfers in the UI

schaerfo:
This is how it looks:

grafik

Feel free to make changes, I am not very familiar with CSS/Tailwind

### DIFF
--- a/ui/src/lib/AdvancedOptions.svelte
+++ b/ui/src/lib/AdvancedOptions.svelte
@@ -26,7 +26,7 @@
 		requireBikeTransport = $bindable(),
 		requireCarTransport = $bindable(),
 		transitModes = $bindable(),
-		maxTransfers = $bindable(undefined),
+		maxTransfers = $bindable(),
 		maxTravelTime = $bindable(undefined),
 		possibleMaxTravelTimes = [],
 		preTransitModes = $bindable(),
@@ -46,7 +46,7 @@
 		requireBikeTransport: boolean;
 		requireCarTransport: boolean;
 		transitModes: TransitMode[];
-		maxTransfers: number | undefined;
+		maxTransfers: number;
 		maxTravelTime: number | undefined;
 		possibleMaxTravelTimes?: NumberSelectOption[];
 		preTransitModes: PrePostDirectMode[];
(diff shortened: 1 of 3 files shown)


traines-source:
A maxTransfers UI element is added for the isochrones over in #891 as well. While I don't recall rn how the unlimited transfers case was handled there (although actually it's not unlimited, but 14 :), I think we should unify this between the two – especially since the isochrones tab is already reusing the AdancedOptions component. And I personally think that a dropdown like over there or maybe a `input type="number"` should be sufficient.

MichaelKutzner:
Thank you for bringing this to my attention.
For isochrones, the upper limit is set to match the server limit. As the number isn't that high, I found a drop-down menu still easy to use.
Right now, this option is not enabled on the “Connection” tab. But this can be changes without much afford.
The current design looks like that:
image

MichaelKutzner:
With #891 merged, I made the adjustments to show the dropdown to select `maxTransfers` for connection searches as well: https://github.com/MichaelKutzner/motis/commit/9af28848c24930433c7d82534ddb1817841a3dc3
As I don't know, what options are reasonable for `maxTravelTime` here, the corresponding dropdown is disabled for now.
Do you plan to update your Pull Request? Or should I create a new one for this?

schaerfo:
Great, I'll take a look at the `NumberSelect` you merged and update this PR

schaerfo:
Ok, I rebased the patch and it got a quite a bit smaller :) It now looks like this:
grafik
* The connections and isochrones tabs share state, so setting max transfers in one tab will also change it in the other
* I removed the switch to enable limiting the number of transfers

traines-source:
Nice, what do you think, what is less confusing? Always showing the actual technical upper limit of 14 transfers or showing some special label like "Unlimited" for that in the dropdown which is technically wrong?
Also, the Linter is unhappy, you can probably just run `(p)npm run format`

felixguendling:
> some special label like "Unlimited"

I think the concrete number is better. Maybe even configurable in the server because for small datasets a smaller number might improve performance without any disadvantage for the user.

traines-source:
> configurable in the server

Ok but I think this is out of scope for this PR, since we have no straightforward way yet to pass this to the UI.
I.e. we can leave everything as it is apart from the formatting/linting.

schaerfo:
Alright, I ran the formatter. Also, I didn't check if the server-side limit for max transfers is the same for `plan` and `one-to-all`

traines-source:
It is, thank you!

PR #913, opened 2025-06-30
VDV AUS: auser

mority:


mority:
During my latest experiments, `auser` is running stable. It handles the VDV AUS subscription stuff and buffers the VDV AUS updates from upstream.

I incorporated the hints from the conversation at #840
`auser` would answer `DatenBereit` and `ClientStatus` requests from upstream. 'Would' because upstream does not send any. Moreover, it would immediately fetch again from upstream if the response contained a `WeitereDaten` set to `true`. But again, this part of the VDV AUS protocol is not used by upstream.

Therefore, right now `auser` simply requests upstream updates every 15 seconds. It buffers the updates from the current and the previous subscription. With a subscription interval of 1 hour, a total of about 300 - 700 MB of updates have to be stored this way.

We are getting a lot of updates that contain incomplete runs, for which we have not been shown a full run in a previous update (`incomplete not seen before`). For many, we are shown a full run in a later update.

The matching ratio is ~93% (`match attempts` / `matched runs`). We only do a `match attempt` for full runs. I tried matching on incomplete runs as well, but that lead to some false positives and nonsensical output in the UI.

There is lots of redundance in the DELFI GTFS. About 37% of runs matched have multiple matches in the static timetable. I have not done a detailed analysis on this but on first glance many are perfect duplicates, i.e., X out of X stops match perfectly for more than one run of the static timetable.

```
unsupported additional runs: 2470
unsupported additional stops: 1463
current matches total: 8682
current matches non-empty: 7628
total runs: 171232
complete runs: 77802
unique runs: 8682
match attempts: 8202
matched runs: 7628
multiple matches: 2868
incomplete not seen before: 2485
complete after incomplete: 2005
no transport found at stop: 1558
total stops: 2960490
resolved stops: 2828649
runs without stops: 3798
cancelled runs: 265
skipped vdv stops: 121229
excess vdv stops: 37244
updated events: 3494321
propagated delays: 523172
```

mority:
requires motis-project/nigiri#217

PR #921, opened 2025-07-07
add running scheduled with realtime metrics

traines-source:


src/endpoints/metrics.cc
```diff
@@ -1,19 +1,137 @@
...
+    auto& real =
+        metrics.current_trips_running_scheduled_with_realtime_count_.Add(
+            labels);
+    sched.Set(0);
+    real.Set(0);
+    metric_by_agency.emplace_back(
+        std::pair{std::reference_wrapper{sched}, std::reference_wrapper{real}});
+  }
+
+  auto const get_provider_idx = [&](n::rt::frun const& fr) {
```
felixguendling: could be better added to frun directly?

src/endpoints/metrics.cc
```diff
@@ -1,19 +1,137 @@
...
+    auto const from = n::stop_idx_t{0U};
+    auto const to = static_cast<n::stop_idx_t>(seq.size() - 1);
+    auto const [start_day, _] = tt.day_idx_mam(time_interval.from_);
+    auto const [end_day, _1] = tt.day_idx_mam(time_interval.to_);
+
+    auto const dep_times = tt.event_times_at_stop(r, from, n::event_type::kDep);
+    for (auto const [i, t_idx] :
+         utl::enumerate(tt.route_transport_ranges_[r])) {
+      auto const day_offset =
+          static_cast<n::day_idx_t::value_t>(dep_times[i].days());
```
felixguendling: should always be zero?

src/endpoints/metrics.cc
```diff
@@ -1,19 +1,137 @@
...
+                           {"agency_name", std::string{agency_name}},
+                           {"agency_id", std::string{agency_id}}};
+    auto& sched = metrics.current_trips_running_scheduled_count_.Add(labels);
+    auto& real =
+        metrics.current_trips_running_scheduled_with_realtime_count_.Add(
+            labels);
+    sched.Set(0);
+    real.Set(0);
+    metric_by_agency.emplace_back(
+        std::pair{std::reference_wrapper{sched}, std::reference_wrapper{real}});
```
felixguendling: emplace_back means you provide ctor `args...`
-> either do this (which means no `std::pair{...}`)
-> or use `push_back` (which has the same effect as what you do now, just without so many template instanciations)

felixguendling: also there's `std::ref` as shorthand

src/endpoints/metrics.cc
```diff
@@ -1,19 +1,137 @@
...
 
+#include "utl/enumerate.h"
+
+#include "nigiri/rt/frun.h"
+#include "nigiri/rt/rt_timetable.h"
+#include "nigiri/timetable.h"
+#include "nigiri/types.h"
+
+#include "motis/data.h"
+
```
felixguendling: group together

felixguendling (approved):

felixguendling:
what's the performance of checking the whole timetable for active transports in the Transitous case?

PR #926, opened 2025-07-11
separate route / trip in API response

jbruechert:
Depends on https://github.com/motis-project/nigiri/pull/208


felixguendling:
superseded by #977

PR #933, opened 2025-07-12
Add translated quickstart documentation

MichaelKutzner:
Add an explicit quickstart guide, that can be used by non developers.
As the documentation is to be read by operators, the file has been translated as well.

### DIFF
--- a/docs/quickstart.de.md
+++ b/docs/quickstart.de.md
@@ -1,0 +1,30 @@
+# Schnellstart
+
+1. Den [aktuellen Release](https://github.com/motis-project/motis/releases) herunterladen
+
+1. OpenStreetMap-Daten im `.osm.pbf`-Format herunterladen
+
+1. Eine oder mehrere _timetable_s herunterladen
+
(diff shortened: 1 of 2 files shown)


docs/quickstart.md
```diff
@@ -0,0 +1,29 @@
+# Quickstart
```
felixguendling: English version is already available in README.md. What's the difference here?

docs/quickstart.de.md
```diff
@@ -0,0 +1,30 @@
+# Schnellstart
```
felixguendling: should this be linked in the README.md and be a direct translation including the copy & paste bash/PowerShell script?

docs/quickstart.md
MichaelKutzner: True. I misread the request. I will update to explain more complex points, that are still missing.

docs/quickstart.de.md
MichaelKutzner: We can also not add it, to make maintenance easier.

PR #934, opened 2025-07-12
Add a tool to inspect the binary data

MichaelKutzner:
This adds a simple tool, that can be used to inspect the binary data.

Currently implemented features:
* ~~Show basic statistics~~ _Edit: Will be superseded by #935_
* Show the shape offsets for a given trid-id

~~As the tool is intended for debugging purposes, the features are not part of the main binary. But I wouldn't mind, if this changes, once the features get extended and are actively used.~~ _Edit: The `analyze-shapes` sub command has been merged to the `motis` binary, to be easier usable._

### DIFF
--- a/src/analyze_shapes.cc
+++ b/src/analyze_shapes.cc
@@ -3,6 +3,7 @@
 #include "utl/verify.h"
 
 #include "fmt/base.h"
+#include "fmt/ranges.h"
 
 #include "nigiri/shapes_storage.h"
 #include "nigiri/types.h"
@@ -32,9 +33,9 @@ bool analyze_shape(nigiri::shapes_storage const& shapes,
 }
 
 bool analyze_shapes(data const& d, std::vector<std::string> const& trip_ids) {
-  utl::verify(d.tt_, "Missing timetable");
-  utl::verify(d.tags_, "Missing tags");
-  utl::verify(d.shapes_.get() != nullptr, "Missing shapes");
+  utl::verify(d.tt_ != nullptr, "Missing timetable");
+  utl::verify(d.tags_ != nullptr, "Missing tags");
+  utl::verify(d.shapes_ != nullptr, "Missing shapes");
 
   auto success = true;
   for (auto const& trip_id : trip_ids) {


CMakeLists.txt
```diff
@@ -149,6 +149,18 @@ target_link_libraries(motis
     pbf_sdf_fonts_res-res
     tiles_server_res-res
 )
+# --- INSPECT ---
```
felixguendling: Why a separate binary that's not part of the distribution? Or: why not a motis binary sub command?

MichaelKutzner:
I think, the code is now cleaned up.
As loading the data might take a moment, I decided to support passing multiple trip ids.
The sub command should work with all `tripId`s copied from the UI, that will be decoded automatically. Therefore it's not required to replace `%3A` by `:` manually, for example. However, if a trip id contains `%` and is not encoded, this might not find the correct `trip_idx`. But as these must be built manually anyways, I don't think, this will be a problem.

As a side note, I realized, that the `server` and `import` sub commands don't add the `help` options. Is this intended?

src/analyze_shapes.cc
```diff
@@ -0,0 +1,59 @@
...
+  for (auto const& o : offsets) {
+    fmt::print("{}, ", o);
+  }
+  fmt::println("");
+  return true;
+}
+
+bool analyze_shapes(data const& d, std::vector<std::string> const& trip_ids) {
+  utl::verify(d.tt_, "Missing timetable");
+  utl::verify(d.rt_ && d.rt_->rtt_, "Missing realtime data");
```
felixguendling: why?

src/analyze_shapes.cc
```diff
@@ -0,0 +1,59 @@
...
+  utl::verify(d.tags_, "Missing tags");
+  utl::verify(!!d.shapes_, "Missing shapes");
+
+  auto success = true;
+  for (auto const& trip_id : trip_ids) {
+    fmt::println("Searching trip-id '{}' ...", trip_id);
+    auto const [run, trip_idx] =
+        d.tags_->get_trip(*d.tt_, &*d.rt_->rtt_, trip_id);
```
felixguendling:
```suggestion
        d.tags_->get_trip(*d.tt_, nullptr, trip_id);
```

Will always be nullptr?

src/analyze_shapes.cc
```diff
@@ -0,0 +1,59 @@
...
+    return false;
+  }
+
+  auto const offsets = shapes.offsets_[offset_idx];
+  if (offsets.empty()) {
+    fmt::println("Empty shape for '{}'", trip_idx);
+    return false;
+  }
+
+  fmt::println("Offsets for '{}':", trip_idx);
```
felixguendling: can't we use `fmt/ranges.h`?

src/analyze_shapes.cc
```diff
@@ -0,0 +1,59 @@
...
+  for (auto const& o : offsets) {
+    fmt::print("{}, ", o);
+  }
+  fmt::println("");
```
felixguendling:
```suggestion
 fmt::println("{}", offsets);
```

MichaelKutzner: This totally works and is more readable. I just wasn't aware of that function.
Has been updated.

src/analyze_shapes.cc
MichaelKutzner: If this refers to the comment below, it's a good suggestion. The header, however, doesn't seem required for formatting. So I'm probably missing something.

src/analyze_shapes.cc
MichaelKutzner: I assumed this is always required, as it's used here: https://github.com/motis-project/motis/blob/79516e8956be22fb1598b53be4c23883092a326b/src/endpoints/trip.cc#L28
But changing it seems to work as well.

src/analyze_shapes.cc
MichaelKutzner: See comment below.

src/analyze_shapes.cc
felixguendling: in terms of IWYU = include what you use
it doesn't matter if the header is already there from another include
as long the "include fmt/ranges.h" is not part of the external interface for the other header (where ever it comes from), we can't rely on it being there
changing other headers should NOT break unrelated code!!

src/analyze_shapes.cc
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  auto const offsets = shapes.offsets_[offset_idx];
+  if (offsets.empty()) {
+    fmt::println("Empty shape for '{}'", trip_idx);
+    return false;
+  }
+
+  fmt::println("Offsets for '{}':", trip_idx);
+  fmt::println("{}", offsets);
```
felixguendling: print statements can be joined together

CMakeLists.txt
MichaelKutzner: This seems outdated. All changes to this file have been reverted before.

src/analyze_shapes.cc
```diff
@@ -0,0 +1,55 @@
...
+
+  return true;
+}
+
+bool analyze_shapes(data const& d, std::vector<std::string> const& trip_ids) {
+  utl::verify(d.tt_, "Missing timetable");
+  utl::verify(d.tags_, "Missing tags");
+  utl::verify(!!d.shapes_, "Missing shapes");
```
felixguendling: no `!!` - it's not JavaScript here...

src/analyze_shapes.cc
```diff
@@ -0,0 +1,55 @@
...
+    fmt::println("Searching trip-id '{}' ...", trip_id);
+    auto const [run, trip_idx] = d.tags_->get_trip(*d.tt_, nullptr, trip_id);
+    if (!run.valid()) {
+      success = false;
+      fmt::println("Did not find trip idx for trip-id '{}'", trip_id);
+      continue;
+    }
+    fmt::println("Found trip idx for trip-id '{}': {}", trip_id, trip_idx);
+    success &= analyze_shape(*d.shapes_, trip_idx);
+    fmt::println("");
```
felixguendling: not needed, if you really want another \n, add it to the previous println

src/analyze_shapes.cc
```diff
@@ -0,0 +1,55 @@
...
+}
+
+bool analyze_shapes(data const& d, std::vector<std::string> const& trip_ids) {
+  utl::verify(d.tt_, "Missing timetable");
+  utl::verify(d.tags_, "Missing tags");
+  utl::verify(!!d.shapes_, "Missing shapes");
+
+  auto success = true;
+  for (auto const& trip_id : trip_ids) {
+    fmt::println("Searching trip-id '{}' ...", trip_id);
```
felixguendling: not really needed as this should be fast?

src/analyze_shapes.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include "motis/tag_lookup.h"
+
+namespace motis {
+
+bool analyze_shape(nigiri::shapes_storage const& shapes,
+                   nigiri::trip_idx_t const& trip_idx) {
+  auto const offset_idx = shapes.trip_offset_indices_[trip_idx].second;
+  if (offset_idx == nigiri::shape_offset_idx_t::invalid()) {
+    fmt::println("No shape offsets for '{}'", trip_idx);
```
felixguendling: is the trip_idx really helping the user? user works with trip ids from the URL, right?

src/analyze_shapes.cc
MichaelKutzner: I have updated the output messages.
I found the previous version slightly more readable and didn't think these optimizations were necessary here.

src/analyze_shapes.cc
MichaelKutzner: I agree, I didn't like this as well, but missed a better solution. Should be fixed now.

src/analyze_shapes.cc
MichaelKutzner: Updated, see comment above.

src/analyze_shapes.cc
MichaelKutzner: It's fast, but as it's intended for analyzing and debugging, having some more messages didn't seem an issue to me.
The output is now removed and the other messages have been updated.

src/analyze_shapes.cc
MichaelKutzner: That's right.
I have updated the messages to use the (decoded) tridId instead.

src/analyze_shapes.cc
```diff
@@ -0,0 +1,52 @@
...
+
+  fmt::println("Offsets for trip-id '{}':\n{}\n", trip_id, offsets);
+
+  return true;
+}
+
+bool analyze_shapes(data const& d, std::vector<std::string> const& trip_ids) {
+  utl::verify(d.tt_, "Missing timetable");
+  utl::verify(d.tags_, "Missing tags");
+  utl::verify(d.shapes_.get() != nullptr, "Missing shapes");
```
felixguendling: inconsistent with the above - either always use != nullptr or never
`.get()` is not necessary as there are overloads for `std::unique_ptr`

src/analyze_shapes.cc
```diff
@@ -0,0 +1,52 @@
...
+    return false;
+  }
+
+  auto const offsets = shapes.offsets_[offset_idx];
+  if (offsets.empty()) {
+    fmt::println("Empty shape for trip-id '{}'\n", trip_id);
+    return false;
+  }
+
+  fmt::println("Offsets for trip-id '{}':\n{}\n", trip_id, offsets);
```
felixguendling, line 30: still I think we should include `fmt/ranges.h` to make sure this code will compile in the future when whatever other place includes `fmt/ranges.h` might not have it anymore (IWYU)

src/analyze_shapes.cc
MichaelKutzner: I know. My issue here is, that the LSP claims, that the header is not used directly. So it wants to delete the `#include` anyways. That's what I meant, that I'm probably missing something. Maybe a bug with the LSP, so I will add it.

I also missed this response before.

src/analyze_shapes.cc
MichaelKutzner: Would this be better?
```c++
utl::verify(bool{d.shapes_}, "Missing shapes");
```
Using it without `bool` doesn't seem possible, as the function is marked `explicit`: https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool.html
Should I add this to all checks then, to keep it consistent?

felixguendling: `utl::verify(d.shapes_ != nullptr, "Missing shapes");` seems most idiomatic to me then

MichaelKutzner: It's updated

src/analyze_shapes.cc
MichaelKutzner: You are right. I still don't understand why the LSP has a different opinion on that, but as it's used anyways, it's better to include than to miss it in the future.

PR #941, opened 2025-07-18
extend calendar

### LINKED ISSUE #936: Loop expired service days (mark in output)
- Continue `calendar.txt` into present if feed is expired
- Mark information as expired and projected into present/future in API

felixguendling:
closes #936
unblocks https://github.com/public-transport/transitous/pull/1331

new timetable.dataset.extend_calendar=true/false flag (default = false)


jbruechert:
Thank you for working on this :)

traines-source:
I think we should add a notice in the API/UI though when a feed is outdated

PR #946, opened 2025-07-20
Fix OpenAPI description of transit mode aliases

schaerfo:
For compatibility, the behaviour was left as-is. Instead, the documentation was adapted to match the behaviour.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1880,14 +1880,14 @@ components:
 
         # Transit modes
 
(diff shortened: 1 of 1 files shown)


felixguendling:
`TRANSIT` should include `SUBWAY` as it's just activating all bits in the bitmask:
https://github.com/motis-project/motis/blob/master/src/timetable/modes_to_clasz_mask.cc#L15

schaerfo:
That would be redundant, as `SUBWAY` is already included in `RAIL`

felixguendling:
You're right, thank you!

PR #950, opened 2025-07-22
fix direct arrival times and intervals for arriveBy=true

### LINKED ISSUE #918: Direct non-PT connections are wrongly timed
I've searched for a connection with arrival at a certain hour, public transport worked well with that, however a direct pedestrian route had a departure at that time, contrary to the query

For the moment I think it works anywhere, if you can't reproduce it, I'll give the example

Edit: looks like all non-PT means have that

traines-source:
fixes #918


src/endpoints/routing.cc
```diff
@@ -329,7 +329,7 @@ std::pair<n::routing::query, std::optional<n::unixtime_t>> get_start_time(
     auto const t = std::chrono::time_point_cast<n::i32_minutes>(
         *query.time_.value_or(openapi::now()));
     auto const window = std::chrono::duration_cast<n::duration_t>(
-        std::chrono::seconds{query.searchWindow_ * (query.arriveBy_ ? -1 : 1)});
+        std::chrono::seconds{query.searchWindow_});
     return {{.start_time_ = query.timetableView_
                                 ? n::routing::start_time_t{n::interval{
                                       query.arriveBy_ ? t - window : t,
```
traines-source: For a arriveBy=true query at 9:00, this would lead to a from=11:00 to=9:00 interval. Or is this intentional because then by default (with `numItineraries=5`) it only got extended to a 4 hour interval and not to 8 hours as now? This exacerbates the issue described here https://github.com/motis-project/prima/issues/199 also for "normal" MOTIS deployments.

traines-source:
reverted the arriveBy start interval until we have found a good way to fix this in a non-breaking breaking way, so this PR can be merged independently

Issue #935, opened 2025-07-13
MOTIS Import: generate statistics about loaded feeds


felixguendling:
- earliest transport
- last transport
- #locations
- #trips
- #transports x active traffic days

MichaelKutzner:
I started working on this

traines-source:
IMO would be cool if this were also exposed as prometheus metrics similar to the scheduled coverage metrics (`metrics.cc` , with the difference that this only needs to be calculated once?). But we can also add this later.

MichaelKutzner:
Some questions so far:

* Should earliest and latest transport extend up to the maximal interval defined in the feed? Or restricted to the configured range only? Or maybe both?
    **Example**: Assume a feed covering `(2025-01-01, 2025-12-31)` and the server `(2025-07-01, 2026-06-30)`. Should we then  only show `(2025-07-01, 2025-12-31)` or `(2025-01-01, 2025-12-31)` as well? Or something else?
* Should on demand trips count towards the `#transports × traffic days` metric? Or should  it be stored and reported separately?


Creating an endpoint for this later on would be nice. But it's probably better to make this a new change.

felixguendling:
> Should earliest and latest transport extend up to the maximal interval defined in the feed? 

Keep it simple. First and last bit in the traffic day bitfields should be enough. For now also for Flex. If we have a use case to report Flex separately, this can easily be changed. For now it's just about a minimal initial version that's as minimal and simple as possible.

> Creating an endpoint for this later on would be nice

Not sure why we need a separate endpoint? We already have a metrics endpoint.

MichaelKutzner:
> > Should earliest and latest transport extend up to the maximal interval defined in the feed?
> 
> Keep it simple. First and last bit in the traffic day bitfields should be enough. For now also for Flex. If we have a use case to report Flex separately, this can easily be changed. For now it's just about a minimal initial version that's as minimal and simple as possible.
>

Thank you for the fast response.

Suggesting the bitfields was quite helpful. I already thought, that my approach would be too convoluted.

About handling the flex trips separate: While my first version was pretty simple, it didn't include flex trips. So I wasn't sure, if we need to handle them differently.
 
> > Creating an endpoint for this later on would be nice
> 
> Not sure why we need a separate endpoint? We already have a metrics endpoint.

Sorry about that. I assumed you wanted to make it a different endpoint, as the data doesn't change between imports. Adding it to the existing one is fine for me too.

-----

[...]

PR #970, opened 2025-08-06
Add timetable metrics

MichaelKutzner:
Timetable metrics will now be written to the data directory after an import. They will also be available at the metrics endpoint. Both features were suggested in #935.

Example for `data/timetable_metrics.json`:

```json
[{"idx":0,"firstDay":"2205-01-04","lastDay":"2025-07-31","noLocations":6102,"noTrips":20826,"transportsXDays":0},{"idx":1,"firstDay":"2025-08-03","lastDay":"2025-12-14","noLocations":514078,"noTrips":2375352,"transportsXDays":89051313},{"idx":2,"firstDay":"2025-08-04","lastDay":"2025-11-04","noLocations":138984,"noTrips":319960,"transportsXDays":21136},{"idx":3,"firstDay":"2025-08-03","lastDay":"2025-10-07","noLocations":664,"noTrips":2075,"transportsXDays":35941}]
```

Metric endpoint example:

```
timetable_metrics_info{first_day="2025-08-03",last_day="2025-12-14",no_locations="514078",no_trips="2375352",src="1",tag="delfi",transports_x_days="89051313"} 1
```

Notice that the `timetable_metrics.json` file does not contain any tags, as these are not contained in the timetable.

For the Prometheus data, tags are available. But I'm still not sure, if it's reasonable to expose all metrics in a single _Info_ object. Or we should split it into multiple Counter or Gauge objects, e.g.

```
timetable_metrics_count{first_day="2025-08-03",last_day="2025-12-14",src="1",tag="delfi",field="no_locations"} 514078
timetable_metrics_count{first_day="2025-08-03",last_day="2025-12-14",src="1",tag="delfi",field="no_trips"} 2375352
timetable_metrics_count{first_day="2025-08-03",last_day="2025-12-14",src="1",tag="delfi",field="transports_x_days"} 89051313
```
 
----

Reminder: As the timetable format didn't change, increasing the binary version is not needed.


src/endpoints/metrics.cc
```diff
@@ -113,6 +115,24 @@ void update_all_runs_metrics(nigiri::timetable const& tt,
...
   }
+
+  if (metrics.timetable_metrics_.Collect().empty()) {
+    auto const m = get_metrics(tt);
+    auto const from = std::chrono::time_point_cast<date::sys_days::duration>(
+        tt.internal_interval().from_);
+    for (auto src = n::source_idx_t{0U}; src != tt.n_sources(); ++src) {
+      auto const& fm = m.feeds_[src];
+      auto const labels = prometheus::Labels{
+          {"src", fmt::format("{}", src)},
```
felixguendling:
```suggestion
```

src/endpoints/metrics.cc
```diff
@@ -113,6 +115,24 @@ void update_all_runs_metrics(nigiri::timetable const& tt,
...
+          {"tag", std::string{tags.get_tag(src)}},
+          {"first_day", fmt::format("{:%F}", from + date::days{fm.first_})},
+          {"last_day", fmt::format("{:%F}", from + date::days{fm.last_})},
+          {"no_locations", fmt::format("{}", fm.locations_)},
+          {"no_trips", fmt::format("{}", fm.trips_)},
+          {"transports_x_days", fmt::format("{}", fm.transport_days_)}};
```
felixguendling: separate metrics with tag as label

MichaelKutzner: I will update this.
Any suggestions how to handle the first and last day? I don't think using a counter for days since epoch would be easy to read. But I don't see a better approach for now.

felixguendling: Gauge should work? timestamps in seconds since 1970 work fine

MichaelKutzner: I updated the code. Using unixtime here works. But as we only store days, the timestamp for the last transport day might be misleading. Should I change the time to show `23:59:59` instead?

src/endpoints/metrics.cc
MichaelKutzner: This is now removed

src/metrics_registry.cc
```diff
@@ -91,6 +91,30 @@ metrics_registry::metrics_registry(
...
               .Help("The number of currently running trips that have RT data")
+              .Register(registry_)},
+      timetable_first_day_unixtime_{
+          prometheus::BuildGauge()
+              .Name("timetable_first_day_unixtime")
+              .Help("Day of the first transport in unixtime")
+              .Register(registry_)},
+      timetable_last_day_unixtime_{
+          prometheus::BuildGauge()
+              .Name("timetable_last_day_unixtime")
```
felixguendling: since in other places, the unit is part of the name, maybe follow this naming logic to be consistent
example:
- `nigiri_vdvaus_last_update_timestamp_seconds`
- `nigiri_gtfsrt_last_update_timestamp_seconds`

Instead of `timetable`, existing metrics seem to use `nigiri`. So in order to stay consistent, it might make sense to keep this naming logic. At least that's easier than changing all existing metrics because there are already Grafana Dashboards configured to use the current names.

src/endpoints/metrics.cc
felixguendling: probably depends on how the dashboards can be configured. I'm not an expert in this regard. Probably it doesn't matter. But maybe @traines-source has some insight?

src/metrics_registry.cc
MichaelKutzner: You are right, I missed the prefix. Using `timetable` seemed to some degree reasonable, as I mostly compared to [metrics_registry.h](https://github.com/motis-project/motis/blob/master/include/motis/metrics_registry.h). So I guess, we probably want to add both.

Using a consistent unit would be fine. But I also don't want to confuse users, who might assume there were no transports after midnight of the final day. If setting the time as well is fine, this probably wouldn't be an issue.
I think, I will wait for feedback on the previous comment before updating this.

src/endpoints/metrics.cc
traines-source: Yeah I think 23:59:59 should be fine, or just the next day (exclusivity of upper bound?). In a dashboard maybe we would display it as "expires in x days" anyways.

src/metrics_registry.cc
felixguendling: I meant `timetable_last_day_unixtime` -> `timetable_last_day_timestamp_seconds`

src/endpoints/metrics.cc
MichaelKutzner: Thank you for the feedback. I decided to go with `23:59:59` for now, as this probably is the last ambiguous variant. Using the next day would be reasonable as well, if this makes handling the dashboards easier. But then we need to change the description, to make it clear, that the timestamp is exclusive.

src/metrics_registry.cc
MichaelKutzner: It's updated now. I just didn't think, that `_seconds` is suitable, when the last day timestamps were the beginning of the day. At least for me, this would have implied a more precise timestamp than just the day. As the timestamp is now the end of the day (or start of the next day), the naming is fine, I think.

MichaelKutzner:
After applying all changes, the metrics now look like this:
```
nigiri_timetable_first_day_timestamp_seconds{tag="delfi"} 1754179200
nigiri_timetable_last_day_timestamp_seconds{tag="delfi"} 1765756799
nigiri_timetable_locations_count{tag="delfi"} 514078
nigiri_timetable_trips_count{tag="delfi"} 2375352
nigiri_timetable_transports_x_days_count{tag="delfi"} 89051313
```

PR #971, opened 2025-08-06
Add option to set log level

MichaelKutzner:
This adds an option to set the logging level, as it's currently hard coded as `debug` (https://github.com/motis-project/utl/blob/master/src/logging.cc#L5). And I didn't find any issues or comments on making this configurable.
To make usage easier, the `server` subcommand now supports the `--help` option.


exe/main.cc
```diff
@@ -106,15 +108,33 @@ int main(int ac, char** av) {
...
+          if (ll == "ERROR"sv) {
+            utl::log_verbosity = utl::log_level::error;
+          } else if (ll == "INFO"sv) {
+            utl::log_verbosity = utl::log_level::info;
+          } else if (ll == "DEBUG"sv) {
+            utl::log_verbosity = utl::log_level::debug;
+          } else {
+            std::cerr << fmt::format("Unsupported log level '{}'\n", ll);
+            return_value = 1;
+            break;
```
felixguendling, line 145: server will still be started? shouldn't this quit?

exe/main.cc
```diff
@@ -106,15 +108,33 @@ int main(int ac, char** av) {
...
+          std::transform(ll.begin(), ll.end(), ll.begin(),
+                         [](unsigned char const c) { return std::toupper(c); });
+          if (ll == "ERROR"sv) {
+            utl::log_verbosity = utl::log_level::error;
+          } else if (ll == "INFO"sv) {
+            utl::log_verbosity = utl::log_level::info;
+          } else if (ll == "DEBUG"sv) {
+            utl::log_verbosity = utl::log_level::debug;
+          } else {
+            std::cerr << fmt::format("Unsupported log level '{}'\n", ll);
```
felixguendling: fmt::println is sufficient

no need to combine stream `<<` and `fmt`

exe/main.cc
```diff
@@ -1,12 +1,14 @@
...
 #include "boost/url/decode_view.hpp"
 
+#include <cctype>
 #include <filesystem>
 #include <iostream>
 #include <string_view>
 
 #include "google/protobuf/stubs/common.h"
 
+#include "utl/logging.h"
```
felixguendling, line 13: most of the logging is done by nigiri, not utl
https://github.com/motis-project/nigiri/blob/master/include/nigiri/logging.h

we didn't proceed on the unified logging

exe/main.cc
MichaelKutzner: It already terminates, if an invalid log level is used.

exe/main.cc
MichaelKutzner: Thank you, I somehow missed this, as setting `utl::log_verbosity` already removed many message for me. But `nigiri::s_verbosity` seems unset?
If it's different for now, what are the plans for this in the future? I can change the code, to update both variables. I also need to check, which messages will be affected.

felixguendling: the plan would be to switch everything to utl logging which at one point could be made configurable to log to a OTEL collector or console or both as required

there's also https://github.com/motis-project/nigiri/pull/165

felixguendling: but for this PR it should be sufficient to configure the existing mechanisms

exe/main.cc
MichaelKutzner: Updated

exe/main.cc
MichaelKutzner: Unless I'm still missing something, both should now be configurable.
As configurable logging destinations seem useful for at least imports as well, I already moved the code to a new file. If this isn't necessary yet, I can revert this.

PR #975, opened 2025-08-09
replace dummy leg with surface-level routing fallback

### LINKED ISSUE #966: Impossible interchanges in Motis
During planning my journey, I found out that Motis proposed an interchange between bus and train, with stops on opposite sides of river, with no bridge and no footpath calculated
Image

Here's the link to itinerary:

https://api.transitous.org/?time=2025-08-02T23%3A37%3A00.000Z&fromPlace=lu-administration-des-transports-publics_000210109002&toPlace=de-DELFI_de%3A10042%3A8004771&withFares=true&fastestDirectFactor=1.5&joinInterlinedLegs=false&maxMatchingDistance=250&fromName=Waldbredimus%2C+Kiirch%2C+Bous-Waldbredimus&toName=Perl%2C+Perl

Edit: moving the destination on the map gives an even more weird itinerary...

DomeQdev:
This PR improves the fallback mechanism for pedestrian routing when a path cannot be found, particularly between different levels (e.g., underground to surface).

Currently, a failed routing attempt results in a "dummy leg" based on a straight-line distance, which is unrealistic and misleading for the user.

Fixes #966

### DIFF
--- a/src/street_routing.cc
+++ b/src/street_routing.cc
@@ -219,12 +219,34 @@ api::Itinerary street_routing(osr::ways const& w,
       out.is_time_dependent() ? bound_time : n::unixtime_t{n::i32_minutes{0}}};
   auto const path = utl::get_or_create(cache, cache_key, [&]() {
     auto const& [e_nodes, e_states] = *s;
-    return osr::route(
+    auto p = osr::route(
         w, l, out.get_profile(), from, to,
         static_cast<osr::cost_t>(max.count()), osr::direction::kForward,
         max_matching_distance,
         s ? &set_blocked(e_nodes, e_states, blocked_mem) : nullptr,
         out.get_sharing_data(), elevations, osr::routing_algorithm::kAStarBi);
+
+    if (!p.has_value()) {
+      osr::location from_surface = from;
+      from_surface.lvl_ = osr::level_t{0.0f};
+
+      osr::location to_surface = to;
+      to_surface.lvl_ = osr::level_t{0.0f};
+
+      p = osr::route(
+          w, l, out.get_profile(), from_surface, to_surface,
+          static_cast<osr::cost_t>(max.count()), osr::direction::kForward,
+          max_matching_distance,
+          s ? &set_blocked(e_nodes, e_states, blocked_mem) : nullptr,
+          out.get_sharing_data(), elevations, osr::routing_algorithm::kAStarBi);
+
+      if (p.has_value()) {
+        constexpr auto const kLevelChangePenaltySeconds = 120;
+        p->cost_ += kLevelChangePenaltySeconds;
+      }
+    }
+
+    return p;
   });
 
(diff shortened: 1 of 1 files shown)


felixguendling:
Any concrete example where this helps?

As with the other PR:

I would also recommend to discuss your approach before starting to work on something for several reasons:

- what's the best way to implement it?
- who else might already be working on something in the same direction?
- does this have chances to be merged at all?

felixguendling:
I'm pretty sure this does not fix https://github.com/motis-project/motis/issues/966 as there we don't want any footpath to be found at all.

DomeQdev:
> Any concrete example where this helps?

Yes: https://api.transitous.org/?time=2025-08-10T13%3A16%3A00.000Z&fromPlace=52.24055097297125%2C20.896091645223777&toPlace=53.127698421641725%2C23.12741118266274&withFares=true&fastestDirectFactor=1.5&joinInterlinedLegs=false&maxMatchingDistance=250&fromName=52.24055097297125%2C20.896091645223777%2C0

There is no footpath between Rondo ONZ and Warszawa Centralna

image


> as there we don't want any footpath to be found at all.

why?

felixguendling:
> why?

It requires a 1,5h walk to the next bridge. This is not something we want to consider for transfers. It's not about how the transfer is being displayed but more about the wrong transfer time and that we have a transfer time limit that's way below 1,5h.

DomeQdev:
Okay, so if we knew the travel time was 1.5h, we wouldn’t show it at all.

felixguendling:
Not just that. Since this unrealistic journey probably dominated other (more useful and realistic) alternatives during the routing, we would maybe even show something more useful instead. In order to show these alternatives, we are not allowed to produce a transfer here at all (at least not with 2min when it actually takes 1,5h).

Besides that it doesn't solve #966 regarding this actual PR:
- please use "AAA style" (see docs/STYLE.md) which means the left hand side of an assignment is always `auto` or `auto const` if possible
- why not use the wildcard level which matches any level. This should probably be even more robust than fixing the level to ground level?

DomeQdev:
How do you plan to fix this when some OSM data is simply broken for certain stations (e.g. Rondo ONZ in Warsaw)?

How will you distinguish between missing stairs that connect the platform with the surface, and missing a bridge across a river?

felixguendling:
OSM is not set in stone. More or less the opposite is the case: everyone is invited to fix broken data. So "simply broken" today could mean that it's already fixed tomorrow. That would always be the preferred option as it provides the best information to users.

However, I am not opposing this PR to make the routing for the path display more robust against broken OSM data.

With a transfer time limit of 20min the case in Warsaw will produce a valid transfer while the bridge case will still find anything on OSM data. The primary problem with the bridge case is that no routing was involved at all because we don't have `osr_footpath` setting active in Transitous yet.

PR #991, opened 2025-09-01
fix(api): properly set displayName in stop_times

networkException:
While the `displayName` property was introduced to the `StopTime` OpenAPI type, it wasn't properly set by the implementation.


felixguendling:
Thanks, seems like I missed that. Maybe you can run the formatter (like in the CI file)? Then I can merge it.

PR #997, opened 2025-09-05
Handle trip_ids containing spaces in shape analysis

MichaelKutzner:
Fixes an issue, if a `trip_id` contains spaces.
As the web UI encodes spaces as `+`, the analyzer needs to decode these correctly.
**Example**: `20250905_15%3A53_au-Translink-SEQ_33394889-SBL+25_26-39719`

### DIFF
--- a/exe/main.cc
+++ b/exe/main.cc
@@ -259,7 +259,9 @@ int main(int ac, char** av) {
         auto const ids = utl::to_vec(
             vm["trip-id"].as<std::vector<std::string> >(),
             [](auto const& trip_id) {
-              auto const decoded = boost::urls::decode_view{trip_id};
+              // Set space_as_plus = true
+              auto const opts = boost::urls::encoding_opts{true};
+              auto const decoded = boost::urls::decode_view{trip_id, opts};
               return std::string{decoded.begin(), decoded.end()};
             });
 


felixguendling:
Is this a problem also in the rest of the code? Or does it get handled correctly there?

MichaelKutzner:
I tested various variants through the Web interface, but I didn't notice any issues. I haven't tested the API directly, though. I will check, if there is an error.

MichaelKutzner:
After checking the API and code more closely, I don't think, that this issue occurs anywhere else in the code.

For the API:
A `tripId` may encode spaces (`%20`) or use a plus sign (`+`). Both will return the same result. Using a space is not possible, as this would be malformed.

For the code:
I couldn't find any other code, that uses `boost::urls::decode_view` or `boost::urls::encoding_opts` directly. Instead classes like `boost::urls::params_view` are used.
As we don't use or pass any `encoding_opt`, these instances will always set `spaces_as_plus`, which handles this issue:
https://github.com/boostorg/url/blob/8d5cc6e34884b4f3a8a43c8f9706c0f5800238a0/include/boost/url/params_view.hpp#L242-L244
and
https://github.com/boostorg/url/blob/8d5cc6e34884b4f3a8a43c8f9706c0f5800238a0/src/params_view.cpp#L37-L45

PR #1022, opened 2025-09-18
add option to fetch other stops on each run when querying stoptimes

schaerfo:
A while ago, I proposed in the Matrix channel to add an option to extend the stop times endpoint such that other stops are also returned for each found trip, eliminating the need for separate calls to the trip endpoint.

The purpose was to build a replacement for direkt.bahn.guru, which is already online at direct-trains.eu. Now, I got finally around to prepare the changes into a PR.

* The API was modeled after hafas-client. Consequently, only the subsequent/previous stops are returned when looking up departures/arrivals. Also, the keys in the JSON response differ.
* It would also be possible to return all stops on the trip (and make the response more similar to the trip endpoint). I did not choose this option because I figured that this way would be more useful in most situations as you don't need to look for the requested stop in the response.
* Interlined legs are joined.
* I marked the new API as experimental in case we need to change it

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1379,6 +1379,17 @@ paths:
             If set to `true`, only stations that are phyiscally in the radius are considered.
             If set to `false`, additionally to the stations in the radius, equivalences with the same name and children are considered.
 
+        - name: fetchStops
+          in: query
+          schema:
+            type: boolean
+          required: false
+          description: |
+            Experimental. Expect unannounced breaking changes (without version bumps).
+
+            Optional. Default is `false`. If set to `true`, the following stops are returned
+            for departures and the previous stops are returned for arrivals.
+
         - name: pageCursor
           in: query
           required: false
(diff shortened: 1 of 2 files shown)


felixguendling:
Thanks! This looks good! :+1: 

> Interlined legs are joined.

This will return hundreds of stops in some cases. Maybe it could help to check if the route is circular and then only return one cycle?


> I marked the new API as experimental in case we need to change it

It's likely that we have to change it a bit in the future if we want to properly support merge/split trips.

schaerfo:
> > Interlined legs are joined.
> 
> This will return hundreds of stops in some cases. Maybe it could help to check if the route is circular and then only return one cycle?

Yes, that would make sense. I guess this would be relevant for the Ringbahn lines of the Berlin S-Bahn

felixguendling:
If you click on a random route in San Francisco there's also a high likelihood that you will hit a route with >200 stops going back and forth. Probably the `block_id` connection in those datasets has only operational reasons and isn't intended for passenger information. But we currently don't have any logic to detect which `block_id` connections to discard and which to keep. So we probably have a lot of `block_id` connections we don't want.

In case of the Ringbahn we want it and it's a good example where stopping after the first loop would make sense.

I merged it because it's marked as experimental anyway and we can still change the logic and/or response format as long as it's not declared stable.

PR #1027, opened 2025-09-19
add details on maxTravelTime option of one-to-all to openapi.yaml and error message

e-kotov:
This extends the error message when `maxTravelTime` in `one-to-all` endpoint exceeds the pre-configured value and adds similar information to the openapi spec file.


e-kotov:
I reformatted the `cc` file with clang-format. I did not know I needed to to that. Hopefully this will pass the formatting check now.

traines-source:
Thanks :+1: Docs improvements are always welcome!

Issue #964, opened 2025-08-01
UI: fix time display bug for next day time


felixguendling:
<img width="514" height="341" alt="Image" src="https://github.com/user-attachments/assets/bd30fd54-fbd2-4e6c-a508-7672c8517735" />

traines-source:
Proposition: When day changes during a trip (or journey?), show a separate line indicating that change, and at the very end ("arrival on {day}, {date}")? Completely remove indicator next to time.

PR #1036, opened 2025-09-23
display tz if different from browser

traines-source:
(with temporary (until we have reworked the layout) hack for #964 since tz also overflows, fallback: title)

### DIFF
--- a/ui/src/lib/ConnectionDetail.svelte
+++ b/ui/src/lib/ConnectionDetail.svelte
@@ -52,7 +52,7 @@
 	<span>
 		{#if p.stopId}
 			<Button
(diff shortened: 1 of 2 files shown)


felixguendling:
image

probably has to be after the real-time if real-time is available?

traines-source:
There's still not enough space and it's ugly when there is no RT... That's why I figured we could live with this state until we have reworked the layout. But maybe there is another temporary solution.

I just think it's quite important to have to avoid confusion, the other day I was confused why it was in browser time again until I realised I hadn't loaded OSM for that area...

felixguendling:
image
image

PR #1045, opened 2025-09-24
GBFS radius

traines-source:


src/endpoints/routing.cc
```diff
@@ -194,12 +196,30 @@ std::vector<n::routing::offset> get_offsets(
...
+                       [](auto const f) {
+                         return f == api::RentalFormFactorEnum::CAR ||
+                                f == api::RentalFormFactorEnum::MOPED ||
+                                f == api::RentalFormFactorEnum::OTHER;
+                       })) ||
+          (propulsion_types.has_value() &&
+           !utl::all_of(*propulsion_types, [](auto const f) {
+             return f == api::RentalPropulsionTypeEnum::HUMAN ||
+                    f == api::RentalPropulsionTypeEnum::ELECTRIC_ASSIST;
+           }))) {
```
pablohoch: I don't think we need the `propulsion_types` checks, because the osr profile used for routing only depends on the rental form factor (`gbfs::get_osr_profile` in `osr_profile.cc`) and this could match e.g. electric scooters (which are probably using `api::RentalPropulsionTypeEnum::ELECTRIC`, but routed like bikes).

Apparently we currently only use `osr::search_profile::kCarSharing` for form factor CAR, so mopeds are currently treated like bikes as well. Not sure if that is correct or not (Wikipedia says "Mopeds typically travel only slightly faster than bicycles on public roads.", so maybe...), but it should be consistent.

Another option would be to refactor the `get_osr_profile` function to take `vehicle_form_factor` instead of `provider_products`, convert the API form factor using `vehicle_form_factor from_api_form_factor(api::RentalFormFactorEnum const ff)` and then use the same `get_osr_profile` method in both places (e.g. if there are form factors specified in the API request and it returns `kBikeSharing` for all of them, it is safe to use the bike sharing radius).

traines-source: Ah yes that makes sense, I have changed it accordingly.

Issue #1014, opened 2025-09-15
Enhancements for small screens


traines-source:
* [ ] on mobile devices, have the UI elements be contained in a card that draws up from the bottom (similar to Google Maps, just that the card initially covers everything initially and map is only shown and loaded when swiping it down a bit?). This means that there is one common scrolling bar for the query form (from, to, etc, options, etc) and the query results. On larger screens, everything should stay where it is (the merge of the scrolling bar/the two cards may be acceptable though).
* [ ] the three tabs (Connections, Departures, Isochrones) are exceeding width of small screens currently

PR #1060, opened 2025-09-29
UI enhancements

DIDONAX:
- display track code in every view including map tootip and display early rt trip and times red.

### DIFF
--- a/ui/src/lib/ConnectionDetail.svelte
+++ b/ui/src/lib/ConnectionDetail.svelte
@@ -379,9 +379,7 @@
 		{:else if !(isLast && !isRelevantLeg(l)) && ((i == 0 && isRelevantLeg(l)) || !next || !next.displayName || l.mode != 'WALK' || (pred && (pred.mode == 'BIKE' || (l.mode == 'WALK' && pred.mode == 'CAR') || pred.mode == 'RENTAL')))}
 			<Route {onClickTrip} {l} />
 			<div class="pt-4 pl-6 border-l-4 left-4 relative" style={routeBorderColor(l)}>
-				--
 				{@render stopTimes(l.startTime, l.scheduledStartTime, l.realTime, l.from, l.mode, -1, true)}
-				--
 				{#if l.mode == 'FLEX'}
 					<div class="mt-2 flex items-center leading-none">
 						<span class="ml-1 text-sm">


ui/src/lib/ConnectionDetail.svelte
```diff
@@ -57,6 +59,9 @@
 				onclick={() => onClickStop(p.name, p.stopId!, new Date(timestamp))}
 			>
 				{p.name}
+				{#if p.track}
+					- {t.track} {p.track}
+				{/if}
```
traines-source: I think instead of just putting it behind the stop name, it would look better to make it more similar to how the existing track indicators look (right aligned, also similar on e.g. bahn.de). If that takes up too much space, maybe we can shorten the "Gleis" label to Gl./Pl./voie/.... (In that context maybe we could also distinguish for buses/trams to "Steig", i.e. Stg./Bay/quai/..., at least in German it's very weird to have a Gleis for buses...)
If it really does use too much space, we could postpone adding track information to the intermediate stops until we have reworked the UI a bit more to work better on smaller screens (#1014 and others)

ui/src/lib/RailViz.svelte
```diff
@@ -270,8 +279,9 @@
 					}
 					return {
 						html: `${object.trips[0].displayName}<br>
-                  ${formatTime(new Date(object.departure), object.from.tz)} ${object.from.name}<br>
-                  ${formatTime(new Date(object.arrival), object.to.tz)} ${object.to.name}`
+						${object.from.track ? object.from.track + '<br>' : ''}
```
traines-source: I think in Railviz we don't necessarily need the track info (sorry for imprecise issue), but if we put it, it should appear behind the corresponding stop name I guess.

ui/src/lib/StopTimes.svelte
```diff
@@ -125,6 +125,10 @@
 							{#if !stopTime.headsign || !stopTime.tripTo.name.startsWith(stopTime.headsign)}
 								({stopTime.tripTo.name})
 							{/if}
+
+							{#if stopTime.place.track}
+								- {t.track} {stopTime.place.track}
+							{/if}
```
traines-source: Same as in ConnectionDetail. Also, the times look somehow misaligned vertically, this is already due to https://github.com/motis-project/motis/commit/93114888e71e6e2e97acfd24f108c3ace59fb915 I think, but maybe you can fix/improve it?

ui/src/lib/Time.svelte
```diff
@@ -19,13 +20,15 @@
...
 		timeZone: string | undefined;
+		isArrival?: boolean | undefined;
 	} = $props();
 
 	const t = $derived(new Date(timestamp));
 	const scheduled = $derived(new Date(scheduledTimestamp));
 	const delayMinutes = $derived((t.getTime() - scheduled.getTime()) / 60000);
 	const highDelay = $derived(isRealtime && delayMinutes > 3);
 	const lowDelay = $derived(isRealtime && delayMinutes <= 3);
+	const Early = $derived(isRealtime && delayMinutes < -1);
```
traines-source: In JS, we use lowerCaseCamelCase for variable names.

ui/src/lib/Time.svelte
```diff
@@ -59,7 +62,11 @@
 		</div>
 		<div class="text-xs font-normal h-4">{isSameAsBrowserTimezone ? '' : timeZoneOffset}</div>
 	{:else if variant === 'realtime-show-always' || (variant === 'realtime' && isRealtime)}
-		<span class:text-destructive={highDelay} class:text-green-600={lowDelay}>
+		<span
+			class:text-destructive={isArrival ? highDelay : Early}
+			class:text-green-600={isArrival ? lowDelay : lowDelay && !Early}
```
traines-source: If high and lowDelay aren't used anywhere else, wouldn't it be easier to merge this logic into lines 29-30, and rename to say `notOnTime` and `roughlyOnTime` or something similar?

ui/src/lib/Time.svelte
```diff
@@ -59,7 +62,11 @@
...
+			class="bg-white"
```
traines-source: I think `class="bg-white"` has been removed on purpose because it looks ugly/breaks in dark mode

ui/src/lib/RailViz.svelte
```diff
@@ -136,18 +139,21 @@
...
-		const getDelayColor = (d: number, realTime: boolean): RGBA => {
-			const delay = d / 60000;
+		const getDelayColor = (d: number, a: number, realTime: boolean): RGBA => {
+			const departureDelay = d / 60000;
+			const arrivalDelay = a / 60000;
 			if (!realTime) {
 				return [100, 100, 100, 255];
 			}
-			if (delay <= 3) {
+			if (departureDelay < -1) {
```
traines-source: I.e. -1 minute would still be green right?

ui/src/lib/ConnectionDetail.svelte
DIDONAX: as you said i put it behind the stop name because it took too much space, as for the other suggestion it could be implemented. I also need to mention that I noticed some track codes already have prefixes (Gl.) and others don't so we would have some inconsistencies like Gleis Gl.1 or Gleis 1 or Gl.Gl1 (if we want to shortened version), if we decide to just put the track code then again we would have inconsistencies like 1 or Gl.1 which takes less space but not ideal, i thought maybe we extract the track code from the p.track then append it to whatever label we want ?

traines-source: Yes ok. I think extracting the actual track is something that should rather be done with the [new scripting feature](https://github.com/motis-project/motis/blob/master/docs/scripting.md) in the backend, since it's a feed-specific thing and difficult to generalize (there are lots of track codes with letters, too...) For e.g. Transitous then also the API consumers can benefit from it. For the time being, I think we can live with "Gl. Gleis 5" quirks.
Maybe we can also reduce the font size of the track indicator to save some space.

ui/src/lib/RailViz.svelte
DIDONAX: I considered the track code information to be just as important as the stop times, which is why I displayed it with them. It also seemed cool to hover over a vehicle and see which track it’s currently on.

traines-source: Mmh but like that it isn't really clear whether it's the departure or arrival track? And I think noone is really using Railviz to find out where exactly their train departs (I hope).

ui/src/lib/ConnectionDetail.svelte
traines-source: That looks much better. But there should at least be a tiny margin to the right, on mobile the track label is even cut off:

image

Also, not sure why some RT times are grey now, while others aren't.

ui/src/lib/Time.svelte
traines-source: I meant/thought it would simplify the logic to put it into one expression. But maybe that's not actually the case. Anyways it's not that important.

ui/.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules
 /build
+/.vscode
```
traines-source:
```suggestion
.vscode
```

ui/.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
traines-source: Please don't commit this. If you change /.vscode to .vscode as proposed above, this should be ignored as well.

ui/src/lib/ConnectionDetail.svelte
```diff
@@ -32,66 +32,81 @@
...
+					<Button
+						class="text-[length:inherit] leading-none justify-normal text-wrap p-0 text-left {textColor}"
+						variant="link"
+						onclick={() => onClickStop(p.name, p.stopId!, new Date(timestamp))}
+					>
+						{p.name}
+					</Button>
+					{#if p.track}
+						<span class="text-nowrap px-1 border text-xs rounded-xl">
+							{t.track.slice(0, 2) + '.'}
```
traines-source: I don't think that this is applicable to all languages. Already in French, I think "Vo." is not really a useful abbreviation? Either change the translations to the abbreviations, or, if these translations are used elsewhere where we want the full label, introduce a new translation key.
Edit: Ah ok if bahn.de abbreviates it with "Vo." it must be correct I guess :D Still I don't think we want to rely on this working for all languages.

ui/src/lib/Time.svelte
```diff
@@ -19,13 +20,18 @@
...
 	} = $props();
 
 	const t = $derived(new Date(timestamp));
 	const scheduled = $derived(new Date(scheduledTimestamp));
 	const delayMinutes = $derived((t.getTime() - scheduled.getTime()) / 60000);
 	const highDelay = $derived(isRealtime && delayMinutes > 3);
 	const lowDelay = $derived(isRealtime && delayMinutes <= 3);
+	const early = $derived(isRealtime && delayMinutes <= -1);
+	const notOnTime = $derived(isArrival ? highDelay : early);
+	const roughlyOnTime = $derived(isArrival ? lowDelay : lowDelay && !early);
```
traines-source: I meant/thought it would simplify the logic to put it into one expression directly for `notOnTime` and one for `roughlyOnTime`. But maybe that's not actually the case. Anyways it's not that important.

ui/src/lib/ConnectionDetail.svelte
DIDONAX: yes it was just a make shift solution, planning on adding translation today.

ui/.gitignore
DIDONAX: yeah i didn't mean to do that, typo mistake i guess.

traines-source:
image

This looks a bit weird. For the footpath we should not show any platform indicator I guess. I'm undecided where to show the platform when boarding a trip, but if we inspire ourselves from DB Navigator and RMVgo, we would remove the larger platform indicator altogether (maybe make the one next to the departure stop (here Tübingen Nonnenhaus) larger in return)? For the arrival of a trip, we can definitely remove the "Ankunft auf Gleis..." part.

ui/src/lib/ConnectionDetail.svelte
```diff
@@ -30,68 +30,88 @@
...
+						<span class="ml-1 leading-none">
+							{pickupNotAllowedOrEnd && dropoffNotAllowedOrStart
+								? t.inOutDisallowed
+								: pickupNotAllowedOrEnd
+									? t.inDisallowed
+									: t.outDisallowed}
+						</span>
+					</div>
+				{/if}
+				{#if p.alerts}
```
traines-source: why has `isStartOrEnd` been removed here? I don't think we want to show alerts on every intermediate stop. There are feeds (e.g. [Réunion](https://github.com/public-transport/transitous/blob/main/feeds/fr.json#L1001-L1019)) where there iare loads of alerts for every stop.

traines-source (approved):

PR #1067, opened 2025-09-30
translation: label REGIONAL_FAST without regional

JanZerebecki:
as the term regional may confuse one into thinking the train can be used with a ticket of regional public transport. In Germany there was special confusion, as for those routes I checked, the class of RegionalExpress trains in the data is labeled with REGIONAL_RAIL (IMHO correctly), and REGIONAL_FAST_RAIL was translated with RegionalExpress. I avoided to use the term InterRegio, as in Switzerland those are REGIONAL_RAIL, but the German railway company uses the term "InterRegio / fast train" for the lowest class of non-regional long distance. AFAIK German railway uses the "fast train" classification only for foreign trains.

The definition in openapi.yaml matches with
https://en.wikipedia.org/wiki/Express_train and a few of the other languages linked for that article. And I have not seen the term further differentiated with something like regional, only with a specific line.

I'm told REGIONAL_FAST_RAIL is also used as a fallback where there is no data and that the structure of the data is
not expressive enough to capture the ticket price structure, but as it is a few test routings on https://transitous.org/ revealed no conflicting usage of REGIONAL_FAST_RAIL. (Whether or not some of these should be a "faster" classification, does not pose any problem here.)

Whether it is because the origin on the data already fits this classification system and/or the fall back is selected well, but overall I think omitting regional is more likely to lead people to select the route options they intent.

### DIFF
--- a/ui/src/lib/i18n/de.ts
+++ b/ui/src/lib/i18n/de.ts
@@ -82,7 +82,7 @@ const translations: Translations = {
 	HIGHSPEED_RAIL: 'Hochgeschwindigkeitszug',
 	LONG_DISTANCE: 'Intercityzug',
 	NIGHT_RAIL: 'Nachtzug',
-	REGIONAL_FAST_RAIL: 'Regionalexpresszug',
+	REGIONAL_FAST_RAIL: 'Schnellzug',
 	REGIONAL_RAIL: 'Regionalzug',
 	OTHER: 'Andere',
 	routingSegments: {
--- a/ui/src/lib/i18n/cz.ts
+++ b/ui/src/lib/i18n/cz.ts
@@ -83,7 +83,7 @@ const translations: Translations = {
 	HIGHSPEED_RAIL: 'Vysokorychlostní železnice',
 	LONG_DISTANCE: 'Dálková železnice',
 	NIGHT_RAIL: 'Noční vlaky',
-	REGIONAL_FAST_RAIL: 'Zrychlená regionální železnice',
+	REGIONAL_FAST_RAIL: 'Zrychlená železnice',
 	REGIONAL_RAIL: 'Regionální železnice',
 	OTHER: 'Jiné',
 	routingSegments: {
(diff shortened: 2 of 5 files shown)


ui/src/lib/i18n/de.ts
```diff
@@ -82,7 +82,7 @@ const translations: Translations = {
 	HIGHSPEED_RAIL: 'Hochgeschwindigkeitszug',
 	LONG_DISTANCE: 'Intercityzug',
 	NIGHT_RAIL: 'Nachtzug',
-	REGIONAL_FAST_RAIL: 'Regionalexpresszug',
+	REGIONAL_FAST_RAIL: 'Schnellzug',
```
felixguendling, line 85: This term is not really used in Germany? D-Zug is not a thing anymore.

traines-source:
As a matter of fact, for GTFS feeds, [REGIONAL_FAST_RAIL is only used for route_type 2](https://github.com/motis-project/nigiri/blob/95f89d4ac2b8b22e5e16424084e9700b8edd58c0/src/loader/gtfs/route.cc#L24), which is described as "Rail. Used for intercity or long-distance travel." – crucially without there being any other rail type for "regional" trains if the extended route types are not used. I.e. it can be anything (and e.g. in France it is used like that, there are both TGVs and local trains with route_type 2).
InterRegio is AFAIK actually part of the long distance offering of SBB, but it doesn't really matter since there is no fare difference.
So yeah, I agree that we should remove the "regional" from the labels.

felixguendling:
Another option would be to map route_type=2 to something else?
Mapping `route_type` to clasz was a quite random "best effort" decision back then.

traines-source:
Since REGIONAL_FAST_RAIL, esp. after the label changes, is the class that best reflects that it could be anything from local to high speed train, I don't see a better option (`kOther` is unnecessarily unspecific IMO). Conversely, one could check whether it would make sense to also classify 113 and 117 as `kRegionalFast`.

traines-source:
ahem but maybe we should also update the description in openapi.yaml?

felixguendling:
> ahem but maybe we should also update the description in openapi.yaml?

:+1:

ui/src/lib/i18n/de.ts
JanZerebecki: Yes, except as a classification on the routing website of German railway. Interestingly the international trains crossing into Germany that showed up in there in my few tests were also in REGIONAL_FAST_RAIL on https://transitous.org/ .

PR #1074, opened 2025-10-07
Add logging configuration

MichaelKutzner:
This adds a logging section to the configuration file.
Using the configuration file instead of command line options provides a more coherent server configuration, that can also easily be shared.

Changes to notice:

* `logging.log_level` needs to be lowercase. Using upper case and mixed case are not supported, to make the file easier to read. (Using uppercase only would be fine for me as well)
* When using both the configuration file and command line options, the command line option will be picked
    * The configuration file setting will still be checked and the server will shutdown, if an invalid `log_level` is set
* For symmetric reasons, the `log_level` also affects the `import` mode. This prevents an import to succeed with an invalid `log_level`.

### DIFF
--- a/include/motis/logging.h
+++ b/include/motis/logging.h
@@ -1,13 +1,10 @@
 #pragma once
 
 #include <string>
-#include <string_view>
 
 #include "motis/config.h"
 
 namespace motis {
-
-int set_log_level(std::string_view log_lvl);
 
 int set_log_level(config const&);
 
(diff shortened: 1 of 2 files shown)


include/motis/logging.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "utl/logging.h"
+
+#include "nigiri/logging.h"
+
+#include "motis/config.h"
+
+namespace motis {
+
+inline int set_log_level(std::string_view log_lvl) {
```
felixguendling: nothing templated - this can all go into a .cc file?

MichaelKutzner: Sure, that's no problem. I've updated it.

PR #1076, opened 2025-10-07
add alert description and impact period

DIDONAX:


ui/src/lib/ConnectionDetail.svelte
```diff
@@ -101,11 +102,31 @@
...
-						<div class="text-destructive text-sm">
-							{alert.headerText}
-						</div>
-					{/each}
+					<details class="text-destructive text-sm">
+						<summary
+							class="flex items-center gap-2 hover:underline underline-offset-4 cursor-pointer transition-all"
+						>
+							<TriangleAlert class="size-4" />
+							<span> View alerts </span>
```
traines-source: This needs translation, maybe we can reuse the "Service alerts present" / `alertsAvailable` translation, but then it would need to be made clearer that this can be expanded (with a chevron/arrow similar to the intermediate stops?) 
Also, I think if we're at it, the alerts should also be displayed in stoptimes, instead of just the "Service alerts present" info.

ui/src/lib/ConnectionDetail.svelte
```diff
@@ -101,11 +102,31 @@
...
+						</summary>
+						{#each p.alerts as alert, i (i)}
+							<button
+								class="text-sm text-left w-full truncate"
+								onclick={(e) => {
+									e.currentTarget.classList.toggle('truncate');
```
traines-source: Do you feel it's sufficiently clear that one can click (or in particular on mobile, where there is no cursor indication, tap) on this to show the full message, or should there be a chevron/arrow as well?
Maybe also add a tiny margin/padding to the right

ui/src/lib/ConnectionDetail.svelte
```diff
@@ -323,14 +344,33 @@
...
+								}}
+							>
+								• {alert.headerText}
+								{#if alert.impactPeriod}
+									{alert.impactPeriod[0]?.start} - {alert.impactPeriod[1]?.end}
+								{/if}
+								{#if alert.descriptionText}
+									{alert.descriptionText}
+								{/if}
+							</button>
```
traines-source: Since all of this is duplicate code now (also for stoptimes if we add it there as well), maybe introduce a new component or at least snippet?

ui/src/lib/ConnectionDetail.svelte
```diff
@@ -101,11 +102,31 @@
...
+							<button
+								class="text-sm text-left w-full truncate"
+								onclick={(e) => {
+									e.currentTarget.classList.toggle('truncate');
+									e.currentTarget.classList.toggle('w-full');
+								}}
+							>
+								• {alert.headerText}
+								{#if alert.impactPeriod}
+									{alert.impactPeriod[0]?.start} - {alert.impactPeriod[1]?.end}
```
traines-source: I think it would make sense if this were, like the departure/arrival times, also displayed in the local timezone, by using `formatTime` and `formatDate` from `toDateTime.ts` (or creating a corresponding new function that displays both). Displaying the timezone itself if it differs from the browser timezone is not necessary I guess because it's already displayed right next to it with the departure time.
EDIT: `alert.impactPeriod[1]` : frequently there is only one entry (e.g. Réunion), then this is undefined. So either we should always show the last end (`alert.impactPeriod.length-1`) or even all créneaux ?

traines-source (approved):

PR #1090, opened 2025-10-13
Add parameter to set polygon resolution

MichaelKutzner:
When using queries with large `maxTravelTime`, the browser may fail to compute the union of the reachable polygons (circles). This adds an option to create much simpler polygons, for that the computation is much faster.

For queries with a large `maxTravelTime`, users will most likely use a low zoom level. In that case, having less vertices per circle will hardly be noticeable.
I don't think, that this needs to be selectable in the UI, as users cannot select travel times larger than 4 hours there anyway.

Example usage with `isochronesCircleResolution=4`
Isochrones example isochronesCircleResolution=4

### DIFF
--- a/ui/src/lib/defaults.ts
+++ b/ui/src/lib/defaults.ts
@@ -37,6 +37,7 @@ export const defaultQuery = {
 	additionalTransferTime: 0,
 	transferTimeFactor: 1,
 	numItineraries: 5,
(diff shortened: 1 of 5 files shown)


MichaelKutzner:
I probably should have picked a better name, like `circleVertices` or `polygonVertices`.

PR #1096, opened 2025-10-16
Rental API + Map improvements

pablohoch:
## Rental Provider Groups

- Rental providers can now be added to groups in the configuration.
- If no group is specified for a provider it is added to a group based on the system name reported by the provider (i.e. if multiple providers report the same system name, they are grouped automatically).
- **IMPORTANT:** Do not include commas in the group ids, otherwise they cannot be used for the provider group filter in the routing API. The names can include commas (see config example below, the ids are what is given in "group: ID" and as the key under "groups:" - the name is the same as the ID unless explicity specified). Auto generated group names don't include commas.
- APIs have been adjusted to accept and provide provider group ids in addition to the previously supported provider ids. This includes the various parameters for the plan API.
- The UI now supports the selection of a subset of rental providers to use for routing (first mile, last mile, direct). By default, all providers are enabled.
- The filter buttons on the map now show provider groups instead of single providers.

Config example:
```yaml
gbfs:
  feeds:
    bird-basel:
      url: https://mds.bird.co/gbfs/v2/public/basel/gbfs.json
      group: Bird
    bird-biel:
      url: https://mds.bird.co/gbfs/v2/public/biel/gbfs.json
      group: Bird

  # The following is optional, specifying a "group: Name" is enough to create a group.
  groups:
    Bird:
      name: Optionally override the name (otherwise it is called "Bird")
```

Provider selection UI:
Provider Selection UI

## Color Overrides

- Colors for both individual providers and groups can now be set in the configuration. The configuration values override colors given by the provider feed and can also be used to set colors for feeds that don't include color information.
- If no color is specified in the configuration, the `brand_assets.color` value from the `system_information` feed is used, if available.

Config example:
```yaml
gbfs:
  feeds:
    de-CallaBike:
      url: https://api.mobidata-bw.de/sharing/gbfs/v2/callabike/gbfs
      color: "#db0016"
    de-VRNnextbike:
      url: https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_vn/gbfs.json
      group: nextbike # uses the group color defined below
    de-NextbikeFrankfurt:
      url: https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_ff/gbfs.json
      group: nextbike
    de-KVV.nextbike:
      url: https://gbfs.nextbike.net/maps/gbfs/v2/nextbike_fg/gbfs.json
      group: nextbike
      color: "#c30937" # override color for this particular feed
  groups:
    nextbike:
      color: "#0046d6"
```

### Aggregated Feeds

For aggregated feeds (manifest.json or Lamassu), groups and colors can either be assigned to all providers listed in the aggregated feed or individually by using the `system_id`:

```yaml
gbfs:
  feeds:
    aggregated-single-group:
      url: https://example.com/one-provider-group/manifest.json
      group: Example
      color: "#db0016" # or assign a color to the group
    aggregated-multiple-groups:
      url: https://example.com/multiple-provider-groups/manifest.json
      group:
        source-nextbike-westbike: nextbike # "source-nextbike-westbike" is the system_id
        source-voi-muenster: VOI
        source-voi-duisburg-oberhausen: VOI
      # colors can be specified for individual feeds using the same syntax,
      # but in this example they are defined for the groups below
      #color:
      #  "source-nextbike-westbike": "#0046d6"
      #  "source-voi-muenster": "#f26961"
  groups:
    nextbike:
      color: "#0046d6"
    VOI:
      color: "#f26961"
```

## Map Updates
- Rental stations and vehicles on the map are now colored by provider.
- Geofencing zones (shown when a provider filter is active) no longer display overlapping zones with confusing colors. Instead, only the relevant zone is shown at each point.
	- Green zones: Parking allowed anywhere within the zone.
	- Yellow zones: Parking only allowed at stations.
	- Red zones: Driving through the zone is not allowed.
- Station and vehicle tooltips and popups (shown when clicking a station/vehicle icon) now show additional information about the vehicle type(s).
- Form factor "other" now has a question mark icon (previously it used the bike icon).
- Seated scooters now have a different icon than standing scooters.
- If debug mode is enabled (`?debug` URL parameter is present), detailed debug information is included in the popups which can be useful to debug issues with rental data.

Rental stations and vehicles in Frankfurt

Station tooltip

Vehicle tooltip

Geofencing Zones in Zürich

Geofencing Zones in Frankfurt

### DIFF
--- a/src/gbfs/osr_mapping.cc
+++ b/src/gbfs/osr_mapping.cc
@@ -126,12 +126,12 @@ struct osr_mapping {
               if (!applies(r.vehicle_type_idxs_, prod.vehicle_types_)) {
                 continue;
               }
-              start_allowed = r.ride_start_allowed_;
-              end_allowed = r.ride_end_allowed_;
-              through_allowed = r.ride_through_allowed_;
               if (r.station_parking_.has_value()) {
                 station_parking = r.station_parking_.value();
               }
+              start_allowed = r.ride_start_allowed_;
+              end_allowed = r.ride_end_allowed_ && !station_parking;
+              through_allowed = r.ride_through_allowed_;
               break;
             }
             if (start_allowed.has_value()) {
@@ -139,7 +139,7 @@ struct osr_mapping {
             }
           }
         }
-        if (end_allowed.has_value() && !station_parking) {
+        if (end_allowed.has_value()) {
           rd.end_allowed_.set(n, *end_allowed);
         }
         if (through_allowed.has_value()) {


ui/src/lib/AddressTypeahead.svelte
```diff
@@ -1,8 +1,7 @@
 <script lang="ts">
 	import { Combobox } from 'bits-ui';
 	import { geocode, type Match, type Mode } from './api/openapi';
-	import House from 'lucide-svelte/icons/map-pin-house';
-	import Place from 'lucide-svelte/icons/map-pin';
+	import { MapPinHouse as House, MapPin as Place } from '@lucide/svelte';
```
felixguendling, line 4: I read somewhere that this includes all icons which impacts bundle size. That's why I include them one-by-one. Not sure if that's still a problem?

ui/src/lib/Debug.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import Bug from 'lucide-svelte/icons/bug';
+	import { Bug, X, LoaderCircle } from '@lucide/svelte';
```
felixguendling, line 2: same

ui/src/lib/IsochronesInfo.svelte
```diff
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import { t } from '$lib/i18n/translation';
-	import LoaderCircle from 'lucide-svelte/icons/loader-circle';
+	import { LoaderCircle } from '@lucide/svelte';
```
felixguendling, line 3: same

ui/src/lib/IsochronesMask.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { t } from '$lib/i18n/translation';
 	import { Slider } from 'bits-ui';
-	import LocateFixed from 'lucide-svelte/icons/locate-fixed';
+	import { LocateFixed } from '@lucide/svelte';
```
felixguendling, line 4: same

ui/src/lib/AddressTypeahead.svelte
felixguendling: ok, lucide was updated - probably fixed there

PR #1104, opened 2025-10-22
Add supported real time protocols

MichaelKutzner:
This adds a brief documentation for the supported real time protocols.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -177,7 +177,7 @@ gbfs:
 
 # Real time protocols
 
-MOTIS supports multiple protocols for real time feeds. This sections shows a list of the protocols, including some pitfalls:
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -172,3 +174,13 @@ gbfs:
   update_interval: 300
   http_timeout: 10
 ```
+
+# Realtime protocols
+
+MOTIS supports multiple protocols for real time feeds. This sections shows a list of each protocol, including some pitfalls:
```
mority: nitpick: I would prefer the wording "This section shows a list of the protocols"

MichaelKutzner: That's fine for me. I've updated the text.

PR #1109, opened 2025-10-24
Show all found journeys on the map

schaerfo:
Dark theme: 
grafik

Light theme:
grafik

The patch still has one minor issue (that I know of): Multiple itineraries can be selected when clicking if they overlap on the map. Then, they have to be closed one-by-one.

### DIFF
--- a/ui/src/lib/ItineraryGeoJSON.svelte
+++ b/ui/src/lib/ItineraryGeoJSON.svelte
@@ -68,12 +68,14 @@
 		itinerary,
 		id,
 		selected,
+		selectItinerary,
 		level,
 		theme
 	}: {
 		itinerary: Itinerary;
 		id?: string;
 		selected: boolean;
+		selectItinerary?: () => void;
 		level: number;
 		theme: 'light' | 'dark';
 	} = $props();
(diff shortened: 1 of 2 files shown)


felixguendling:
> Multiple itineraries can be selected when clicking if they overlap on the map. Then, they have to be closed one-by-one.

I tried it and I'm not sure if this is a good user experience? Any ideas how to improve this?

Lach-anonym:
Well that's good for checking an itinerary going through a certain place instead of guessing by line and trips

schaerfo:
For reference, this is the behavior as of now:
[Bildschirmaufnahme_20251112_163547.webm](https://github.com/user-attachments/assets/a75cd382-96a0-452f-b84f-944ea3c66279)

But I am not sure if it is actually unwanted to see all itineraries that take a specific path, instead of just a single one (which one?).

If we don't want this, I can see two options:

1. Make the `onclick` hander register only for a single itinerary
2. Unselect a previously selected itinerary when clicking (going back in the browser history). Thus, we would cycle through all itineraries taking the selected path and get stuck on the last one.

I couldn't get either of these to work, so feel free to update the PR!

Of course, we could also remove the last commit from the PR and not merge the click-to-select feature as of now. But I (and some others as well it seems) do find it quite useful.

felixguendling:
Merged it. Maybe we can come up with a better alternative. In the previous UI we had a hover effect that would only highlight the connections in the list but not go into the detail view.

PR #1114, opened 2025-10-26
Update setup doc about JSON siri support

pialwis:
Following #1112, this PR updates the setup doc to include SIRI JSON support.

Also should we change in the code `siri` to `siri_xml` to be more explicit?


felixguendling:
Thanks! In general, you're right that this would make it more obvious. It would also break existing configurations. Since SIRI is primarily specified as XSD, I would say that SIRI by default implies XML. I'm not even sure where this JSON translation is standardized.

PR #1146, opened 2025-11-06
take dur into account for odm td_offsets

traines-source:
For a ride share offer 
Niesky, Deutschland -> Weißwasser Bahnhof, Weißwasser/O.L.
16:19 - 17:16 we obtain this:
image

(i.e. too early departure time and non-repaired bus stop time) – can be reproduced on many/all examples on staging.

[In other places](https://github.com/motis-project/motis/blob/2060bc7c52ca993c9f7d65fc26d8a133bccf75cb/src/flex/flex.cc#L350), the offset is reset to `kMaxDuration` only after the duration of the walk/ride and not after 1min. Not sure why, but this appears to fix the ride share legs again. Haven't yet completely fixed the tests because this might need more work and I wanted to discuss this first.


mority:
I hope that motis-project/nigiri#276 makes this PR obsolete.

Issue #1105, opened 2025-10-22
fetchStops - Could not find stopover location in trip


ton-An:
When using the experimental fetchStops feature (#1022) on the stop times endpoint, the server occasionally returns a 500 error with the message `Could not find stopover location in trip`.

With transitious, two station where it seems to happen often are München Hbf (de-DELFI_de:09162:100:1:1) and Mainz Hbf (fr-horaires-sncf_StopArea:OCE80190512). With the radius set to 200.

The error occurs at: https://github.com/motis-project/motis/blob/77f4dc234f5a30f3ca1ea12abf527de6ffc45b5d/src/endpoints/stop_times.cc#L321
I haven’t used C++ before, so sadly I don’t really understand the reason it fails.

felixguendling:
Do you have an example on how to reproduce this problem?

ton-An:
> Do you have an example on how to reproduce this problem?

```
https://api.transitous.org/api/v5/stoptimes?stopId=fr-horaires-sncf_StopArea:OCE80190512&n=100&fetchStops=true&radius=200&pageCursor=LATER%7C1761218040
```

Sorry for not including an example request!

felixguendling:
I hope this is fixed with #1153, if not please re-open.

ton-An:
Can confirm. The issue is fixed

PR #1153, opened 2025-11-09
Fix loop lines when querying next/previous stops on stoptimes

schaerfo:


felixguendling:
Does this also fix https://github.com/motis-project/motis/issues/1105?

schaerfo:
Yes, it should be fixed since we are now using the original stop index instead of searching for the location of the stop. However, I could not reproduce the issue so I could not actually verify.

PR #1161, opened 2025-11-11
Declare workspace for UI build

MichaelKutzner:
This explicitly declares the workspaces to use for the UI build.
I think this might be needed, as I encountered errors otherwise: `[vite]: Rollup failed to resolve import "@motis-project/motis-client" from "/root/ui/src/routes/+page.svelte".`
As building the API requires `pnpm`, I think this should be listed as a dependency as well. Not sure, if we should add this to `ui/api/package.json` or `ui/package.json` though. Any preferences?

### DIFF
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -193,6 +193,9 @@ importers:
       '@hey-api/openapi-ts':
         specifier: ^0.53.12
         version: 0.53.12(typescript@5.9.3)
+      pnpm:
+        specifier: ^10.0.0
+        version: 10.21.0
       tslib:
         specifier: ^2.8.1
         version: 2.8.1
@@ -2005,6 +2008,11 @@ packages:
   playwright@1.56.1:
     resolution: {integrity: sha512-aFi5B0WovBHTEvpM3DzXTUaeN6eN0qWnTkKx4NQaH4Wvcmc153PdaY2UBdSYKaGYw+UyWXSVyxDUg5DoPEttjw==}
     engines: {node: '>=18'}
+    hasBin: true
+
+  pnpm@10.21.0:
+    resolution: {integrity: sha512-2jM3Jn5AD909R5psaAeaxtsB2MpPZ1cgg+cid1p5Z4inqZVmE3SeAA+sINQktZT3p5Gl9OLhNYHF75R/JpaKQA==}
+    engines: {node: '>=18.12'}
     hasBin: true
 
   point-in-polygon-hao@1.2.4:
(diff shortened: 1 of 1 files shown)


pablohoch:
> As building the API requires `pnpm`, I think this should be listed as a dependency as well. Not sure, if we should add this to `ui/api/package.json` or `ui/package.json` though. Any preferences?

I don't think that is very useful because you would already need a pnpm installation to install the package because of this:

```json
"devEngines": {
		"packageManager": {
			"name": "pnpm",
			"onFail": "error"
		}
	}
```

At least recent npm versions should fail `npm install` with an error:

```
❯ npm install
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid devEngines.packageManager
npm error EBADDEVENGINES Invalid name "pnpm" does not match "npm" for "packageManager"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'npm', version: '11.6.2' },
npm error EBADDEVENGINES   required: { name: 'pnpm', onFail: 'error' }
npm error EBADDEVENGINES }
```

Previously we had an entry for corepack in the package.json, but that was removed in one of the recent Node.js versions. AFAIK pnpm just needs to be installed by the user somehow.

MichaelKutzner:
Thank you, I didn't noticed that section.
On the other hand, missing that dependency at all also feels a bit off. And having a minimal required version shouldn't cause any harm? So I need to think and test, if there is a better solution for this.

MichaelKutzner:
Closed by accident

felixguendling:
> AFAIK pnpm just needs to be installed by the user somehow.

I would rather put this notice somewhere in the README than to add it as a dependency into a file that requires you to already have `pnpm` installed to be able to install `pnpm` as a dependency.

MichaelKutzner:
For documentation:
I now tested the build with both Bun and Deno.

With Bun the `pnpm` configuration is read and the `workspace` entry is added to `package.json` automatically. Building the API client fails, unless `npnm` is installed in some way. This can be solved by adding `pnpm` as a dependency.
Build order:
```sh
cd api
bun run build
cd ..
bun run build
```

For Deno it is a bit more complicated. The workspace will not be found, unless explicitly added to `package.json`. It also requires `pnpm` and `node` to build. For `pnpm` a dependency might be used. But `node` has to be installed in the `PATH`.
Building:
```sh
deno task -r build
```

I didn't spend much time in testing this. And I'm aware, that we cannot support all engines. I just wanted to check, how these behave, so we can adopt in the future, if necessary.

I still think, that the `workspace` entry might be needed though. Also improving the documentation is a good idea.

felixguendling:
What would be the advantage of supporting other runtimes?

We just need a way to create the build artifacts - that's it.

MichaelKutzner:
There is currently none. But as I need to update my setup anyway and was curious about that, I ran some brief tests. And sharing these results might help someone in the future.

PR #1186, opened 2025-11-20
Reflect config in the UI

### LINKED ISSUE #1035: Reflect configured capabilities and limits of server in UI
Do not show/disable options in the UI that are not enabled in the config/setup, like:
* Routed transfers without osr_footpath
* avoid inclines without loaded elevation
* other osm related stuff without loaded OSM

Adapt selectable ranges in UI to configured limits:
* onetoall_max_travel_minutes (Isochrones)
* max preTransit/postTransit/directTime (TBD)

Proposition: Expose these limits/configured values in /map/initial or another endpoint, to be able to act on them in the UI.

DIDONAX:
fixes #1035
- Routed transfers without osr_footpath
- Avoid inclines without loaded elevation
- Adapt selectable ranges in UI to configured limits: ( onetoall_max_travel_minutes/ max preTransit/postTransit/directTime)

### DIFF
--- a/ui/src/routes/+page.svelte
+++ b/ui/src/routes/+page.svelte
@@ -565,35 +565,35 @@
 				<Tabs.Trigger value="isochrones">{t.isochrones.title}</Tabs.Trigger>
 			</Tabs.List>
 			<Tabs.Content value="connections">
-        <Card class="overflow-y-auto overflow-x-hidden bg-background rounded-lg">
-          <SearchMask
-            geocodingBiasPlace={center}
-            {serverConfig}
-            bind:from
-            bind:to
-            bind:time
-            bind:arriveBy
-            bind:useRoutedTransfers
-            bind:maxTransfers
-            bind:pedestrianProfile
-            bind:requireCarTransport
-            bind:requireBikeTransport
-            bind:transitModes
-            bind:preTransitModes
-            bind:postTransitModes
-            bind:directModes
-            bind:elevationCosts
-            bind:maxPreTransitTime
-            bind:maxPostTransitTime
-            bind:maxDirectTime
-            bind:ignorePreTransitRentalReturnConstraints
-            bind:ignorePostTransitRentalReturnConstraints
-            bind:ignoreDirectRentalReturnConstraints
-            bind:preTransitProviderGroups
-            bind:postTransitProviderGroups
-            bind:directProviderGroups
-          />
-        </Card>
+				<Card class="overflow-y-auto overflow-x-hidden bg-background rounded-lg">
+					<SearchMask
+						geocodingBiasPlace={center}
+						{serverConfig}
+						bind:from
+						bind:to
+						bind:time
+						bind:arriveBy
+						bind:useRoutedTransfers
+						bind:maxTransfers
+						bind:pedestrianProfile
+						bind:requireCarTransport
+						bind:requireBikeTransport
+						bind:transitModes
+						bind:preTransitModes
+						bind:postTransitModes
+						bind:directModes
+						bind:elevationCosts
+						bind:maxPreTransitTime
+						bind:maxPostTransitTime
+						bind:maxDirectTime
+						bind:ignorePreTransitRentalReturnConstraints
+						bind:ignorePostTransitRentalReturnConstraints
+						bind:ignoreDirectRentalReturnConstraints
+						bind:preTransitProviderGroups
+						bind:postTransitProviderGroups
+						bind:directProviderGroups
+					/>
+				</Card>
 			</Tabs.Content>
 			<Tabs.Content value="departures">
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -1725,6 +1725,10 @@ paths:
                   - lat
                   - lon
                   - zoom
+                  - routeFootPath
```
traines-source: `hasElevation` is missing (doesn't really matter, but for completeness)

openapi.yaml
```diff
@@ -1735,6 +1739,26 @@ paths:
                   zoom:
                     description: zoom level
                     type: number
+                  hasElevation: 
+                    description: true if elevation is loaded
+                    type: boolean
+                  routeFootPath:
```
traines-source: I think `routedTransfersAvailable` would be a better name, since this is really only for intermediate transfers, pre and post transit is always available when osm is loaded. For the config.yml field `street_routing` there should be an additional field e.g. `streetRoutingAvailable` that removes all settings after the maximum number of transfers in the advanced settings (first/last/direct/elevation). For consistency, then rename to `elevationAvailable` (or everything with has...)

Maybe also extend the description of the initial endpoint to better explain what these values are about.

ui/src/lib/AdvancedOptions.svelte
```diff
@@ -150,16 +131,18 @@
...
-				bind:checked={useRoutedTransfers}
-				label={t.useRoutedTransfers}
-				id="useRoutedTransfers"
-				onCheckedChange={(checked) => {
-					if (wheelchair && !checked) {
-						wheelchair = false;
-					}
-				}}
-			/>
+			{#if routeFootPath}
```
traines-source: This should also encompass wheelchair below (because this in turn is only selectable if routedTransfers are enabled). Alternative: disable instead of hide

ui/src/lib/AdvancedOptions.svelte
```diff
@@ -18,9 +18,14 @@
...
 	import TransitModeSelect from '$lib/TransitModeSelect.svelte';
 	import { type NumberSelectOption } from '$lib/NumberSelect.svelte';
+	import { generateTimes } from './generateTimes';
 
 	let {
 		useRoutedTransfers = $bindable(),
+		maxPrePostTransitTimeLimit,
+		maxDirectTimeLimit,
+		hasElevation,
+		routeFootPath,
```
traines-source: maybe put this into an object so that we don't have to pass around 4-5 values everywhere?

openapi.yaml
```diff
@@ -1735,6 +1739,26 @@ paths:
...
                     description: zoom level
                     type: number
+                  hasElevation: 
+                    description: true if elevation is loaded
+                    type: boolean
+                  routeFootPath: 
+                    description: enable routing footpaths
+                    type: boolean
+                  maxTravelTimeLimit: 
+                    description: limit for maxTravelTime API param
```
traines-source: specify that this applies to onetoall only (at least in the description, maybe also in the name of the field?)

felixguendling:
Eventually, the routed transfers will be available per profile (car, bike, wheelchair, baby buggy, etc etc).
https://github.com/motis-project/motis/pull/982

ui/src/routes/+page.svelte
```diff
@@ -217,17 +220,32 @@
 	let maxTransfers = $state<number>(
 		parseIntOr(urlParams?.get('maxTransfers'), defaultQuery.maxTransfers)
 	);
-	let maxTravelTime = $state<number>(
-		parseIntOr(urlParams?.get('maxTravelTime'), defaultQuery.maxTravelTime)
+	let maxTravelTime = $derived<number>(
+		parseIntOr(
+			urlParams?.get('maxTravelTime'),
+			Math.min(defaultQuery.maxTravelTime, serverConfig?.maxOneToAllTravelTimeLimit ?? Infinity)
```
traines-source: The UI maxTravelTime variable and defaultQuery seems to expect seconds, i.e. `serverConfig?.maxOneToAllTravelTimeLimit` needs to be multiplied by 60 (otherwise, the default is displayed as 2 minutes). Also, the `defaultQuery.maxTravelTime` is set to 45 minutes, which is not selectable anymore, so maybe set it to 30.

ui/src/lib/generateTimes.ts
```diff
@@ -0,0 +1,25 @@
+export const generateTimes = (limit: number): number[] => {
+	const times: number[] = [];
+	let t = 1;
+	const defaultLimit = 6 * 60 * 60;
```
traines-source: For maxDirectTime it would actually be useful to be able to go beyond 6 hours if the server is configured like that (e.g. car only journeys). Just for isochrones/maxOneToAllTravelTimeLimit I think the UI can't really handle displaying more than 4-6 hours (I think, maybe worth trying out).

PR #1196, opened 2025-11-26
change earlier trips color

felixguendling:
Screenshot from 2025-11-26 12-45-37
Screenshot from 2025-11-26 12-45-06


felixguendling:
update:

grafik
grafik

PR #1211, opened 2025-12-16
display error messages from the backend

jendrikw:
before:

image

after:

image

### DIFF
--- a/ui/src/lib/preprocessItinerary.ts
+++ b/ui/src/lib/preprocessItinerary.ts
@@ -1,6 +1,12 @@
-import type { Itinerary, Place, PlanResponse } from '@motis-project/motis-client';
+import type {
+	Itinerary,
+	Place,
+	PlanResponse,
+	Error as ApiError
+} from '@motis-project/motis-client';
 import type { Location } from '$lib/Location';
 import polyline from '@mapbox/polyline';
+import type { RequestResult } from '@hey-api/client-fetch';
 
 export const joinInterlinedLegs = (it: Itinerary) => {
 	const joinedLegs = [];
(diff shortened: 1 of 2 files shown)


ui/src/lib/preprocessItinerary.ts
```diff
@@ -41,12 +47,13 @@ export const preprocessItinerary = (from: Location, to: Location) => {
...
 	};
 
-	return (r: { data: PlanResponse | undefined; error: unknown }) => {
+	return (r: Awaited<RequestResult<PlanResponse, ApiError, false>>): PlanResponse => {
+		if (r.error?.error) throw new Error(r.error.error);
 		if (r.error) throw new Error(String(r.error));
 
-		r.data!.itineraries.forEach(updateItinerary);
-		r.data!.direct.forEach(updateItinerary);
+		r.data.itineraries.forEach(updateItinerary);
```
jendrikw, line 54: Due to the better typing of `r`, the non-null assertions can be removed. I noticed in this case, but in general it is something typescript-eslint can catch with type-aware linting, which might be useful to have, but I don't know how to set that up in .svelte files.

https://typescript-eslint.io/rules/no-unnecessary-type-assertion/

ui/src/lib/preprocessItinerary.ts
```diff
@@ -41,12 +47,13 @@ export const preprocessItinerary = (from: Location, to: Location) => {
 		joinInterlinedLegs(it);
 	};
 
-	return (r: { data: PlanResponse | undefined; error: unknown }) => {
+	return (r: Awaited<RequestResult<PlanResponse, ApiError, false>>): PlanResponse => {
```
jendrikw, line 50: If you know of a better way of writing this type please let me know.

felixguendling:
Thank you! :)

PR #1212, opened 2025-12-20
Add more advanced options to ui

### LINKED ISSUE #1015: Add all advanced options from API to UI
Add options available in the API https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/motis-project/motis/refs/heads/master/openapi.yaml#tag/routing/operation/plan to the UI.

* via, viaMinimumStay
* minTransferTime, additionalTransferTime, transferTimeFactor (one of them?)
* ...?

Should also be accessible/persisted in the URL (cf. existing params).
Should probably not be done before #1014, otherwise UI with advanced options becomes completely unusable on mobile.

DIDONAX:
fixes #1015 
added:
- transferTimeFactor
- additionalTransferTime
- cyclingSpeed
- pedestrianSpeed


ui/src/lib/AdvancedOptions.svelte
```diff
@@ -193,8 +216,15 @@
 						labelFormatter={formatDurationSec}
 					/>
 				{/if}
+				<div class="text-sm">{t.routingSegments.additionalTransferTime}</div>
```
traines-source, line 235: maybe add unit (min) in description or somewhere apart from placeholder, or make field initially empty so that placeholder is shown

ui/src/lib/AdvancedOptions.svelte
```diff
@@ -232,6 +262,20 @@
...
+					min={pedestrianProfile == 'FOOT' ? 0.8 : 0.5}
+					step={0.1}
+					max={pedestrianProfile == 'FOOT' ? 2 : 1.6}
+					bind:value={pedestrianSpeed}
+				/>
+			</div>
+			<div class="grid grid-cols-[1fr_2fr_1fr] items-center gap-2">
+				<span class="text-sm">{t.routingSegments.cyclingSpeed}</span>
+				<Slider min={2.7} max={7} step={0.1} disabled={!hasBikeMode} bind:value={cyclingSpeed} />
+			</div>
```
traines-source, line 294: in a common grid so that sliders are aligned?
description of extremes on both sides of the slider, slow-fast or even km/h if conversion into transferTimeFactor is reliable? – would also make the end of the slider more visible which for some reason is not visible

ui/src/lib/i18n/de.ts
```diff
@@ -109,7 +111,11 @@ const translations: Translations = {
 		direct: 'Direktverbindung',
 		maxPreTransitTime: 'Max. Vorlaufzeit',
 		maxPostTransitTime: 'Max. Nachlaufzeit',
-		maxDirectTime: 'Max. Direktzeit'
+		maxDirectTime: 'Max. Direktzeit',
+		additionalTransferTime: 'Zusätzliche Umstiegszeit',
+		pedestrianSpeed: 'Fußgängergeschwindigkeit',
+		cyclingSpeed: 'Fahrradtgeschwindigkeit',
```
traines-source:
```suggestion
		cyclingSpeed: 'Fahrradgeschwindigkeit',
```

ui/src/lib/i18n/fr.ts
```diff
@@ -108,7 +110,11 @@ const translations: Translations = {
 		direct: 'Connexion directe',
 		maxPreTransitTime: 'Durée max. avant transit',
 		maxPostTransitTime: 'Durée max. après transit',
-		maxDirectTime: 'Durée max. directe'
+		maxDirectTime: 'Durée max. directe',
+		additionalTransferTime: 'Durée de transfert supplémentaire',
```
traines-source:
```suggestion
		additionalTransferTime: 'Durée de correspondance supplémentaire',
```

ui/src/lib/AdvancedOptions.svelte
```diff
@@ -203,8 +226,15 @@
 						labelFormatter={formatDurationSec}
 					/>
 				{/if}
+				<div class="text-sm">{t.routingSegments.additionalTransferTime}</div>
+				<input
+					type="number"
+					min="0"
+					bind:value={additionalTransferTime}
+					placeholder="Duration (min)"
```
traines-source: i18n for placeholder

traines-source (commented):
image

I think grid alignment should be improved.
Also, I think that if we display the speeds, they should be converted to km/h, because at least I don't really have an intuitive understanding of m/s.
Also not sure why the bike speed slider does not distinguish itself visually if it's disabled, that can be confusing I guess...

DIDONAX:
maybe for cycling speed, but for pedestrian i find m/s more intuitive because i can approximate my speed on the fly assuming 1step ~ 1m, and same goes for wheelchair i am sure they have similar intuition. I would say for running/jogging km/h is better but the speed limit doesn't reach that unless i increase it, which i think i should.

traines-source:
I hadn't thought about jogging before, you're right that this could be a use case at least for direct itineraries. So yeah why not increase the limit. I would still favor km/h. E.g. 1.2 m/s sounds really fast to me but 4.32 km/h is actually quite slow (for myself I'm always calculating with 6km/h - 10min/km (which would be yet another unit that one could use :D). But yeah maybe not worth the effort with the stepping etc. so I'm open to leave it with m/s for the time being.

traines-source:
The transfer time issues now finally being fixed, we could merge this as soon as the conflicts are resolved.

PR #1214, opened 2025-12-21
via ui

felixguendling:
TODO review (vibe-coded)

### DIFF
--- a/ui/src/routes/+page.svelte
+++ b/ui/src/routes/+page.svelte
@@ -400,10 +400,10 @@
 	let stopNameFromResponse = $state<string>('');
 	$effect(() => {
 		if (baseQuery && activeTab == 'connections') {
-			const q = baseQuery.query;
 			clearTimeout(searchDebounceTimer);
 			searchDebounceTimer = setTimeout(() => {
 				const base = plan(baseQuery).then(preprocessItinerary(from, to));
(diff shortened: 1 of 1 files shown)


ui/src/lib/ViaStopOptions.svelte
```diff
@@ -0,0 +1,122 @@
...
+	}: {
+		via: undefined | Location[];
+		viaMinimumStay: undefined | number[];
+		viaLabels: Record<string, string>;
+	} = $props();
+
+	const possibleViaStayDurations = $derived([0, ...generateTimes(2 * 60 * 60)]);
+	const viaMinimumStayOptions = $derived(
+		possibleViaStayDurations.map((duration) => ({
+			value: duration.toString(),
```
Findus23, line 24: FYI: It seems like the number for the `viaMinimumStay` API argument should be in minutes, but this is passing the selected length in seconds causing the stays to be too long by a factor of 60

felixguendling: thank you! should be fixed with https://github.com/motis-project/motis/commit/20d4cb5c08f073b654dbed47282934164b7618bd

Findus23: Thanks!

PR #1215, opened 2025-12-22
Improve error handling

### LINKED ISSUE #1012: Return better HTTP error codes
Do not return HTTP 500 on client errors (rather 400 Bad Request) or when something is not found (HTTP 404) or... Necessary so that reverse proxies can correctly handle 500 to mean "something went catastrophically wrong, maybe try another instance?"

Searching for all utl::verify is probably a good start.

DIDONAX:
fixes #1012


openapi.yaml
```diff
@@ -847,12 +847,24 @@ paths:
               - TB
             default: PONG
       responses:
+        '404':
```
traines-source, line 856: Add 422

src/endpoints/adr/geocode.cc
```diff
@@ -16,6 +16,7 @@
 #include "motis/endpoints/adr/suggestions_to_response.h"
 #include "motis/parse_location.h"
 #include "motis/timetable/modes_to_clasz_mask.h"
+#include "net/bad_request_exception.h"
```
traines-source: please put it into a separate group (separated by newline) so that it appears above the nigiri/timetable.h import (groups go from generic to specific)

src/endpoints/adr/suggestions_to_response.cc
```diff
@@ -13,6 +13,7 @@
 #include "motis/journey_to_response.h"
 #include "motis/tag_lookup.h"
 #include "motis/timetable/clasz_to_mode.h"
+#include "utl/verify.h"
```
traines-source: idem

src/endpoints/one_to_all.cc
```diff
@@ -3,6 +3,8 @@
 #include <chrono>
 #include <vector>
 
+#include "net/bad_request_exception.h"
```
traines-source, line 8: idem

src/endpoints/stop_times.cc
```diff
@@ -3,6 +3,8 @@
 #include <algorithm>
 #include <memory>
 
+#include "net/bad_request_exception.h"
```
traines-source: idem

src/endpoints/stop_times.cc
```diff
@@ -155,9 +158,12 @@ struct rt_ev_iterator : public ev_iterator {
             !n::routing::is_allowed(
                 allowed_clasz, rtt.rt_transport_section_clasz_[rt_t].at(0)) ||
             (dir == n::direction::kForward ? time() < start : time() > start)} {
-    assert((ev_type == n::event_type::kDep &&
-            stop_idx_ < rtt_.rt_transport_location_seq_[rt_t].size() - 1U) ||
-           (ev_type == n::event_type::kArr && stop_idx_ > 0U));
+    utl::verify(
```
traines-source: if you transform asserts into verifys, we have to be a bit careful regarding performance and actually harmless assertions that then suddenly throw in production (since asserts are left out in release builds). But I hope this should be fine.

src/endpoints/map/stops.cc
```diff
@@ -16,15 +17,17 @@ api::stops_response stops::operator()(boost::urls::url_view const& url) const {
   auto const query = api::stops_params{url.params()};
   auto const min = parse_location(query.min_);
   auto const max = parse_location(query.max_);
-  utl::verify<openapi::bad_request_exception>(
-      min.has_value(), "min not a coordinate: {}", query.min_);
-  utl::verify<openapi::bad_request_exception>(
-      max.has_value(), "max not a coordinate: {}request_exception", query.max_);
+  utl::verify<net::bad_request_exception>(
+      min.has_value(), "Min not a coordinate: {}", query.min_);
```
felixguendling: please keep casing like it is everywhere

it's neither a sentence (with a full stop, etc.) and in this case the parameter is named `min`, so it should stay "min not a coordinate: {}"

src/endpoints/map/rental.cc
```diff
@@ -15,6 +15,7 @@
 #include "geo/box.h"
 #include "geo/polyline.h"
 #include "geo/polyline_format.h"
+#include "utl/verify.h"
```
felixguendling: header grouping

src/endpoints/map/flex.cc
```diff
@@ -9,6 +9,7 @@
 #include "motis-api/motis-api.h"
 #include "motis/parse_location.h"
 #include "motis/tag_lookup.h"
+#include "net/bad_request_exception.h"
```
felixguendling: header grouping

src/endpoints/elevators.cc
```diff
@@ -1,5 +1,6 @@
 #include "motis/endpoints/elevators.h"
 
+#include "net/too_many_exception.h"
```
felixguendling, line 3: header grouping

src/compute_footpaths.cc
```diff
@@ -177,7 +178,11 @@ elevator_footpath_map_t compute_footpaths(
                 [](auto&& a, auto&& b) { return a.target() < b.target(); },
                 [](auto&& a, auto&& b) { return a.target() == b.target(); },
                 utl::overloaded{
-                    [](n::footpath, n::footpath) { assert(false); },
+                    [](n::footpath, n::footpath) {
+                      utl::verify(
```
felixguendling: no need

src/compute_footpaths.cc
```diff
@@ -208,7 +213,9 @@ elevator_footpath_map_t compute_footpaths(
     for (auto const [i, out] : utl::enumerate(transfers)) {
       auto const l = n::location_idx_t{i};
       for (auto const fp : out) {
-        assert(fp.target() < tt.n_locations());
+        utl::verify(fp.target() < tt.n_locations(),
```
felixguendling: no need

felixguendling:
Please keep this PR targeted on producing the right error codes.
- asserts stay asserts
- error messages stay the same

PR #1216, opened 2025-12-24
mem

felixguendling:


felixguendling:
This PR removes `http_client` code because it was somehow leaking/accumulating memory somewhere in libressl. GBFS works now based on `http_req.h` functionality.

PR #1226, opened 2026-01-10
Add doc about supported SIRI services and the expected SIRI JSON structure

pialwis:
This PR adds the following in the setup doc:

- which SIRI services are supported by MOTIS (right now only SIRI ET)
- the expected SIRI JSON response structure. Lately I was playing with IDFM (Paris) SIRI ET feeds and realized they don't work out of the box in MOTIS, the server throwing `[VERIFY FAIL] unable to parse time ""` errors. The problem was the two top keys of the IDFM JSON response being `{"Siri": "ServiceDelivery": {...}}` with nigiri not expecting these keys for the JSON to XML conversion. So I thought documenting this would be helpful to others. Note that I don't really know if this is a specific IDFM issue or if it occurs for other SIRI JSON providers.


felixguendling:
Thank you! :)

PR #1236, opened 2026-01-18
UI performance optimization

### LINKED ISSUE #1016: Improve resource consumption of UI
It's quite a resource hog (probably mostly Railviz/the map).

DIDONAX:
fixes #1016
- Refactored RailViz mostly following  https://deck.gl/docs/developer-guide/performance
- Reduced bundle size by ~50% by eliminating duplicate packages (mostly maplibre-gl). 
- Faster initial render by inlining css.

### DIFF
--- a/ui/src/app.html
+++ b/ui/src/app.html
@@ -4,12 +4,6 @@
 		<meta charset="utf-8" />
 		<title>MOTIS</title>
 		<link rel="icon" href="%sveltekit.assets%/favicon.ico" />
-		<link
-			rel="preload"
-			href="https://unpkg.com/maplibre-gl@^5.16.0/dist/maplibre-gl.css"
-			as="style"
-			onload="this.onload=null;this.rel='stylesheet'"
-		/>
 		<meta name="viewport" content="width=device-width, initial-scale=1" />
 		%sveltekit.head%
 	</head>
--- a/ui/src/lib/RailViz.svelte
+++ b/ui/src/lib/RailViz.svelte
@@ -9,6 +9,7 @@
 	import { onClickTrip } from './utils';
 	import { getDelayColor, rgbToHex } from './Color';
 	import type { MetaData } from './types';
+	import Control from './map/Control.svelte';
 	let {
 		map,
 		bounds,
@@ -24,6 +25,7 @@
 	//QUERY
 	let startTime = $state(new Date(Date.now()));
 	let endTime = $derived(new Date(startTime.getTime() + 180000));
+	let canceled = $derived(colorMode === 'none');
 	let query = $derived.by(() => {
 		if (!bounds || !zoom) return null;
 		const b = maplibregl.LngLatBounds.convert(bounds);
@@ -108,6 +110,7 @@
 		}
 	};
 	const createLayer = () => {
+		if (!DATA.positions || DATA.positions.byteLength === 0) return;
 		return new IconLayer({
 			id: 'trips-layer',
 			data: {
@@ -134,7 +137,7 @@
 	};
 	let animationId: number;
 	const animate = () => {
-		if (DATA.positions.byteLength === 0) return;
+		if (!DATA.positions || DATA.positions.length === 0) return;
 		worker.postMessage(
 			{
 				type: 'update',
@@ -150,9 +153,8 @@
 
 	// UPDATE
 	$effect(() => {
-		if (!query || isProcessing) return;
+		if (!query || isProcessing || canceled) return;
 		untrack(() => {
-			if (colorMode === 'none') return;
 			isProcessing = true;
 			worker.postMessage({ type: 'fetch', query });
 		});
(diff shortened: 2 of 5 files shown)


ui/src/app.html
```diff
@@ -4,6 +4,12 @@
 		<meta charset="utf-8" />
 		<title>MOTIS</title>
 		<link rel="icon" href="%sveltekit.assets%/favicon.ico" />
+		<link
+			rel="preload"
+			href="https://unpkg.com/maplibre-gl@^5.16.0/dist/maplibre-gl.css"
```
traines-source: Due to privacy reasons, we should not load resources from external sources.

ui/src/lib/RailViz.svelte
```diff
@@ -26,329 +21,202 @@
...
+	*/
 	$effect(() => {
-		if (overlay && bounds && zoom && colorMode) {
-			untrack(() => {
-				console.debug(`updateRailviz: effect ${overlay} ${bounds} ${zoom} ${colorMode}`);
-				updateRailviz();
+		if (!map || !overlay) return;
+		map.addControl(overlay);
+		/*
+		untrack(() => {
```
traines-source: is this not needed anymore?

ui/src/lib/tripsWorker.ts
```diff
@@ -0,0 +1,203 @@
...
+import { client, trips, type TripSegment } from '@motis-project/motis-client';
+import type { Trip, Position, TransferData, MetaData, Query } from './types';
+import type { QuerySerializerOptions } from '@hey-api/client-fetch';
+import { getDelayColor, hexToRgb } from './Color';
+import { getModeStyle, getColor } from './modeStyle';
+
+// MATH
+const toRad = (deg: number) => (deg * Math.PI) / 180;
+const toDeg = (rad: number) => (rad * 180) / Math.PI;
+const getSpatialData = (p0: Position, p1: Position) => {
```
traines-source, line 12: I gather you have replaced the turf lib with this to reduce dependencies or because it's faster? If this is a bottleneck, further considerable speedups may be possible by using more approximate formulas (see e.g. https://github.com/motis-project/geo/blob/1450091a6ee2d541e91d26bf06157342ec0599f6/src/latlng.cc#L19-L40 which may still be precise enough despite the possible large distances since we only use the distances in a relative manner? – but even if, maybe let's merge this PR first and keep this for later)

ui/src/routes/+page.svelte
```diff
@@ -75,7 +75,7 @@
 					: 'connections')
 	);
 	let dataAttributionLink: string | undefined = $state(undefined);
-	let colorMode = $state<'rt' | 'route' | 'mode' | 'none'>('none');
+	let colorMode = $state<'rt' | 'route' | 'mode' | 'none'>('route');
```
traines-source: I think it was intentional that we now by default show nothing (especially on mobile). Also, when switching from none there is no initial /trips request made and nothing is shown before panning/zooming the map.

ui/src/lib/RailViz.svelte
```diff
@@ -26,329 +21,202 @@
...
-			map.removeControl(overlay);
-		}
+		if (animationId) cancelAnimationFrame(animationId);
+		if (overlay) map?.removeControl(overlay);
+		worker.terminate();
+		popup.remove();
 	});
 </script>
-
-{#if railvizError}
```
traines-source, line 350: I think it would still be good to show the error, or is there a particular reason we can't anymore?

ui/src/app.html
DIDONAX: But the resources are being loaded externally anyway and block the initial render by about 700ms, if I remember correctly. This change just makes them discovered earlier. Or am missing something ?

ui/src/lib/RailViz.svelte
DIDONAX: yes forgot to take the comments out

ui/src/routes/+page.svelte
DIDONAX: I did this for testing, to compare the current status quo with the improved version in parallel under the same initial conditions. I agree it should be off by default .

ui/src/lib/tripsWorker.ts
DIDONAX: It’s slightly faster because it calculates and allocates values common to both bearing and distance once and reuses them for both, instead of doing this separately. This is no longer the bottleneck, I would say the allocations caused by the worker copying metadata are what’s limiting peak performance.

ui/src/app.html
traines-source: I think before nothing was loaded from elsewhere, but if so, that should not be the case. But preloading should also be possible for relative resources?

felixguendling: > But the resources are being loaded externally anyway

The UI has to work offline without any internet connection. Just the compiled UI + MOTIS.

PR #1245, opened 2026-01-23
meta_router: reduce search interval to 2 h

mority:
please merge after motis-project/prima#570


felixguendling:
IMO only makes sense together with `PONG`

PR #1251, opened 2026-01-24
Platform code matching

### LINKED ISSUE #1238: matching platform often matches on wrong level
Even though Motis works really well with Vienna public transport, it struggles a lot with routing changes within the Karlsplatz metro station (the largest metro station in Vienna) going as far as doing [loops around the station](https://api.transitous.org/?time=2026-01-17T16%3A19%3A00.000Z&fromPlace=at-PTA-Eastern-Region-Flex-2026_Pat%3A49%3A560&toPlace=at-PTA-Eastern-Region-Flex-2026_Pat%3A49%3A949&withFares=true&fastestDirectFactor=1.5&pedestrianProfile=WHEELCHAIR&joinInterlinedLegs=false&useRoutedTransfers=true&maxMatchingDistance=8&fromName=Wien+H%C3%BCtteldorf%2C+Wien&toName=Wien+Oberlaa%2C+Wien) with "Accessible transfers". 

Which seems odd, as on OSM the station seems to be really well mapped indoor. 

I originally blamed https://github.com/motis-project/osr/issues/76 which causes the U1 platform on level -5 to instead appear on -4 and be visually broken in the UI. But with that fixed and pedestrian routing inside the station working without issues to all platforms, the routing involving public transport there did not improve. 

Checking [the stops](https://api.transitous.org/api/v1/map/stops?min=48.20237485958768%2C16.36641554862473&max=48.19953705671725%2C16.37164437407168) in Motis again, I realized that many of them have the wrong level.

For reference these are the levels in GTFS:
```
"at:49:654:0:1","Level 0"
"at:49:654:0:2","Level 0"
"at:49:655:0:1","Level 0"
"at:49:655:0:2","Level 0"
"at:49:657:0:10","Level -3"
"at:49:657:0:11","Level 0"
"at:49:657:0:12","Level 0"
"at:49:657:0:13","Level -3"
"at:49:657:0:14","Level -2"
"at:49:657:0:15","Level -5"
"at:49:657:0:16","Level 0"
"at:49:657:0:17","Level 0"
"at:49:657:0:2","Level 0"
"at:49:657:0:6","Level 0"
"at:49:657:0:7","Level 0"
"at:49:657:0:8","Level -5"
"at:49:657:0:9","Level -2"
```
(they are correct and match the OSM levels)

This causes e.g. a bus station on ground level (`at:49:657:0:2`) to match to the level -5 U1 platform ([OSM](https://www.openstreetmap.org/way/412071636)) below it ([this is the bus stop in OSM](https://www.openstreetmap.org/node/3584027179)): [example route using the 59A bus](https://api.transitous.org/?time=2026-01-18T18%3A15%3A00.000Z&fromPlace=at-PTA-Eastern-Region-Flex-2026_at%3A49%3A104%3A0%3A1&toPlace=48.20171220756839%2C16.369141588868672&withFares=true&fastestDirectFactor=1.5&joinInterlinedLegs=false&maxMatchingDistance=250&fromName=Wien+Bacherplatz%2C+Wien&toName=48.20171220756839%2C16.369141588868672%2C0).

Or the U1 platform (`at:49:657:0:8`) that should be on level -5 match the level -3 U2 platform ([OSM](https://www.openstreetmap.org/way/412014903)) right next to it [as seen here](https://api.transitous.org/?time=2026-01-18T18%3A15%3A00.000Z&fromPlace=48.20098482461162%2C16.369454736408215%2C-3&toPlace=at-PTA-Eastern-Region-Flex-2026_Pat%3A49%3A1366&withFares=true&fastestDirectFactor=1.5&joinInterlinedLegs=false&useRoutedTransfers=true&maxMatchingDistance=250&toName=Wien+Taubstummengasse%2C+Wien).

I am not completely sure why this happens, but this is my guess: On a first glance the code finding the best match seems to consider the level and give it a bonus. 
https://github.com/motis-project/motis/blob/0e1405816869bccb4da8d16876c7b26dbbb6689a/src/match_platforms.cc#L214-L235
But it does not check if the GTFS stop and the OSM platform have the same level, just that the OSM platform has *a* level.

I am not sure what the best solution would be here as there is not much information to match it on here. But I guess storing the GTFS stop level and comparing that to OSM might work (but then again while this is the case here, there is no reason why they would be equivalent).

Findus23:
fixes #1238 

"fixes" might be a too strong word, as this replaces one vague heuristic with a slightly different one.
But in my tests in Vienna and Lower Austria this improves matching of platforms by quite a bit.

The `has_number_match(names, *get_track(ref))` score bonus felt too strong to me. In my case it seemed to be mostly looking at the end of IFOPT codes like `at:43:3952:0:2` and then applying a lot of weight to the nearby train platform `2` even though it explicitly has a `platform_code` of `E`. It feels far too easy for me for this bonus to match with random numbers (this also happens a bit with the `routes_bonus` if e.g. the city bus line `2` is also involved).

On top of this, I added a new bonus that looks for exact matches between the GTFS platform_code and the OSM platform names (so things like ref, local_ref and name). This works really well in most cases, but of course still has flaws. Especially in multi-layer metro stations (where there are many platform 1/2) and in simple train stations with bus stops next to it. But I think it should not cause completely wrong/confusing matches. And at least in complicated metro stations any heuristic will fail and there is likely an unambiguous `ref:IFOPT` that can be added in OSM instead. 

That said, this is impossible to test thoroughly and I just checked different areas inside Vienna and around Lower Austria.

### DIFF
--- a/src/endpoints/matches.cc
+++ b/src/endpoints/matches.cc
@@ -54,7 +54,6 @@ json::value matches::operator()(json::value const& query) const {
                                }},
                osr::to_ref(pl_.platform_ref_[p][0]));
 
-
     matches.emplace_back(json::value{
         {"type", "Feature"},
         {"properties", platform_data},
--- a/src/match_platforms.cc
+++ b/src/match_platforms.cc
@@ -136,7 +136,6 @@ int compare_platform_code(Collection&& names, std::string_view platform_code) {
   return bonus;
 }
 
-
 struct center {
   template <typename T>
   void add(T const& polyline) {
@@ -242,8 +241,7 @@ osr::platform_idx_t get_match(n::timetable const& tt,
     auto const routes_bonus = get_routes_bonus(tt, l, pl.platform_names_[x]);
     auto const code_bonus = compare_platform_code(
         pl.platform_names_[x],
-        tt.get_default_translation(tt.locations_.platform_codes_[l])
-        );
+        tt.get_default_translation(tt.locations_.platform_codes_[l]));
 
     auto const score =
         dist - match_bonus - way_bonus - lvl_bonus - routes_bonus - code_bonus;


felixguendling:
Thank you! :)

PR #1253, opened 2026-01-26
Expose bike transport in API

jbruechert:


felixguendling:
GTFS docs:
0 or empty - No bike information for the trip.
1 - Vehicle being used on this particular trip can accommodate at least one bicycle.
2 - No bicycles are allowed on this trip.

In the future we will probably want to omit the field in the response if the column is either missing or set to 0/empty.

PR #1254, opened 2026-01-28
fix: Czech locale is properly named cs not cz.

severak:
This was causing autodetection of locale not working.


felixguendling:
Nice :)

PR #1257, opened 2026-01-29
Support tripFrom in StopTime and Leg

### LINKED ISSUE #1247: Support `tripFrom` in `stoptimes` endpoint
The `StopTime` objects in a response from a `/api/v5/stoptimes` request include a `tripTo` `Place` attribute which is the final stop of the trip. It would make sense to also add a `tripFrom` attribute which would be the first stop of the trip.

Current workarounds are either:
- Make a new `/api/v5/trip` request with the trip id from the `StopTime` object, and retrieve the `from` `Place` attribute;
- Use the experimental `fetchStops` query parameter for the `/api/v5/stoptimes` request, and get the first stop from the `previousStops` attribute (first element of the list). The issue with that method is that if the first stop is the queried stop for the `/api/v5/stoptimes` request, then `previousStops` is empty, so you need to take that case into account.

It would be simpler to just have a `tripFrom` attribute tbh. Looking at the code, maybe implement in nigiri `get_first_trip_stop` the same way [`get_last_trip_stop`](https://github.com/motis-project/nigiri/blob/master/src/rt/frun.cc#L126) is used to [compute `tripTo`](https://github.com/motis-project/motis/blob/master/src/endpoints/stop_times.cc#L473)?

pialwis:
This PR aims to fix #1247. I also added a `tripFrom` attribute for Leg objects.

### DIFF
--- a/test/endpoints/stop_times_test.cc
+++ b/test/endpoints/stop_times_test.cc
@@ -147,6 +147,8 @@ TEST(motis, stop_times) {
     auto const& ice = res.stopTimes_[0];
     EXPECT_EQ(api::ModeEnum::HIGHSPEED_RAIL, ice.mode_);
     EXPECT_EQ("20190501_00:35_test_ICE", ice.tripId_);
+    EXPECT_EQ("test_DA_10", ice.tripFrom_.stopId_);
+    EXPECT_EQ("test_FFM_12", ice.tripTo_.stopId_);
     EXPECT_EQ("ICE", ice.displayName_);
     EXPECT_EQ("FFM Hbf", ice.headsign_);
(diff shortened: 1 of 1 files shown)


felixguendling:
CI is red

> C++ exception with description "bucket::at: index out of range" thrown in the test body.

pialwis:
That's greener :-). Before you merge, I'm thinking it would be better to add `tripFrom` and `tripTo` tests to `test/endpoints/stop_times_test.cc`

felixguendling:
Sounds reasonable. Will you add some checks for stop_times_test.cc? :)

pialwis:
Yes working on it on that PR

felixguendling:
Thank you :)

PR #1270, opened 2026-02-05
Add option to compute missing shapes

pablohoch:
Adds an option to compute missing shapes using OSM data.

Config example:
```yaml
timetable:
  # route_shapes and all its options are optional
  route_shapes:
    missing_shapes: true # default = false (no shapes are computed)
    clasz: # default: include all classes
      COACH: false # disable for class COACH (currently recommended)
    #max_stops: 100 # disable for routes with >100 stops
    #n_threads: 4 # can be used to limit the number of threads for shapes computation
    #cache: true # reuse routed shapes from previous import
    debug: # can be used to write debug data for specific routes (see osr PR for more info)
      path: /path/to/debug/folder
      # the following options determine for which routes debug files are created
      # all these options are optional and off by default
      all: false
      all_with_beelines: true # write debug info for all routes that contain at least one beelined segment
      slow: 10000 # write debug info for all routes that take >10.000ms to route
      trips:
        - "19.T3.24-52-E-j26-1.1.R"
        - "45.T0.32-102-C-j26-1.2.H"
      route_ids:
        - "22-331-E-j26-4"
        - "22-332-E-j26-4"
      route_indices:
        - 907
        - 908
```

The UI also contains a routes layer in debug mode (`?debug` in the URL).

Depends on:
- osr:
	- https://github.com/motis-project/osr/pull/78
	- https://github.com/motis-project/osr/pull/83
- nigiri:
	- https://github.com/motis-project/nigiri/pull/309
	- https://github.com/motis-project/nigiri/pull/318
	- https://github.com/motis-project/nigiri/pull/321

Example routes with calculated shapes in Hamburg


include/motis/route_shapes.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <optional>
+
+#include "nigiri/types.h"
+#include "cista/containers/hash_map.h"
+#include "geo/box.h"
+#include "osr/routing/profile.h"
```
felixguendling: new lines

docs/setup.md
```diff
@@ -339,3 +339,70 @@ following JSON structure:
...
+  # with_shapes must be set to true to enable shapes support, otherwise no shapes will be loaded or computed
+  with_shapes: true
+  route_shapes: # all these options are optional
+    # enable this to compute shapes for routes that don't have shapes in the dataset (default = false)
+    missing_shapes: true
+    # if replace_shapes is enabled, all shapes will be recomputed based on OSM data, even if shapes are already present in the dataset (default = false)
+    replace_shapes: true
+    # routing for specific clasz types can be disabled (default = all enabled)
+    # currently long distance street routing is slow, so in this example
+    # we disable routing shapes for COACH
```
felixguendling, line 368: from the documentation it's unclear whether disabled clasz shapes will be deleted or kept?

docs/setup.md
```diff
@@ -339,3 +339,70 @@ following JSON structure:
...
+- `street_routing` set to `true`
+- `timetable.route_shapes` config:
+
+```yaml
+timetable:
+  # with_shapes must be set to true to enable shapes support, otherwise no shapes will be loaded or computed
+  with_shapes: true
+  route_shapes: # all these options are optional
+    # enable this to compute shapes for routes that don't have shapes in the dataset (default = false)
+    missing_shapes: true
```
felixguendling, line 363: not sure if I understand this correctly, but a user who sets `timetable.route_shapes` always wants to route missing shapes, right? otherwise I would just not add `timetable.route_shapes` to the config?

docs/setup.md
pablohoch: Yes, that could be improved in the documentation. If there are shapes for these clasz in the timetable, those are kept. This applies to all options (e.g. also `max_stops`).

docs/setup.md
pablohoch: True, we could remove `missing_shapes` and enable that whenever `route_shapes` is present.
(I previously had `missing_shapes` and `route_shapes` directly under `timetable`, but now that they are grouped under `route_shapes` it's probably better to remove the `missing_shapes` option. Unless we later want to add some other modes e.g. to compute only some shapes, like an opt-in mode rather than the current opt-out mode.)

pablohoch: Another option could be to replace both `missing_shapes` and `replace_shapes` by something like `mode: missing | all`.

felixguendling:
I started with clasz COACH excluded and it went through for Transitous data but now I wanted to have COACH active again but the import didn't trigger. Is this expected/intended?

felixguendling:
Would also be nice to have some timings in the logs :)

pablohoch:
> I started with clasz COACH excluded and it went through for Transitous data but now I wanted to have COACH active again but the import didn't trigger. Is this expected/intended?

Not intended, I forgot to include the clasz mapping in the hash.

felixguendling:
maybe we need more matching distance?

grafik

https://test.motis-project.org/?tripId=20260222_22%3A48_de-DELFI_DE%3A%3AServiceJourney%3A3103194493_5%3A%3A&debug=

another example:
https://test.motis-project.org/?tripId=20260223_09%3A24_de-DELFI_DE%3A%3AServiceJourney%3A3103319714_1%3A%3A

felixguendling:
Probably we need a way to at one point have a penalty for sharp turns.

grafik

https://test.motis-project.org/?tripId=20260222_23%3A56_de-DELFI_DE%3A%3AServiceJourney%3A3102033412_0%3A%3A&debug=

vikiawv:
Hi I have also noticed the computing missing shapes does also overwrite existing shapes like this bus route.
https://api.transitous.org/?tripId=20260315_14%3A09_at-PTA-Eastern-Region-Flex-2026_143.T3.23-14A-j26-1.2.R
image

https://test.motis-project.org/?tripId=20260315_13%3A59_at-PTA-Eastern-Region-Flex-2026_142.T3.23-14A-j26-1.2.R&debug=
image

The wrong new shape is caused by this turn restriction, which is not valid for busses. (not sure if this is a OSM fault or not)
https://www.openstreetmap.org/relation/12244727/history

vikiawv:
Also from what I have seen the current algorithm doesn't penaltise railway sidings as such, and instead treats them if it was a shorter path: 
https://test.motis-project.org/?tripId=20260315_14%3A00_at-Railway-Current-Reference-Data-2026_22.TA.1-S50-W-j26-1.3.R&debug=
image

This is the rail in the example, tagged correctly as: `service=yard`
https://www.openstreetmap.org/way/1412191618

felixguendling:
> overwrite existing shapes like this bus route

We currently have enabled this, so we:
- find more bugs in the algorithm (not the case here, because the restriction is missing `except=psv`)
- know that everything you see is our mistake (so we don't have to first check the source of the geometry)

For production, we can use the feature that existing shapes won't be overwritten but testing it has been helpful to have only our own shape computation for now.

felixguendling:
> This is the rail in the example, tagged correctly as: service=yard
> https://www.openstreetmap.org/way/1412191618

Yes, we should definitely heavily penalize `service=yard` and other parts of the railway network that are not used for regular operations with passengers.

PR #1272, opened 2026-02-07
Reorganise transit modes in OpenAPI doc

pialwis:
This PR improves the OpenAPI doc about the transit modes:

- Remove `CABLE_CAR` in the `TRANSIT` description as it is deprecated and as it falls back to `FUNICULAR`. Rename `AREAL_LIFT` to `AERIAL_LIFT` (the former is deprecated). Finally reorganize the list to be more coherent with the mode list.
- In the enum list, put `CABLE_CAR` at the bottom to be more consistent with the description (all deprecated modes at the bottom).

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2391,7 +2391,7 @@ components:
 
         # Transit modes
 
-          - `TRANSIT`: translates to `RAIL,TRAM,BUS,FERRY,AIRPLANE,COACH,CABLE_CAR,FUNICULAR,AREAL_LIFT,OTHER`
+          - `TRANSIT`: translates to `TRAM,FERRY,AIRPLANE,BUS,COACH,RAIL,FUNICULAR,AERIAL_LIFT,OTHER`
           - `TRAM`: trams
           - `SUBWAY`: subway trains (Paris Metro, London Underground, but also NYC Subway, Hamburger Hochbahn, and other non-underground services)
           - `FERRY`: ferries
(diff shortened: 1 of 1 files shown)


piotrski:
Should we also consider adding support for `TROLLEY`?

felixguendling:
> Should we also consider adding support for `TROLLEY`?

You can get this information already from `route_type`.

piotrski:
> > Should we also consider adding support for `TROLLEY`?
> 
> You can get this information already from `route_type`.

However, `route_type` isn’t available on `Place`

felixguendling:
Then I would rather add `route_type` to `Place` than add `TROLLEY` to `Mode`.
`Mode` is mainly to filter search and I don't really see a reason to filter for `TROLLEY` specifically. It's not that this is technically not possible - it's only that once we add every `route_type` to `Mode`, there's no need to have `Mode` at all. Currently, 16bit are enough for route type filtering. Adding another (without removing one) will double the data type.

Lach-anonym:
Maybe in the future there could be an option to merge/divide the transport types for each instance? Maybe someone will want to launch a MOTIS instance and (for example) ignore trolley buses... if we contain such infos then why not let it be flexible?

felixguendling:
The names are hardcoded in the openapi.yml and can't easily be changed with a configuration file.

But I think it's possible to recompile MOTIS with a changed `openapi.yml` and adapted `route_type` -> clasz mapping if necessary: https://github.com/motis-project/nigiri/blob/b100035a44f4821f923bc8c593e858dda19360f9/src/loader/gtfs/route.cc#L21

mlundblad:
Also, _if_ we'd add support for trolley busses as a transport mode, I'd suggest TROLLEY_BUS (as trolley could possibly be confused with trams, I think they are referred to as "trolleys" in some dialects of English, like in San Francisco).

PR #1273, opened 2026-02-08
Add ODM mode in TRANSIT description in OpenAPI doc

pialwis:
The `ODM` transit mode was added in MOTIS a couple months ago. But it was not included in the `TRANSIT` translation description in the OpenAPI doc. This PR fixes that.

Also reorganise the transit mode description to be more coherent with the order of the mode list in `src/timetable/modes_to_clasz_mask.cc`.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2391,7 +2391,7 @@ components:
 
         # Transit modes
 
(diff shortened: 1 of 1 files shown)


Findus23:
Semi-related (and sorry if it has been fixed since), but ODM also doesn't show up in the (English) translations:
https://github.com/motis-project/motis/blob/2c8e946fa2349273c482768d40ef8d695d7e866d/ui/src/lib/i18n/en.ts#L94-L117

And therefore is shown as an empty entry in the UI.

PR #1278, opened 2026-02-10
tools: make try-reproduce.py portable across environments

escapables:
`tools/try-reproduce.py` currently hard-codes a machine-specific `motis` path, so it fails by default on other systems.

This change makes executable resolution portable:

1. `MOTIS_EXEC` env var (if set)
2. Repo-local builds: `build/motis`, `cmake-build-relwithdebinfo/motis`
3. `motis` from `PATH`

This keeps existing workflows intact while removing local path coupling.

created by @escapables, written by Codex

### DIFF
--- a/tools/try-reproduce.py
+++ b/tools/try-reproduce.py
@@ -8,14 +8,35 @@ import yaml
 from multiprocessing import Pool
 from pathlib import Path
 
+
+def resolve_motis_exec():
+    env_exec = os.environ.get("MOTIS_EXEC")
+    if env_exec:
+        return env_exec
+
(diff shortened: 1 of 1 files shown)


felixguendling:
This script is more like an example script that's supposed to be adapted to the use case.

escapables:
I see. All good! 👍

felixguendling:
More info: in some cases, this script should run with different MOTIS versions, in some cases with the same MOTIS version but different datasets or algorithm settings. So there's no one-size-fits-all solution and the easiest is to adapt the script directly once you need it.

PR #1280, opened 2026-02-10
Add one-to-many POST request in OpenAPI doc

pialwis:
#1164 introduced a POST endpoint for `/api/v1/one-to-many`. That PR updated the OpenAPI doc to specify the expected body, but forgot to document the POST request itself.

This PR completes the OpenAPI doc by adding the missing POST request, and also adds in the code a `Post` suffix to the relevant one-to-many POST code so to comply with OpenAPI specs (unique `operationId`).

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -932,7 +932,7 @@ paths:
       summary: |
         Street routing from one to many places or many to one.
         The order in the response array corresponds to the order of coordinates of the \`many\` parameter in the query.
-      operationId: oneToManyGet
+      operationId: oneToMany
       parameters:
         - name: one
           in: query
--- a/include/motis/endpoints/one_to_many.h
+++ b/include/motis/endpoints/one_to_many.h
@@ -15,7 +15,7 @@
 namespace motis::ep {
 
 template <typename Params>
-api::oneToManyGet_response one_to_many_handle_request(
+api::oneToMany_response one_to_many_handle_request(
     Params const& query,
     osr::ways const& w_,
     osr::lookup const& l_,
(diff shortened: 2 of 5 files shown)


pialwis:
Not sure why CI is red, I just want to update the doc :-)

openapi.yaml
```diff
@@ -932,7 +932,7 @@ paths:
       summary: |
         Street routing from one to many places or many to one.
         The order in the response array corresponds to the order of coordinates of the \`many\` parameter in the query.
-      operationId: oneToMany
+      operationId: getOneToMany
```
felixguendling: this is not compatible with existing client implementations?

openapi.yaml
```diff
@@ -1008,6 +1008,54 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+    post:
```
felixguendling, line 1023: looks like our openapi generator is not ready for this

felixguendling: ah, no. I think it's only that `oneToMany` is used in the code, so bringing it back should fix the build error

openapi.yaml
```diff
@@ -932,7 +932,7 @@ paths:
       summary: |
         Street routing from one to many places or many to one.
         The order in the response array corresponds to the order of coordinates of the \`many\` parameter in the query.
-      operationId: oneToMany
+      operationId: oneToManyGet
```
felixguendling: changing operation names in general is something I would try to avoid in order to keep existing client implementations compatible

pialwis: Fixed this. However I had to rename in the code `oneToManyPost` in order to comply with OpenAPI specs (unique `operationId`)

traines-source:
Now I recall that back in the day I semi-intentionally did not put that in there for security through obscurity, because if I'm not mistaken, we have no limit on the number of "manys" you could give. Maybe we should add such a limit to the limit config. (Or hard code it, since it can be quite high without causing issues I think.)

Issue #1158, opened 2025-11-10
Feature Request: more fields in responses


arnoclr:
Hello, I'm implementing a view in my app that list departures for a single line. I want to split departures into two directions, but the field `direction_id` is not exposed.

To filter by line, `route_id` is not exposed, so I use a combination of `route_short_name` / `route_color`, but there is not warranty to be unique.

Also, I want to show stops on a map, and group it by `parent_station` when there is a large amount of stops points, but this field is not available. I can't group it by name because some GTFS has more specific names for points and a generic name for the parent station.

Do you think it will be possible to add these 3 fields in the responses, or if it's not, can you recommend me some workarounds ? 

And thank you for your hard work.

felixguendling:
#1163 should solve this. Please reopen if necessary.

PR #1284, opened 2026-02-11
Prefix routeId with dataset tag in stoptimes and journey responses

yankeppey:
Hello!
Thank you for your work.
I noticed that `routeId` returned by MOTIS is not prefixed by dataset tag (i.e. agency name), unlike `stopId` and `tripId`, so it might be not unique across all the agencies, and so should be used with caution in the client apps when filtering, associating, etc. 
It seemed fairly easy to add the dataset tag, which would make it unique and consistent with stopIds, so here is the PR.

### Related
- https://github.com/motis-project/motis/issues/1158
- https://github.com/motis-project/motis/pull/1163

### DIFF
--- a/include/motis/tag_lookup.h
+++ b/include/motis/tag_lookup.h
@@ -31,9 +31,7 @@ struct tag_lookup {
   std::string id(nigiri::timetable const&,
                  nigiri::rt::run_stop,
                  nigiri::event_type) const;
-  std::string route_id(nigiri::timetable const&,
-                       nigiri::rt::run_stop,
-                       nigiri::event_type) const;
+  std::string route_id(nigiri::rt::run_stop, nigiri::event_type) const;
 
   trip_id<std::string> id_fragments(nigiri::timetable const&,
                                     nigiri::rt::run_stop,
--- a/src/endpoints/stop_times.cc
+++ b/src/endpoints/stop_times.cc
@@ -481,7 +481,7 @@ api::stoptimes_response stop_times::operator()(
                     std::string{tt_.strings_.try_get(agency.id_).value_or("?")},
                 .agencyName_ = std::string{tt_.translate(lang, agency.name_)},
                 .agencyUrl_ = std::string{tt_.translate(lang, agency.url_)},
-                .routeId_ = tags_.route_id(tt_, s, ev_type),
+                .routeId_ = tags_.route_id(s, ev_type),
                 .directionId_ = s.get_direction_id(ev_type) == 0 ? "0" : "1",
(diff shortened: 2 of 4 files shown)


felixguendling:
Not sure if that's worth a breaking change? Breaking changes require a new API version.

What's your use case of having scoped `route_id`? `route_id` is not fed back to MOTIS, so there's no issue with it not being scoped. Do you see any future use case that requires this change?

My guess is that a API user would want to match them with internal data which does not contain the tag. So this use case would not work anymore with this change.

yankeppey:
For my specific use case I added a custom endpoint `/api/v1/routes/stops?routeId=...` that would return stops for a given route, specified by `routeId`, so I had to feed routeId back to MOTIS. This is when I realized that the routeId by itself is not enough. So any potential filtering by route in the future would not work just with the "bare" route_id without knowing what agency/dataset it's coming from.

Even without those custom endpoints I think it'd make sense. 
- If MOTIS is my only data source, i.e. I don't have any internal data to do the matching with, having a non-unique routeId defeats the purpose of identifying a route. If I build a local cache for data coming from MOTIS, I'd need to use some sort of composite key `routeId:agencyName`, whereas the stopId can be used as a key directly to cache stops.
- If MOTIS is not my only data source, let's say I compare the raw GTFS data with the MOTIS-processed responses, then yes, I'd need to understand that MOTIS prefixes the raw route ids in a certain way. But that's already true for Stops, so it's arguably less of a cognitive burden to have the same matching logic for stops and for routes.

felixguendling:
In its current form, the PR destroys backward compatibility which is a no-go. For this change we would definitely need a new API version.

yankeppey:
Just to highlight a few points:
- `routeId` was added just three months ago as an attempt to provide a unique identifier for the route
- the `routeId` actually is not guaranteed to be unique across the datasets so it might become a source of subtle bugs for the clients that assume otherwise
- `routeId` format has never been documented ([in the same way stopIds or tripIds were](https://github.com/motis-project/motis/blame/master/docs/setup.md#L72)) so maybe it not too late to change it to the format consistent with stopIds and tripIds and document it

I am happy to change the PR so that it introduces, say, a separate field `scopedRouteId` alongside the existing `routeId`, in order to keep backward compatibility. 

But the decision is yours, I am not familiar with the project policies about API versions etc, feel free to close the PR or leave it open until other API changes accumulate, but IMHO the uniqueness issue should be addressed in some form.

felixguendling:
Then let's hope nobody is using it yet.

Issue #1275, opened 2026-02-09
Stoptimes: duration param


traines-source:
Offer a duration param (minutes?) in the API in addition to the `n` (number of results) param.
If both are given, use whatever returns more (i.e. interpret as minN and minDuration).

felixguendling:
> minutes

Another option would be to try to have all parameters in SI units.
I would not change existing parameters (to not break compatibility with existing clients) but maybe for new parameters, that could be an option?

traines-source:
As a string like `10min`? Will always be ugly in clients. But maybe a param can have multiple types (would also be useful for backwards compat)

felixguendling:
no, I just wanted to propose the unit to be seconds instead of minutes (still one number)

traines-source:
Ah ok. Yeah I think we have this in some places. For the duration that would be ok I guess but for longer timespans this is much less intuitive.

felixguendling:
In this specific case another option would be to have two timestamps (from+to).

traines-source:
Possible, but the existing param is named `time` (and it is non obvious in conjunction with arriveBy=true).

Issue #1020, opened 2025-09-18
one-to-all with coordinates as destinations


e-kotov:
Hello, and thank you for this awesome software!

I am trying to use `MOTIS` for some scientific work, where I need to calculate a large origin-destination trip duration matrix for trips on public transport. Usually routers provide an appropriate API for that, so does `MOTIS`, but currently only `WALK`, `BIKE`, `CAR` are supported. Do you have any estimates of when 'TRANSIT' would be supported for the `one-to-many` API endpoint? Calculating the routes one-by-one via `plan` API is very slow (I am guessing due to all the extra work of generating geometries and other info for the construction of the JSON response).

For context, calculation of a rather modest 183x183 time/distance matrix with `MOTIS`'s `plan` API takes about 1 minute on my laptop with 8 cores, with `detailedTransfers=false` and `timetableView=false` to minimize the amount of data to be returned from the API (I could not find any further options to minimize the output), whereas, from my tests, `R5` ( https://github.com/conveyal/r5 ) calculates this matrix in about 1-2 seconds. Further, comparing `R5` and `MOTIS` in the same mode, where each route plan is requested individually, instead of using `R5`'s table API, I get better speeds with `MOTIS`. So I would expect that implementation of 'TRANSIT' support in the `one-to-many` API easily beat even the `R5`.

felixguendling:
Have you looked at `/api/v1/one-to-all`?

e-kotov:
> Have you looked at `/api/v1/one-to-all`?

To be fair, I have not, as from the description of this API I expected it to return only the stop IDs and/or their coordinates. Now I have tested it and it works as I expected, but that's not what I need to do and it does not seem to be a replacement for my use case due to two reasons:

1. `/api/v1/one-to-all` has a limit of 90 minutes, which will not work for long distance travel on trains or trips with multiple interchanges with long waiting and/or walking times. This actually is a significant limitation for the public transit isochrones which I would want to do at some point too.
[...]

PR #1285, opened 2026-02-11
Implement One to Many for intermodal routing

MichaelKutzner:
Adds a first version for One-to-Many with intermodal routing.

There will be a GET and POST endpoint, as GET requests are limited to roughly 8.000 characters. Using POST does not cause this limitation, which can be relevant for analytic use cases.

Notice that there will be some changes between both endpoints as well as already existing endpoints:

* GET will be consistent to the already existing One-to-Many endpoint, which requires coordinates with a semicolon instead of the usual comma, i.e. `<lat>;<long>`.  
    For POST this restriction is not required. Therefore the POST endpoint allows the usual notation, as well as stop ids, which is identical to other endpoints.
* All units will be in seconds, even if other endpoints use the same parameter name with minutes, e.g. `minTransferTime`. This follows the suggestion made in #1275 
* Some often optional parameters will be required, e.g. `arriveBy` and `maxMatchingDistance`, to be similar to the existing One-to-Many endpoint. But I'm fine with making them optional as well.

Furthermore, defaults will be handled in the implementation, at least for now, as I couldn't make them optional for POST requests otherwise. But I think, that's something we can fix later without affecting users.

### DIFF
--- a/src/endpoints/one_to_all.cc
+++ b/src/endpoints/one_to_all.cc
@@ -133,7 +133,6 @@ api::Reachable one_to_all::operator()(boost::urls::url_view const& url) const {
 
   auto reachable = nigiri::bitvec{tt_.n_locations()};
   for (auto i = 0U; i != tt_.n_locations(); ++i) {
-    // TODO Use get_tmp() when implementing last mile
     if (state.get_best<0>()[i][0] != unreachable) {
       reachable.set(i);
     }


e-kotov:
Looking forward to this one! Will likely close my request in #1020 (maybe partially)

src/endpoints/one_to_all.cc
```diff
@@ -133,6 +133,7 @@ api::Reachable one_to_all::operator()(boost::urls::url_view const& url) const {
 
   auto reachable = nigiri::bitvec{tt_.n_locations()};
   for (auto i = 0U; i != tt_.n_locations(); ++i) {
+    // TODO Use get_tmp() when implementing last mile
```
felixguendling: get_tmp only contains the last update (max transfers looked at)
that's not sufficient for use here?

get_tmp also only contains trip arrivals, not walk arrivals, so it's not complete

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
+    Endpoint const& ep,
+    Query const& query,
+    place_t const& one,
+    std::vector<place_t> const& many,
+    auto const& time,
+    std::chrono::seconds const max_travel_time,
+    api::PedestrianProfileEnum const pedestrian_profile,
+    api::ElevationCostsEnum const elevation_costs,
+    osr_parameters const& osr_params) {
+  // TODO Should this always be calculated?
```
felixguendling: I don't see a way to nicely deduplicate it

openapi.yaml
```diff
@@ -3793,6 +4087,218 @@ components:
...
           type: boolean
+
+    OneToManyIntermodalParams:
+      type: object
+      required:
+        - one
+        - many
+        - maxTravelTime
+        - maxMatchingDistance
+        - arriveBy
```
felixguendling: those three can be defaulted?

openapi.yaml
```diff
@@ -1009,6 +1009,300 @@ paths:
...
+        - name: minTransferTime
+          in: query
+          required: false
+          description: |
+            Optional. Default is 0 seconds.
+
+            Minimum transfer time for each transfer in seconds.
+          schema:
+            type: integer
+
```
felixguendling, line 1192: sometimes new lines, sometimes not?
maybe consistently add new lines for clarity?

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
   return one_to_many_handle_request(query, w_, l_, elevations_);
 }
 
+template <typename Endpoint, typename Query>
+api::oneToManyIntermodal_response run_one_to_many_intermodal(
+    Endpoint const& ep,
+    Query const& query,
+    place_t const& one,
+    std::vector<place_t> const& many) {
+  ep.metrics_->routing_requests_.Increment();
```
felixguendling: I think that's a different kind of routing request and should have its own counter (or no counter at all)

@traines-source any opinions?

src/endpoints/one_to_all.cc
MichaelKutzner: True, it is not complete. But also causes incorrect results, if we only use `get_best()`, as targets might appear reachable within `x` minutes, also it's only reachable from a stop close to the final stop.
Last departure stop   --- Footpath --->   nearby stop   --- x Minutes --->   Destination

I will add a test, that should make this clear (for One-to-Many).

As I didn't intend to change the current (not correct) behavior in this Pull Request as well, I only added a note for now.

felixguendling: Using `get_tmp()` is no fix at all so this TODO is misleading and should be removed.

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
+    osr::location const& one,
+    std::vector<osr::location> const& many,
+    double const max_direct_time,
+    double const max_matching_distance,
+    osr::direction const dir,
+    osr_parameters const& params,
+    api::PedestrianProfileEnum const pedestrian_profile,
+    api::ElevationCostsEnum const elevation_costs,
+    osr::elevation_storage const* elevations_) {
+
```
felixguendling, line 45:
```suggestion
```

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
+      max_prepost_seconds);
+  auto const one_dir =
+      query.arriveBy_ ? osr::direction::kBackward : osr::direction::kForward;
+  auto const unreachable = query.arriveBy_
+                               ? n::kInvalidDelta<n::direction::kBackward>
+                               : n::kInvalidDelta<n::direction::kForward>;
+  constexpr auto const kInf = std::numeric_limits<double>::infinity();
+
+  auto const delta_to_seconds = [&](n::delta_t const d) -> double {
+    return 60.0 * (query.arriveBy_ ? -1 * d : d);
```
felixguendling: delta is a value defined relative to the base_ of raptor

if you mean a duration here, this can't be negative (travel times can't be negative, no time travel)

if get_fastest_one_to_all_offsets returns travel times, it should be a n::duration_t, not a delta_t

felixguendling:
> Looking forward to this one! Will likely close my request in #1020 (maybe partially)

@e-kotov you probably need the number of transfers as well (a Pareto set, not just the fastest (which means most transfers))?

@MichaelKutzner I don't see why we should artificially limit the output to not contain the full Pareto set?

src/endpoints/one_to_many.cc
felixguendling: with this adjustment, delta_to_seconds and duration_to_seconds can both be defined as functions (instead of lambda closures that require scope context)

felixguendling:
@e-kotov 

This also does not contain the much needed optimization to reuse street routing results

As I stated in https://github.com/motis-project/motis/issues/1020#issuecomment-3311202765

> The OSM routing from every last stop to the final destination can be shared across all one-to-all timetable routing requests. So to handle a WALK,TRANSIT (or any other mode + TRANSIT) table routing efficiently (i.e. reuse OSM routing results), and still keep MOTIS stateless, it would be easier to do one big many-to-many routing request.

So I think ideally, we would have some kind of cache (probably in data) with configurable cache size in `config.yml` to store street routing results and reuse them across 1:N requests in order to make N:M requests reasonably fast.

@MichaelKutzner what do you think?

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
+      *query.time_.value_or(openapi::now()));
+
+  auto const pedestrian_profile =
+      query.pedestrianProfile_.value_or(api::PedestrianProfileEnum::FOOT);
+  auto const elevation_costs =
+      query.elevationCosts_.value_or(api::ElevationCostsEnum::NONE);
+  auto const osr_params = get_osr_parameters(query);
+
+  // Get street routing durations
+  utl::verify(
```
felixguendling: `directModes` is not really correct, because there's only one mode supported. Are there any plans to support multiple modes? Otherwise, I would probably just call it `directMode`.

e-kotov:
> @e-kotov you probably need the number of transfers as well (a Pareto set, not just the fastest (which means most transfers))?

Yes, it would be very useful.

openapi.yaml
```diff
@@ -1009,6 +1009,300 @@ paths:
...
+            Maximum time in seconds for the last street leg.
+            Is limited by server config variable `street_routing_max_prepost_transit_seconds`.
+          schema:
+            type: integer
+            minimum: 0
+        - name: maxDirectTime
+          in: query
+          required: false
+          description: |
+            Optional. Default is 30min which is `1800`.
```
felixguendling, line 1369: there is no default?

default: 1800 is missing

felixguendling: same for all others?!

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
+                         ep.rt_,         nullptr, ep.gbfs_,    nullptr,
+                         nullptr,        nullptr, nullptr,     ep.metrics_};
+  auto gbfs_rd = gbfs::gbfs_routing_data{ep.w_, ep.l_, ep.gbfs_};
+
+  auto prepare_stats = std::map<std::string, std::uint64_t>{};
+
+  auto q = n::routing::query{
+      .start_time_ = time,
+      .start_match_mode_ = get_match_mode(one),
+      .start_ = r.get_offsets(nullptr, one, one_dir, one_modes, std::nullopt,
```
felixguendling, line 134: rtt=nullptr here but elsewhere it's used?

src/endpoints/one_to_many.cc
```diff
@@ -1,11 +1,290 @@
...
+
+  if (ep.tt_.locations_.footpaths_out_.at(q.prf_idx_).empty()) {
+    q.prf_idx_ = n::kDefaultProfile;
+  }
+
+  // Compute and update durations using transits
+
+  auto const state =
+      query.arriveBy_ ? n::routing::one_to_all<n::direction::kBackward>(
+                            ep.tt_, ep.rt_ ? ep.rt_->rtt_.get() : nullptr, q)
```
felixguendling: this will crash
rt live time management should be like in other endpoints
or remove (like I did it for one-to-all)

openapi.yaml
MichaelKutzner: That's true. But currently also intentional, as stated in the first post.
I assume it's more reasonable if `api::oneToManyIntermodal_params` and `api::OneToManyIntermodalParams` have the same member types than having `T` for `api::oneToManyIntermodal_params` and `optional<T>` for `api::OneToManyIntermodalParams`. And I couldn't set default values for the POST object, without having them mandatory. So any client had to set all default values again.
This might be a missing feature of openapi-cpp. And I assumed, that fixing this later shouldn't break too many clients, probably none. But even if it might break, I wasn't too worried about that, as the endpoint is not finalized (stable) yet.

Why fields become mandatory for POST requests, when members have type `T`, not `optional<T>`:
https://github.com/triptix-tech/openapi-cpp/blob/59e952c16208e653fecaaecc48e6db227b343b36/src/gen_types.cc#L495-L499
https://github.com/triptix-tech/openapi-cpp/blob/59e952c16208e653fecaaecc48e6db227b343b36/include/openapi/json.h#L16-L24

felixguendling: > is not finalized (stable) yet

Experimental for me means "we think it's good but maybe it isn't". Here it is more like "we know it's not good"? It should be possible to fix openapi-cpp. It was until now only used to produce JSON, not to consume JSON. So it's natural that it has some deficiencies in this area. But that doesn't mean we have to release it like that.

openapi.yaml
MichaelKutzner: That's something I suggested in my initial post. They are no longer required.

openapi.yaml
MichaelKutzner: Should be unified with a new line in between now.

src/endpoints/one_to_many.cc
MichaelKutzner: I removed the metric update for now. But we can add a better suited later.

src/endpoints/one_to_many.cc
MichaelKutzner: Not for now. I just intended it to be similar to the already existing endpoints.
But as we currently only support one mode, changing it as suggested makes it easier to maintain.

MichaelKutzner:
> @MichaelKutzner I don't see why we should artificially limit the output to not contain the full Pareto set?

The limitation was only, as I intended to make the API similar to the existing One-to-Many endpoint. But it's definitely something I can add next. Shouldn't be a big change.

felixguendling:
I think the routing at the destination could be one single osr call instead of `many` osr calls. Initialization at osr nodes can be done with the offsets computed by nigiri. nigiri can be initialized with the offsets from the initial osr call. So in the end it's 2x osr + 1x nigiri. But I'm not sure if caching might be a better solution.

grafik

Issue #998, opened 2025-09-06
`journey` / `refresh` endpoint


dabund24:
Hi, I think it would be cool if there was the possibility of getting one specific journey from a generated journey id included in the itinerary object. Not only transitious.org would benefit from this by making single journeys shareable, but also other downstream projects and libraries like motis-fptf-client.

Such a journey id would need to include the following data for each leg having a trip id:
- the trip id
- the id of the stop where the vehicle is boarded
- the scheduled date-time when the vehicle is boarded (for edge cases where a trip passes through a stop multiple times)
- the id of the stop where the vehicle is alighted
- the scheduled date-time when the vehicle is alighted

Moreover, including information about the legs to the first transit stop and from the last transit stop would be necessary if they do exist:
- the `fromPlace` string of the search
- the pre-transit mode
- the `toPlace` string of the search
- the post-transit mode

This information *should* be enough to reconstruct full journey by first getting all trips using the trip ids, then "truncating" 
them such that they correspond exactly to the journey leg using the departure/arrival date-times/stop ids. Lastly, all gaps need to be filled with walks and the first and last legs need to be computed if they not of a public transport type.
As I am not familiar at all with the structure of motis, I'm not sure how feasible this is in practice, so let me know if this is not possible for some reason. I have already seen the second bullet point at #771, but I'm not really sure if the restriction only refers to that specific pr or to the whole api in general.

related: https://github.com/public-transport/transitous/issues/899

felixguendling:
### Design of a stable itinerary ID that allows an itinerary to be updated with real-time information even if identifiers that we usually use are not stable due to static timetable updates

What is not stable but can be used as a "hint" for faster + more reliable lookup after we load a new static timetable:
- location_idx_t, trip_idx_t are only internal and are not even supposed to be stable (so they are irrelevant for our goal of having a stable itinerary ID)
- stop IDs and trip IDs can and will change across static timetable updates
- coordinates might change slightly, though a physical stop won't usually move a lot
- trip statically scheduled times might change slightly, but usually won't be hours off

An itinerary is made up of several legs. Each leg has a trip that needs to be identified.

Ideally, we can keep the ID string as short as possible.

The ID should contain
- trip_id
- from stop_id
- to stop_id
- from stop scheduled time
- to stop scheduled time
- from stop coordinate
- to stop coordinate

Compression ideas:
- Coordinates could be compressed using Google polyline encoding (already implemented in our own geo library decode+encode functions).
[...]

PR #1286, opened 2026-02-11
Itinerary-id

Kiarash-Parvizi:
Replaces #1269

Related Issue: #998


src/itinerary_id.cc
```diff
@@ -0,0 +1,398 @@
...
+#include "nigiri/rt/frun.h"
+
+#include "motis/journey_to_response.h"
+#include "motis/timetable/time_conv.h"
+
+#include "motis/place.h"
+#include "motis/tag_lookup.h"
+
+#include "geo/polyline_format.h"
+
```
traines-source, line 50: group by package

src/itinerary_id.cc
```diff
@@ -0,0 +1,398 @@
...
+namespace {
+constexpr auto kItineraryTimeBase = std::chrono::sys_days{
+    std::chrono::year{2026} / std::chrono::month{1} / std::chrono::day{1}};
+constexpr auto TimeMUL = 1.0 / 60.0 * 7;
```
traines-source:
```suggestion
constexpr auto kTimeMUL = 1.0 / 60.0 * 7;
```

src/itinerary_id.cc
```diff
@@ -0,0 +1,398 @@
...
+struct leg_hint {
+  std::string trip_id;
+  std::string from_stop_id;
+  std::string to_stop_id;
+  std::int64_t sched_start;
+  std::int64_t sched_end;
+  //
+  geo::latlng from_latlng;
+  geo::latlng to_latlng;
+  motis::api::ModeEnum mode;
```
traines-source: members at the end of the struct, trailing _
sorry in case we didn't link that https://github.com/motis-project/motis/blob/master/docs/STYLE.md before

src/itinerary_id.cc
```diff
@@ -0,0 +1,398 @@
...
+  std::ostringstream oss;
+  oss << mode;
+
+  auto const t =
+      std::chrono::system_clock::time_point{std::chrono::seconds{sched_start}};
+  auto const iso = fmt::format("{:%FT%TZ}", t);
+
+  auto const query = fmt::format(
+      "?center={},{}&time={}&arriveBy={}&direction=LATER&window={}"
+      "&radius={}&exactRadius=true&fetchStops=true&mode={}",
```
traines-source: I still think it would be worth it slightly refactoring out a stoptimes function which takes a `api::stoptimes_params`, similar to how we've done it with the onetomany endpoint. Building a query string feels ugly
Could possibly also return fruns directly, maybe this makes things more complicated not sure

src/itinerary_id.cc
```diff
@@ -0,0 +1,398 @@
...
+  auto root = boost::json::object{};
+  root["legs"] = std::move(legs);
+  //
+  root["journey_start"] = max_sched_t - itin.duration_;
+  return boost::json::serialize(root);
+}
+
+api::Itinerary reconstruct_itinerary(motis::ep::stop_times const& stoptimes_ep,
+                                     std::string const& itin_id) {
+  constexpr auto lookback_t = 8 * 60;
```
traines-source: This is a bit large?
Correct me if I'm wrong, but you're querying on the stoptimes endpoint including realtime data? I think you need to query on the stoptimes without RT (in order to get the ordering according to scheduled data) and only then build the fruns including RT data. Otherwise the shifts will be much larger still than 8 minutes.
In that regard it would be useful to also have tests with RT data (in nigiri there are lots of RT tests for copypasting, including for ADDED trips)

src/itinerary_id.cc
```diff
@@ -0,0 +1,398 @@
...
+  constexpr auto with_fares = false;
+  constexpr auto with_scheduled_skipped_stops = false;
+  constexpr auto api_version = 0;
+
+  return journey_to_response(
+      stoptimes_ep.w_,
+      nullptr,  // osr::lookup **
+      stoptimes_ep.pl_, stoptimes_ep.tt_, stoptimes_ep.tags_, nullptr, nullptr,
+      rtt, stoptimes_ep.matches_, nullptr,
+      nullptr,  // shapes **
```
traines-source: you will probably want shapes to return polylines?

src/itinerary_id.cc
Kiarash-Parvizi: Yeah, I'm planing to change this later.

src/itinerary_id.cc
Kiarash-Parvizi: Sure. I'll clean up the code once every other problem is dealt with :)
I might have also included more data than we need. Later, all these will be cleaned up.

traines-source (commented):
I think we can just merge src/itinerary_id.cc into endpoints/itinerary_id.cc ?
Other than that and apart from the already discussed issues and code style, LGTM

src/itinerary_id.cc
```diff
@@ -0,0 +1,473 @@
...
+    auto const arr = to_rs.time(n::event_type::kArr);
+
+    legs.push_back({n::direction::kForward, from_rs.get_location_idx(),
+                    to_rs.get_location_idx(), dep, arr,
+                    n::routing::journey::run_enter_exit{fr, from_idx, to_idx}});
+  }
+
+  auto j = n::routing::journey{.legs_ = legs,
+                               .start_time_ = journey_start,
+                               .dest_time_ = legs.back().arr_time_,
```
traines-source: this leads to weird times for delayed trips (because once the delay is taken into account, once not)

src/endpoints/itinerary_id.cc
```diff
@@ -0,0 +1,31 @@
...
+  auto const stop_times_ep =
+      ep::stop_times{config_, w_,         pl_, matches_, ae_,
+                     tz_,     loc_rtree_, tt_, tags_,    kStaticOnlyRt};
+  try {
+    auto const query = api::refreshItinerary_params{url.params()};
+    return reconstruct_itinerary(stop_times_ep, query.itineraryId_, rt.get());
+  } catch (std::exception const& e) {
+    throw net::bad_request_exception{std::string{"invalid itineraryId: "} +
+                                     e.what()};
+  }
```
traines-source: you can use `utl::verify<net::bad_request_exception>` directly in the code instead

openapi.yaml
```diff
@@ -1577,6 +1577,38 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /api/v1/refresh-itinerary:
```
traines-source: if we intend to merge this, we should put it under /api/experimental, otherwise people will try and start using it...

traines-source:
In general, this looks good to me.

I think it would be good to get the ID format right before merging this, because the nature of these IDs is to be stored somewhere and then it gets annoying if they become incompatible. For the same reason, I think we need a backwards-compatible, human-readable format, which for me would be a condensed key-value format:
1. a list delimiter (separating the legs), e.g. `;` (for this iteration not needed, but should be compatible, i.e. character needs to be reserved!)
2. an entry delimiter, e.g. `,`
3. a key-value delimiter, e.g. `=`
4. single-letter keys

e.g. `S=start_stop_id_1,D=dest_stop_id_2,X=abc,Y=def;S=start_stop_id_3,D=dest_stop_id_4,X=hij,Y=klm`

One could choose more rarely used delimiters, but in any case one would need to escape these characters in stop/tripids (maybe by url encoding them? or some other way which still allows simple splitting/joining to obtain the parts). 
If necessary, one can optionally still compress this string later on while keeping backwards compatibilty.

felixguendling:
We need to think about:

- format:
  - I think we could just use the final refresh URL as ID.
  - For this, we need to think about how to represent arrays.
    - The current way with separator is not a good fit here because any separator can be used in IDs as well.
    - We could extend openapi-cpp to support repeated key as arrays `?x=1&x=2&x=3` means `x=[1,2,3]`
- matching criteria:
  - I recently had an example of a change of ~45min for a night train. The train was obviously the same according to the train number (for us: either `displayName` or `tripShortName`). I think for users, the name is the primary way to identify "is this the same trip"
  - Clasz/mode changes are critical for ticketing purposes. If a system displays a tram and we change it to a bus with the same start+destination this might confuse the user because it might be that her ticket is only for the tram, not the bus. We might need to introduce matching configuration flags for this
  - Definitely a must to at least check - and inform the user if we still match but they differ from the search criteria - alternatively we can also not match at all and tell the user to search again:
    - wheelchair accessibility (not yet included in the response, but this will come soon)
    - cars_allows (cars on ferries / trains)
    - bikes_allowed (bikes in buses / trains / etc.)
- another question: one id for the whole itinerary and one for each leg or only for the whole itinerary? I think having one per leg might also have use cases.
- can we leverage a similar approach as in pong.cc `get_earliest_alternative` or directly use the get events function in stop_times.cc to not have to find the `frun`s again?

My goal is to have the least amount of bug reports with this in terms of "it matched a trip that is a completely different than I expected" - then it's probably better to tell the user "your trip doesn't exist anymore -> please search again".

grafik

Kiarash-Parvizi:
>   * I think we could just use the final refresh URL as ID.
>   * For this, we need to think about how to represent arrays.
>     
>     * The current way with separator is not a good fit here because any separator can be used in IDs as well.
>     * We could extend openapi-cpp to support repeated key as arrays `?x=1&x=2&x=3` means `x=[1,2,3]`

I think this wouldn't give us much benefit. It'd make it more unnecessary complex and unclear, especially for multi-leg itineraries. Keeping a single JSON string still seems simpler, and if we later want a more ***compact representation***, we can still re-encode or compress that one field.

Kiarash-Parvizi:
>   * I recently had an example of a change of ~45min for a night train. The train was obviously the same according to the train number (for us: either `displayName` or `tripShortName`). I think for users, the name is the primary way to identify "is this the same trip"
> * another question: one id for the whole itinerary and one for each leg or only for the whole itinerary? I think having one per leg might also have use cases.

If we allow such large changes in time, then it makes no sense to keep both sched_start_ and sched_end_ for all legs. One such delay would have a domino effect, changing the ones after it. Therefore I'm thinking to only keep duration for legs after the first one. Based on this argument, keeping a separate id for each leg is also not a good idea.

Kiarash-Parvizi:
Furthermore, the goal of covering delays of up to 45 minutes or an hour might contradict the following constraint variable: `stoptimes_max_results_{256U}`

src/endpoints/refresh_itinerary.cc
```diff
@@ -5,19 +5,52 @@
...
-      config_, w_, pl_, matches_, ae_, tz_, loc_rtree_, tt_, tags_, static_rt};
   auto const query = api::refreshItinerary_params{url.params()};
+  auto const stop_times_ep = make_scheduled_stop_times(*this);
+
+  auto const rt = std::atomic_load(&rt_);
   return reconstruct_itinerary(stop_times_ep, shapes_, *rt, query.itineraryId_);
 }
 
+api::Itinerary refresh_itinerary_post::operator()(
+    boost::json::object const& body) const {
```
felixguendling: does this also work with the generated type?

Kiarash-Parvizi: Yes. But my concern is to include inputs that protobuf json accepts. Like snake_case field names and quoted int64 values.

felixguendling: > Yes

then use the generated type

src/itinerary_id.cc
```diff
@@ -0,0 +1,495 @@
...
+       i_from < from_cands.size() && i_to < to_cands.size();) {
+
+    switch (candidate_score_cmp_ids(from_cands[i_from], to_cands[i_to])) {
+      case -1: ++i_from; break;
+      case +1: ++i_to; break;
+      case 0:
+        auto score = from_cands[i_from].score_ + to_cands[i_to].score_;
+        if (score > best_score) {
+          best_score = score;
+          best_from_to.emplace(from_cands[i_from].candidate_,
```
felixguendling, line 410: there's no check that they are within the same transport?

itinerary_id.proto
```diff
@@ -0,0 +1,20 @@
...
+  string from_id = 3;
+  string to_id = 4;
+  string coords = 5;
+  int64 sched_start = 6;
+  int64 sched_end = 7;
+  string mode = 8;
+  bool scheduled = 9;
+}
+
+message SingleLegItineraryId {
```
felixguendling: it should work for the whole journey
this looks like a workaround

openapi.yaml
```diff
@@ -4375,13 +4381,61 @@ components:
...
+        schedEnd:
+          type: integer
+          format: int64
+          description: Scheduled arrival time as a Unix timestamp in seconds.
+        mode:
+          $ref: '#/components/schemas/Mode'
+        scheduled:
+          type: boolean
+
+    SingleLegItineraryIdProto:
```
felixguendling: remove Proto postfix

src/itinerary_id.cc
```diff
@@ -0,0 +1,495 @@
...
+  to_cands.reserve(to_resp.size());
+
+  for (auto const& st : from_resp) {
+    if (!st.place_.scheduledDeparture_.has_value()) {
+      continue;
+    }
+
+    from_cands.emplace_back(
+        &st,
+        (st.displayName_ == hint.display_name_ ? kExactTripNameMatchAddScore
```
felixguendling: display name has to match exactly except `requireDisplayName=false` (default true)

src/itinerary_id.cc
```diff
@@ -0,0 +1,495 @@
...
+      place.scheduledArrival_ = {s.scheduled_time(n::event_type::kArr)};
+    }
+    if (fr.stop_range_.from_ != fr.size() - 1U) {
+      place.departure_ = {s.time(n::event_type::kDep)};
+      place.scheduledDeparture_ = {s.scheduled_time(n::event_type::kDep)};
+    }
+
+    auto const trip_id = st_ep.tags_.id(st_ep.tt_, s, ev_type);
+
+    return {.place_ = std::move(place),
```
felixguendling: only work with the run - no place necessary

felixguendling (changes requested):

itinerary_id.proto
Kiarash-Parvizi: Didn't we just force everything to be limited to a single leg?
We agreed to merge this PR only for single leg version.

src/itinerary_id.cc
Kiarash-Parvizi: Why is this the right thing to do?

itinerary_id.proto
felixguendling: The protobuf format should be final and can't be changed anymore to something completely different. It has to be downward compatible.

src/itinerary_id.cc
felixguendling: API parameter is required because we don't know all potential use cases and want to support a broad range.

Most use cases I can imagine require exact name matching because that's what the user considers to be "the same trip".

openapi.yaml
```diff
@@ -2026,6 +2026,74 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /api/experimental/refresh-itinerary:
```
felixguendling, line 2029: we need an API version because this determines how the returned itinerary will look like (and different clients are compatible to different versions)

this will have to be in-sync with the plan endpoint version

and probably it's best to move to "one API version for everything"

itinerary_id.proto
```diff
@@ -0,0 +1,20 @@
...
+  string coords = 5;
+  int64 sched_start = 6;
+  int64 sched_end = 7;
+  string mode = 8;
+  bool scheduled = 9;
+}
+
+message ItineraryId {
+  repeated LegId leg = 1;
+  string lang = 2;
```
felixguendling, line 19: lang and other parameters (detailed legs, etc.) should be URL parameters

src/itinerary_id.cc
```diff
@@ -0,0 +1,506 @@
...
+    return std::tie(a.run_.t_, a.run_.rt_) == std::tie(b.run_.t_, b.run_.rt_);
+  }
+
+  friend bool operator<(st_candidate const& a, st_candidate const& b) {
+    return std::tie(a.run_.t_, a.run_.rt_) < std::tie(b.run_.t_, b.run_.rt_);
+  }
+
+  std::string stopId_{};
+  std::string tripId_{};
+  std::string displayName_{};
```
felixguendling, line 161: not necessary to store as they can be derived from run directly
in general: try to avoid allocations, especially if they are completely not necessary

src/itinerary_id.cc
```diff
@@ -0,0 +1,506 @@
...
+  auto blocked = osr::bitvec<osr::node_idx_t>{};
+  auto gbfs_rd = gbfs::gbfs_routing_data{};
+
+  // routing defaults
+  constexpr auto join_interlined_legs = true;
+  constexpr auto detailed_transfers = false;
+  constexpr auto detailed_legs = true;
+  constexpr auto with_fares = false;
+  constexpr auto with_scheduled_skipped_stops = false;
+  constexpr auto api_version = 0;
```
felixguendling, line 313: need to be from URL

itinerary_id.proto
Kiarash-Parvizi: `lang` is from when the user initially requested and created the id. I think we should store this along side `display_name`. Otherwise we rely on the user not to change settings.

felixguendling: lang and other parameters (detailed legs, etc.) should be URL parameters

Kiarash-Parvizi: Ok. But maybe writing another line isn't a bad idea.

felixguendling: > Otherwise we rely on the user not to change settings.

no user will ever use this API

apps use it and they want to be able to adapt to the browser setting or settings in the customer panel of the user

so being able to change the language with the `language` parameter like most other MOTIS APIs makes more sense than hardcoding it

in general, all parameters that control the contents of a journey from the plan endpoint should also be available here

openapi.yaml
Kiarash-Parvizi: > and probably it's best to move to "one API version for everything"

I didn't understand this part. Is this sth I need to do in this PR?


Also should I just make a few copies of this with different versions:
```
GET<ep::refresh_itinerary>("/api/v5/refresh-itinerary", d);
```

felixguendling: > I didn't understand this part. Is this sth I need to do in this PR?

No.


> Also should I just make a few copies of this with different versions:

Yes, that's the way to go for now until we do it properly.

itinerary_id.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package motis.proto;
+
+message LegId {
+  string display_name = 1;
+  string trip_id = 2;
+  string from_id = 3;
+  string to_id = 4;
+  string coords = 5;
```
felixguendling: should be 2x latlng, not string

openapi.yaml
felixguendling: for now we only need v5 because there can't be any old clients and we can just force old clients to update

openapi.yaml
```diff
@@ -4488,13 +4597,94 @@ components:
...
+      properties:
+        displayName:
+          type: string
+        tripId:
+          type: string
+        fromId:
+          type: string
+        toId:
+          type: string
+        coords:
```
felixguendling: 2x latlng

itinerary_id.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package motis.proto;
```
felixguendling: just motis is fine

itinerary_id.proto
Kiarash-Parvizi: But isn't this polyline format more compact?

felixguendling: depends on the binary representation of string
but probably not much more compact, so not a reasonable tradeoff
-> switch to 2x latlng

src/itinerary_id.cc
```diff
@@ -0,0 +1,674 @@
...
+  void decode_coords(std::string_view const encoded_coords) {
+    auto const coords = geo::decode_polyline(encoded_coords);
+    utl::verify(coords.size() == 2,
+                "itinerary id: coords must decode to exactly 2 points, got {}",
+                coords.size());
+
+    from_pos_ = coords[0];
+    to_pos_ = coords[1];
+  }
+};
```
felixguendling, line 117: this can be removed? - just use the generated type proto_leg_t

Kiarash-Parvizi: I recently removed the other constructor that accepted json.
I think it's a nice abstraction to have and we can always easily remove it if we want to.
It also does some preprocessing and format checking.

src/itinerary_id.cc
```diff
@@ -0,0 +1,674 @@
...
+
+  n::location_idx_t from_loc_;
+  n::location_idx_t to_loc_;
+  std::optional<geo::latlng> from_pos_;
+  std::optional<geo::latlng> to_pos_;
+  n::unixtime_t dep_;
+  n::unixtime_t arr_;
+};
+
+using leg_item = std::variant<resolved_run, walk_segment>;
```
felixguendling: why can't this be a nigiri::journey::leg directly?

src/itinerary_id.cc
```diff
@@ -0,0 +1,674 @@
...
+  auto const parsed_id = decode_itinerary_id(id);
+
+  auto const get_run = [&](leg_hint const& lh) -> resolved_run {
+    if (!lh.scheduled_) {
+      utl::verify(!lh.trip_id_.empty(),
+                  "reconstruct_itinerary: additional trip requires trip_id");
+
+      auto const fr = make_frun_from_trip_id(
+          stop_times_ep.tags_, stop_times_ep.tt_, rt.rtt_.get(), lh.trip_id_);
+      utl::verify(fr.valid(),
```
felixguendling: not found legs should become dummy legs that contain the information from the itinerary-id

if a user presses the refresh button and from 4 legs 1 could not be recovered, it should show the three legs that could be recovered and for the one missing leg, it should show that it could not be recovered

I think we have to add a refresh functionality in the MOTIS debug UI in this PR to test the API for real-world readiness before publishing

src/itinerary_id.cc
Kiarash-Parvizi: We'll be needing fields like:
```
  std::optional<geo::latlng> from_pos_;
  std::optional<geo::latlng> to_pos_;
```
But I'm thinking about changing it.

traines-source:
What's the status here? We want to make a new release soon and I think this would be important (albeit not absolutely required) to add.

Kiarash-Parvizi:
> What's the status here? We want to make a new release soon and I think this would be important (albeit not absolutely required) to add.

There's usually a lot of going back and forth in this PR. When is the new release scheduled for?

PR #1288, opened 2026-02-11
stop_times API: add center coordinate parameter

Kiarash-Parvizi:
Center field is added to stoptimes_ep

The behavior:
check if id is set and exists in the timetable -> just use id like before
otherwise -> use center as anchor coordinate instead
in case center is also null -> panic

Note:
radius is required when using center without stopId

### DIFF
--- a/src/endpoints/stop_times.cc
+++ b/src/endpoints/stop_times.cc
@@ -352,14 +352,13 @@ api::stoptimes_response stop_times::operator()(
   auto const max_results = config_.limits_.value().stoptimes_max_results_;
   utl::verify<net::too_many_exception>(
       query.n_ < max_results, "n={} > {} not allowed", query.n_, max_results);
+  utl::verify<net::bad_request_exception>(
+      query.stopId_.has_value() ||
+          (query.center_.has_value() && query.radius_.has_value()),
+      "no stop and no center with radius (at least one is required)");
 
-  auto const query_stop = query.stopId_.and_then([&](std::string const& x) {
-    auto const loc = tags_.find_location(tt_, x);
-    utl::verify<net::not_found_exception>(
-        loc.has_value() || query.center_.has_value(),
-        "stopId '{}' not found and no alternative center is provided", x);
-    return loc;
-  });
+  auto const query_stop = query.stopId_.and_then(
+      [&](std::string const& x) { return tags_.find_location(tt_, x); });
 
   auto const query_center = query.center_.and_then(
       [&](std::string const& x) { return parse_location(x); });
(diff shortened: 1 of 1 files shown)


src/endpoints/stop_times.cc
```diff
@@ -352,8 +352,28 @@ api::stoptimes_response stop_times::operator()(
   utl::verify<net::too_many_exception>(
       query.n_ < max_results, "n={} > {} not allowed", query.n_, max_results);
 
-  auto const x = tags_.get_location(tt_, query.stopId_);
-  auto const l = tt_.locations_.get_root_idx(x);
+  auto anchor_stop = std::optional<n::location_idx_t>{};
+  if (query.stopId_.has_value() && !query.stopId_->empty()) {
+    try {
+      anchor_stop = tags_.get_location(tt_, *query.stopId_);
```
felixguendling: empty catch = throw/catch for control flow = suspicious

seems like we kind of expect unknown stopIds here -> not exceptional -> no exceptions should be thrown

- implement tags_.find_location (returning optional)
- rewrite tags_.get_location to make use of tags_.find_location
- use tags_.find_location here

src/tag_lookup.cc
```diff
@@ -141,14 +141,24 @@ std::pair<nigiri::rt::run, nigiri::trip_idx_t> tag_lookup::get_trip(
...
+    nigiri::timetable const& tt, std::string_view s) const {
   auto const [tag, id] = split_tag_id(s);
   auto const src = get_src(tag);
   try {
     return tt.locations_.location_id_to_idx_.at({{id}, src});
   } catch (...) {
-    throw utl::fail<net::not_found_exception>(
-        R"(Could not find timetable location "{}", tag="{}", id="{}", src={})",
-        s, tag, id, static_cast<int>(to_idx(src)));
+    return std::nullopt;
```
felixguendling, line 163: that's not better, just moving the same slop around
don't throw exceptions if it's expected
use find instead of at

felixguendling:
Our aim is to always have tests :)

src/endpoints/stop_times.cc
```diff
@@ -352,8 +352,32 @@ api::stoptimes_response stop_times::operator()(
...
+  utl::verify<net::bad_request_exception>(
+      center.has_value(), "no coordinates: stop_found={}, center_parsed={}",
+      query_stop.has_value(), query_center.has_value());
+
+  utl::verify<net::bad_request_exception>(
+      query_stop.has_value() ||
+          (query_center.has_value() && query.radius_.has_value()),
+      "no radius: stop_found={}, center_parsed={}", query_stop.has_value(),
+      query_center.has_value());
+
```
traines-source: I think it would be good if at least in the standard case (just a stop id is given and this stop id doesn't exist), we still return a 404 as before.

openapi.yaml
```diff
@@ -1529,9 +1529,16 @@ paths:
...
-          required: true
+          required: false
           description: stop id of the stop to retrieve departures/arrivals for
 
+        - name: center
+          in: query
+          schema:
+            type: string
+          required: false
+          description: Anchor coordination (can be used instead of stopId)
```
traines-source: I think it would be useful to describe the fallback behavior in case both are given (because my idea was that clients can actually use that if they remember the position from a previous query – currently we get a lot of 404s due to queries with old stopids).

src/endpoints/stop_times.cc
traines-source: Yes for the itinerary use case, we said that for that reason you will just give the center anyways.

I just meant that if a client is requesting just with a stopid, no center (i.e. same request as before), and if then no stop with that stop id exists, we should return a 404, because the client hasn't created a "bad request", it's just that the stop was not found.
(Conversely, one could return 400 already when center is set without radius, even if stopid is set and exists, but that is a minor detail.)

PR #1289, opened 2026-02-11
ignore OS and editor artifacts

Kiarash-Parvizi:


.gitignore
```diff
@@ -21,3 +21,6 @@ config.yaml
 config.yml
 config.ini
 
+.DS_Store
```
felixguendling: move those to `.idea` (also editor artifact)

PR #1293, opened 2026-02-14
precompute bbox

### LINKED ISSUE #1276: /map/initial: precompute bbox
Precompute bbox once on import and store it in `timetable.h`. It takes half a second to compute on Transitous on every map load.

DIDONAX:
- fixes #1276
-  depends on https://github.com/motis-project/nigiri/pull/314

### DIFF
--- a/src/endpoints/initial.cc
+++ b/src/endpoints/initial.cc
@@ -9,58 +9,15 @@
 
 #include "nigiri/timetable.h"
 
-namespace n = nigiri;
-
 namespace motis::ep {
 
 api::initial_response initial::operator()(boost::urls::url_view const&) const {
-  auto const get_quantiles = [](std::vector<double>&& coords) {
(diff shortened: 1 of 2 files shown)


felixguendling:
if it's stored in `nigiri::timetable` it should be computed in [`init_finish.cc`](https://github.com/motis-project/nigiri/blob/master/src/loader/init_finish.cc).

The name `tt.bbox_` implies that it's a bounding box covering every stop/trip/shape of the timetable which is not true.

DIDONAX:
> if it's stored in `nigiri::timetable` it should be computed in [`init_finish.cc`](https://github.com/motis-project/nigiri/blob/master/src/loader/init_finish.cc).

in this case, would it be better to rewrite the necessary part from tiles  in nigiri (~ 15 lines of code), or move them permanently to the geo library since it's already included in nigiri  as dependency?

felixguendling:
I think they can be moved to geo.

Issue #887, opened 2025-06-03
Include distance in meters in /api/v1/one-to-many response


christoph-daehne:
This is a feature request ;) 

In the response of [/api/v1/one-to-many](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/motis-project/motis/refs/heads/master/openapi.yaml#tag/routing/operation/oneToMany) the `distance` in meters to/from the locations can be included next to `duration` in seconds:

```json
[
  {
    "duration": 900,
    "distance": 14
  }
]
```

It seems, the [`distance` is already part of the `paths`](https://github.com/motis-project/osr/blob/master/include/osr/routing/route.h#L37 ) during computation and discarded while [creating the JSON response](https://github.com/motis-project/motis/blob/master/src/endpoints/one_to_many.cc#L44).

felixguendling:
The idea of the one-to-many endpoint is that it's fast because it doesn't reconstruct all actual journeys (path segments are not filled in osr). To compute the exact distance by adding segments a reconstruction would be needed. This should at least not be the default option.

christoph-daehne:
I did not realize this has such an impact, but though about it as a low handing fruit.

Never mind then ;)

felixguendling:
I think it's doable. It's just not something I would enable by default. Doing the resolving of locations to the graph and routing only once has still advantages. It's just more expensive to reconstruct the actual path than to just output the duration. So there should still be the option to do only the routing without path reconstruction for the distances for use cases that require speed but don't need the actual distance.

e-kotov:
This now works beautifully and returns something like this:

```json
[{"duration":1.118E3,"distance":1.1496806713825708E4},{"duration":1.388E3,"distance":1.2063496174616757E4},
{"duration":6.28E2,"distance":6.3267072938822885E3},
{"duration":5.41E2,"distance":5.004119406358705E3},
{"duration":1.471E3,"distance":1.9885868140694765E4},{"duration":9.29E2,"distance":6.736525078081438E3},
{"duration":1.807E3,"distance":1.767186036835344E4},
{"duration":9.38E2,"distance":5.865922498160905E3},{"duration":1.646E3,"distanc
```

As well as some empty ones in case route was not found (points did not snap, out of max time window, etc.):

```json
"distance":5.130436728988314E5},{},{},{},{},{},{},{},{},{},{},{},{},{},{},
[...]

PR #1295, opened 2026-02-15
feat(one-to-many): optional distance in one-to-many

e-kotov:
Adding distance arg and distance output to one-to-many.

I tested a build of this locally on my mac and also on a linux machine, this worked very well.

---

Ideally, I would actually suggest a much bigger revamp of this api endpoint.

While I was working on it, I noticed that it is the only one that asks for seconds, rather than minutes. Also, I would like to add optional debug output to it, but this would be a breaking change requiring new v2 endpoint, as the format of the output would fundamentally change to a different form of json. But that's out of scope for this PR.

### DIFF
--- a/include/motis/endpoints/one_to_many.h
+++ b/include/motis/endpoints/one_to_many.h
@@ -20,6 +20,9 @@ api::oneToMany_response one_to_many_handle_request(
     osr::ways const& w_,
     osr::lookup const& l_,
     osr::elevation_storage const* elevations_) {
+  // required field with default value, not std::optional
+  static_assert(std::is_same_v<decltype(query.withDistance_), bool>);
+
   auto const one = parse_location(query.one_, ';');
   utl::verify(one.has_value(), "{} is not a valid geo coordinate", query.one_);
 
@@ -37,22 +40,23 @@ api::oneToMany_response one_to_many_handle_request(
 
   auto const profile = to_profile(query.mode_, api::PedestrianProfileEnum::FOOT,
                                   query.elevationCosts_);
-  auto const with_distance = query.distance_.value_or(false);
+
   auto const paths = osr::route(
       to_profile_parameters(profile, get_osr_parameters(query)), w_, l_,
       profile, *one, many, query.max_,
       query.arriveBy_ ? osr::direction::kBackward : osr::direction::kForward,
       query.maxMatchingDistance_, nullptr, nullptr, elevations_,
-      std::function<bool(osr::path const&)>{
-          [with_distance](osr::path const&) { return with_distance; }});
+      [&](auto&&) { return query.withDistance_; });
 
   return utl::to_vec(paths, [&](std::optional<osr::path> const& p) {
-    return p.has_value()
-               ? api::Duration{.duration_ = p->cost_,
-                               .distance_ = with_distance
-                                                ? std::optional{p->dist_}
-                                                : std::nullopt}
-               : api::Duration{};
+    return p
+        .transform([&](osr::path const& x) {
+          return api::Duration{.duration_ = x.cost_,
+                               .distance_ = query.withDistance_
+                                                ? std::optional{x.dist_}
+                                                : std::nullopt};
+        })
+        .value_or(api::Duration{});
   });
 }
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -997,7 +997,7 @@ paths:
             false = one to many
           schema:
             type: boolean
-        - name: distance
+        - name: withDistance
           in: query
           required: false
           description: |
@@ -3855,7 +3855,7 @@ components:
             true = many to one
             false = one to many
           type: boolean
-        distance:
+        withDistance:
           description: |
             Optional. Default is `false`.
             If true, the response includes the distance in meters


e-kotov:
Ah, importantly, it directly addresses #887.

include/motis/endpoints/one_to_many.h
```diff
@@ -37,15 +37,34 @@ api::oneToMany_response one_to_many_handle_request(
 
   auto const profile = to_profile(query.mode_, api::PedestrianProfileEnum::FOOT,
                                   query.elevationCosts_);
+  auto const with_distance = [](auto const& v) {
+    if constexpr (requires { v.value_or(false); }) {
+      return v.value_or(false);
+    } else {
+      return static_cast<bool>(v);
+    }
+  }(query.distance_);
```
felixguendling: isn't this `auto const with_distance = query.distance_.value_or(false);`

include/motis/endpoints/one_to_many.h
```diff
@@ -37,15 +37,34 @@ api::oneToMany_response one_to_many_handle_request(
...
   auto const paths = osr::route(
       to_profile_parameters(profile, get_osr_parameters(query)), w_, l_,
       profile, *one, many, query.max_,
       query.arriveBy_ ? osr::direction::kBackward : osr::direction::kForward,
-      query.maxMatchingDistance_, nullptr, nullptr, elevations_);
+      query.maxMatchingDistance_, nullptr, nullptr, elevations_,
+      with_distance
+          ? std::function<bool(osr::path const&)>{[](osr::path const&) {
+              return true;
```
felixguendling: isn't this `return with_distance;`?

include/motis/endpoints/one_to_many.h
```diff
@@ -37,15 +37,34 @@ api::oneToMany_response one_to_many_handle_request(
...
-                         : api::Duration{};
+  return utl::to_vec(paths, [&](std::optional<osr::path> const& p) {
+    if (!p.has_value()) {
+      return api::Duration{};
+    }
+    auto d = api::Duration{.duration_ = p->cost_};
+    if (with_distance) {
+      d.distance_ = p->dist_;
+    }
+    return d;
```
felixguendling:
```suggestion
    return p.has_value() ? api::Duration{.duration_ = p->cost_, .distance_ = with_distance ? std::optional{p->dist_} : std::nullopt}
                         : api::Duration{};
```

(not formatted yet)

openapi.yaml
```diff
@@ -997,6 +997,17 @@ paths:
...
           schema:
             type: boolean
+        - name: distance
+          in: query
+          required: false
+          description: |
+            Optional. Default is `false`.
+            If true, the response includes the distance in meters
+            for each path. This requires path reconstruction and
+            may be slower than duration-only queries.
```
```suggestion
            is slower than duration-only queries.
```

can't be the same speed if it does more work? :)

include/motis/endpoints/one_to_many.h
e-kotov: It is supposed to be a guard against the code generator. `distance` is a non-required argument in the spec, so depending on the generator version it might come through as `bool` or `std::optional<bool>`.

include/motis/endpoints/one_to_many.h
e-kotov: replaced with formatted version

include/motis/endpoints/one_to_many.h
e-kotov: done

openapi.yaml
e-kotov: To be fair, the calculation of distances is not slowing down as much as you expected.

For 1000 destinations:

[onemany_queries_dist.txt](https://github.com/user-attachments/files/25328565/onemany_queries_dist.txt)
[onemany_queries_nodist.txt](https://github.com/user-attachments/files/25328566/onemany_queries_nodist.txt)

With distance:
```
response_time
=============

response_time
      average:        1,728.00
          max:        1,728
  99 quantile:        1,728
  90 quantile:        1,728
  80 quantile:        1,728
  50 quantile:        1,728
          min:        1,728
```

No distance:
```
response_time
=============

response_time
      average:        1,524.00
          max:        1,524
  99 quantile:        1,524
  90 quantile:        1,524
  80 quantile:        1,524
  50 quantile:        1,524
          min:        1,524
```

include/motis/endpoints/one_to_many.h
felixguendling: if the generator changes, we will adjust it then
for now please "keep it simple" ;)

openapi.yaml
felixguendling: looks good - I still think even for 10%, it can't hurt to have a flag to skip this computation

include/motis/endpoints/one_to_many.h
e-kotov: ok, done)

e-kotov: hmmm, seems like we will have to revert it back to how it was... The CI is failing now.

felixguendling:
Can you please check if my changes are ok and it still works?

e-kotov:
> Can you please check if my changes are ok and it still works?

Sure, I will test and report back here.

PR #1300, opened 2026-02-18
stop_times API: add window parameter

Kiarash-Parvizi:
On the topic of #1275


Kiarash-Parvizi:
@felixguendling could you please review this PR?

traines-source:
As I proposed in #1275 I think that when both params are given, we should interpret them as minimums, similar to how numItineraries and searchWindow work in the plan endpoint.
EDIT: for a typical station board in an app, IMO you would want to set this to e.g. n=5 and window=900, so that on a stop where there's lots of departures you at least get the next 15 minutes, and for a stop where there is a departure only every hour, you get the five next departures/hours.

include/motis/config.h
```diff
@@ -206,6 +206,7 @@ struct config {
   struct limits {
     bool operator==(limits const&) const = default;
     unsigned stoptimes_max_results_{256U};
+    unsigned stoptimes_max_window_{2U * 60U * 60U};
```
felixguendling: remove
we already have max results which should be respected anyway
max_window doesn't have any value on top of max results

src/endpoints/stop_times.cc
```diff
@@ -201,7 +201,8 @@ std::vector<n::rt::run> get_events(
     n::direction const dir,
     std::size_t const count,
     n::routing::clasz_mask_t const allowed_clasz,
-    bool const with_scheduled_skipped_stops) {
+    bool const with_scheduled_skipped_stops,
+    std::optional<n::duration_t> const max_time_diff = std::nullopt) {
```
felixguendling: - what's the reason for the default value nullopt?
- this should consume min_count + max_count instead of count

iterate at least until min_count is reached
throw exception if max_count is reached

PR #1310, opened 2026-02-22
Use CMake's built-in functionality to enable -Werror (or equivalent)

vkrause:
Avoids having to manage this per compiler here, and gives users an easy way to disable that passing --compile-no-warning-as-error to CMake.


felixguendling:
needs #1310 which will be merged asap

PR #1317, opened 2026-02-24
Add limits for `many` and `max` parameters in one-to-many requests

pialwis:
Following a discussion in #1280, this PR adds a new limit config parameter called `onetomany_max_many` that aims to limit the accepted number of `many` locations for one-to-many GET and POST requests. I also changed the `/api/v1/map/initial` endpoint so that it returns in the server config object that new limit with the `maxOneToManySize` attribute.

I chose a default value of 5 for the limit. What do you think? Also are the new variable names okay for you?

Finally this PR also fixes and improves some minor issues: I improved error handling for the one-to-many requests (use more precise status code errors) and I fixed some required/not required values in OpenAPI doc regarding these requests and `/api/v1/map/initial`.


traines-source:
Thanks, LGTM, maybe we should also limit the max param? Not sure what currently happens if that is set to something very large. We could reuse the limit `street_routing_max_direct_seconds` for that IMO.
Also, I think 5 as a limit is a bit low, maybe 128? That should still comfortably fit in the URL.

PR #1318, opened 2026-02-24
motis generate: adds option set the routing algorithm

mority:


felixguendling:
the approach from `motis params` (to use a URL to overwrite parameters) is probably the more generic tool (doesn't have to be extended)

PR #1322, opened 2026-02-27
Add pareto sets for One to Many

MichaelKutzner:
This follows #1285, adding pareto sets to the response, instead of the shortest, possible duration. Responses for direct street routing can now contain the distance as well.

This Pull Request is related to motis-project/nigiri#329, to avoid duplicated code.

@e-kotov Would this work for you? Or am I missing something important again?

### DIFF
--- a/include/motis/endpoints/one_to_many.h
+++ b/include/motis/endpoints/one_to_many.h
@@ -19,6 +19,7 @@
 #include "motis/data.h"
 #include "motis/fwd.h"
 #include "motis/match_platforms.h"
+#include "motis/metrics_registry.h"
 #include "motis/osr/parameters.h"
 #include "motis/parse_location.h"
 #include "motis/place.h"
@@ -48,7 +49,9 @@ api::oneToMany_response one_to_many_handle_request(
     Params const& query,
     osr::ways const& w,
     osr::lookup const& l,
-    osr::elevation_storage const* elevations) {
+    osr::elevation_storage const* elevations,
+    metrics_registry* metrics) {
+  metrics->one_to_many_requests_.Increment();
   // required field with default value, not std::optional
   static_assert(std::is_same_v<decltype(query.withDistance_), bool>);
 
@@ -85,6 +88,7 @@ struct one_to_many {
   osr::ways const& w_;
   osr::lookup const& l_;
   osr::elevation_storage const* elevations_;
+  metrics_registry* metrics_;
 };
 
 struct one_to_many_intermodal {
--- a/include/motis/endpoints/one_to_many_post.h
+++ b/include/motis/endpoints/one_to_many_post.h
@@ -21,6 +21,7 @@ struct one_to_many_post {
   osr::ways const& w_;
   osr::lookup const& l_;
   osr::elevation_storage const* elevations_;
+  metrics_registry* metrics_;
 };
 
 struct one_to_many_intermodal_post {
--- a/include/motis/metrics_registry.h
+++ b/include/motis/metrics_registry.h
@@ -13,6 +13,7 @@ struct metrics_registry {
   ~metrics_registry();
   prometheus::Registry registry_;
   prometheus::Counter& routing_requests_;
+  prometheus::Counter& one_to_many_requests_;
   prometheus::Counter& routing_journeys_found_;
   prometheus::Family<prometheus::Histogram>& routing_odm_journeys_found_;
   prometheus::Histogram& routing_odm_journeys_found_blacklist_;
--- a/src/endpoints/one_to_many.cc
+++ b/src/endpoints/one_to_many.cc
@@ -226,7 +226,8 @@ std::vector<api::ParetoSet> transit_durations(
 api::oneToMany_response one_to_many::operator()(
     boost::urls::url_view const& url) const {
   auto const query = api::oneToMany_params{url.params()};
-  return one_to_many_handle_request(config_, query, w_, l_, elevations_);
+  return one_to_many_handle_request(config_, query, w_, l_, elevations_,
+                                    metrics_);
 }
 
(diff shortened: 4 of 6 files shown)


MichaelKutzner:
Not sure if adding some tests to `test/endpoints/map_routes_test.cc` is reasonable in this case. Or if I should add these routes and trips to the existing test case instead. It's something, where reusable imports would be nice.

felixguendling:
Why exactly map_routes_test?

We could just initialize a global `unique_ptr<data>` in `test/main.cc`.

MichaelKutzner:
> Why exactly map_routes_test?
 
It was an already existing configuration, which would allow for more than one entry in the pareto sets.
But as already mentioned, I don't know, if I like that, as it seems really odd. And it wouldn't be much work to copy the trips to the existing test case. (And I could tune the times, without breaking any existing tests.) So I will probably fix this later.

e-kotov:
> adding pareto sets to the response, instead of the shortest, possible duration
Would this work for you? Or am I missing something important again?

Yes, this will works perfectly for me. The transition to Pareto sets is very useful for scientific purposes as it allows for much more accurate generalized cost modeling and accessibility analysis by considering the transfer count rather than just the single shortest duration. This approach provides the necessary data for mobility research without the heavy overhead of full path reconstruction.

MichaelKutzner:
Tests are moved now. Took a bit more effort than expected, as changing `max_footpath_length_` broke other tests. And I didn't want to run a new import. But it should work now.
It's still marked as draft, as I need to update `.pkg` first.

traines-source:
Proposition: In order to allow for backwards compatibility and have unified API endpoints one-to-many, one-to-all (and possibly at one point many-to-many) supporting all modes, couldn't we add the pareto sets under an additional attribute as a list?

felixguendling:
> and possibly at one point many-to-many

in this case, it would probably make sense to name them `from` and `to` from the beginning (and not `one` + `many`)

MichaelKutzner:
> Proposition: In order to allow for backwards compatibility and have unified API endpoints one-to-many, one-to-all (and possibly at one point many-to-many) supporting all modes, couldn't we add the pareto sets under an additional attribute as a list?

Using an additional attribute for the pareto sets makes sense, Especially as not all attributes set by `osr` routing can be filled for `nigiri` routing and vice verse. For example we currently cannot set a `distance` for `nigiri`.

About the implementation, did you think about something like I implemented now? Or something different?
I can think of these approaches:

* (as now): Keep `osr` and `nigiri` routed results separate, i.e. `pair<vector<Duration>, vector<ParetoSet>>`  
    This makes it easy to compute them independently. And it would be very easy to add an option to not compute `osr` durations, if needed. However, it requires users to zip both vectors. (should be easy)
* (similar to previously): Return both results at once, i.e. `vector<pair<Duration, ParsetoSet>>`  
    This doesn't require users to combine two vectors. And the implementation isn't that bad either. But it leads to larger responses (uncompressed), as the attribute names need to be repeated for each `many`, making it harder to read manually. I already attempted to make both attributes optional. But that resulted in two different ways to denote unreachable locations for street routing, which again makes it more difficult for users: `std::nullopt`, `Duration{}`  
    * (a variation of 2nd): Add optional pareto sets to `Duration`  
        This would prevent some issues mentioned for the second approach. But then it will no longer be possible to not compute `osr` durations, which might be useful when running multiple requests for different `time`s.



> > and possibly at one point many-to-many
> 
> in this case, it would probably make sense to name them `from` and `to` from the beginning (and not `one` + `many`)

I currently find it hard to imagine an OpenAPI specification, that would allow for either `from` or `to` be a single element, while the other is an array.
Maybe there is a solution using `oneOf`. But will make thinks much harder. And the feature is not supported by all generators.

traines-source:
If I'm not missing anything, currently it would not be backwards-compatible with the existing one-to-many endpoint at all?

I think we agree that currently the situation is very confusing: There is a one-to-many street-routing only endpoint (which according to the API docs appears to also support public transport modes for the mode parameter since the enum is reused...) and there is a one-to-all endpoint that just does PT routing. The new endpoint(s) would make it even more confusing.

My objective would have been to keep just these two and extend them in a backwards compatible way, i.e. merge the new experimental intermodal endpoint into the existing one-to-many endpoint, e.g. like that:
```
[
{
"duration": 0,
"distance": 0,
"pareto": [
     {
    "duration": 0,
    "distance": 0,
    "k": 1
     },
    ...
  ]
},
...
]
```

But that was just an idea, maybe it has other drawbacks.

MichaelKutzner:
> If I'm not missing anything, currently it would not be backwards-compatible with the existing one-to-many endpoint at all?
> 
> I think we agree that currently the situation is very confusing: There is a one-to-many street-routing only endpoint (which according to the API docs appears to also support public transport modes for the mode parameter since the enum is reused...) and there is a one-to-all endpoint that just does PT routing. The new endpoint(s) would make it even more confusing.
> 
> My objective would have been to keep just these two and extend them in a backwards compatible way, i.e. merge the new experimental intermodal endpoint into the existing one-to-many endpoint, e.g. like that:
> 
> ```
> [
> {
> "duration": 0,
> "distance": 0,
> "pareto": [
>      {
>     "duration": 0,
>     "distance": 0,
>     "k": 1
>      },
>     ...
>   ]
> },
> ...
> ]
> ```
> 
> But that was just an idea, maybe it has other drawbacks.

I agree, it gets quite confusing. Backwards compatibility would be nice. But I had to break enough already, that I wouldn't enforce it, if there is a better solution.
Also I don't know, what would work best in the long run.

MichaelKutzner:
I have added a new metric for One-to-Many now.

@felixguendling @e-kotov Any opinion about the response object? Is there still anything missing? Or what should be improved before it's done?

PR #1325, opened 2026-03-05
On-demand shapes debugging

pablohoch:
Adds on-demand debugging support for generated shapes:

1. Open the UI in debug mode (`?debug` URL parameter)
2. Enable the routes overlay (routes button in the top right corner)
3. Click on a shape
4. Click the "Debug" button for the desired route in the popup. This will download a .json.gz file with map matching debug data. Note that this will re-run the map matching for that route, so it can take a few seconds for long routes.
5. The .json.gz file can be dragged into the map matching debug UI (see `mm-debug-ui` in the osr project).

The debug API is disabeld by default but can be enabled by setting `timetable.route_shapes.debug_api = true` in the config.

Needs https://github.com/motis-project/osr/pull/96


src/endpoints/map/shapes_debug.cc
```diff
@@ -0,0 +1,127 @@
...
+  auto debug_json = route_shape_debug(*w_, *l_, *tt_, route);
+  debug_json["caller"] =
+      build_caller_data(*tt_, *tags_, route, route_idx, clasz);
+  auto payload = osr::gzip_json(debug_json);
+  auto const filename =
+      fmt::format("r_{}_{}.json.gz", route_idx, to_str(clasz));
+
+  auto res = net::web_server::string_res_t{boost::beast::http::status::ok,
+                                           req.version()};
+  res.insert(boost::beast::http::field::content_type, "application/gzip");
```
felixguendling, line 117: isn't this already done automatically if the client requests it?

include/motis/config.h
```diff
@@ -110,6 +111,7 @@ struct config {
       std::optional<std::map<std::string, bool>> clasz_{};
       unsigned max_stops_{0U};
       unsigned n_threads_{0U};
+      bool debug_api_{true};
```
felixguendling: is this a good option for production usage?

src/endpoints/map/shapes_debug.cc
pablohoch: We currently set the Content-Encoding header and gzip the response if it is a string and the client supports it.
This endpoint isn't supposed to be a normal API endpoint like the others (it isn't listed in openapi.yaml), but more like the tiles endpoint. The reason I'm returning a response with custom headers here is so that the browser downloads it to a file with the given name. Also if we'd set Content-Encoding the browser would decompress the file and save it as uncompressed JSON AFAIK (the uncompressed files are quite large).

include/motis/config.h
pablohoch: I've turned it off by default now. It's of course not required for production usage until you want to debug something. It doesn't hurt to turn it on, but it can be used to generate long-running requests. Then again, many other APIs (e.g. plan, or if we decide to add a map matching API later on) can do the same, so I thought we might as well turn it on by default as well. But off is probably a safer default.

PR #1332, opened 2026-03-12
Fix unstable One-to-Many tests

MichaelKutzner:
This implements a custom matcher, as the current tests are quite unstable (expects equals for `double`).

As it turned out, that the matcher requires a lot of additional code, I will check again, if there is an easier way to check that the correct distances are set.

### DIFF
--- a/test/endpoints/one_to_many_test.cc
+++ b/test/endpoints/one_to_many_test.cc
@@ -353,10 +353,8 @@ TEST(motis, one_to_many) {
     auto const& td = durations.transit_durations_.value();
 
     ASSERT_EQ(5U, sd.size());
-    EXPECT_FALSE(sd.at(0).duration_.has_value());
-    EXPECT_FALSE(sd.at(0).distance_.has_value());
-    EXPECT_FALSE(sd.at(1).duration_.has_value());
-    EXPECT_FALSE(sd.at(1).distance_.has_value());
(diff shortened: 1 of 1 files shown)


traines-source:
Isn't there `EXPECT_DOUBLE_EQ` and `EXPECT_FLOAT_EQ`?

MichaelKutzner:
> Isn't there `EXPECT_DOUBLE_EQ` and `EXPECT_FLOAT_EQ`?

There are. But these are intended to compare two `double` values directly. Not for structures, that contains a `double` somewhere.
https://google.github.io/googletest/reference/assertions.html#EXPECT_DOUBLE_EQ
Same for `EXPECT_NEAR`, which I used here (`::testing::DoubleNear`).

The issue here is, that GoogleTest's `EXPECT_EQ` relies on the implementation of `operator==`, which I don't see an option to override during the tests. And as it's a generated class, tinkering there to handle this specific use case seems unreasonable.
We could also round the distance in the result, reducing the likelihood of an error. But that felt like an arbitrary restriction, even though the result wouldn't be that exact anyway.

MichaelKutzner:
Thanks @felixguendling for suggesting a much simpler approach, that doesn't require writing a custom matcher.

As the exact distance is not that important here, it uses `EXPECT_NEAR` to check, that the distance is reasonable. If it turns out to be still too unstable, the maximum error can be increased further.

PR #1334, opened 2026-03-13
Exclude-Geometry: A field to control if transfer polylines and step instructions are returned

### LINKED ISSUE #1320: /plan, /trip: API param to not return legGeometry+steps for pre/post transit, transit and direct legs
Even more important to not waste bandwidth now that soon everything will have shapes.
We have `detailedTransfers` but that just turns legGeometry off for transfers. Not sure in which case one would like to have legGeometry for transit legs, but not for transfers 🤔 
There was some discussion around less heavy plan responses somewhere, haven't found it again. In any case, legGeometry and steps are probably the heaviest parts of responses.

Kiarash-Parvizi:
resolves #1320


openapi.yaml
```diff
@@ -4064,7 +4087,6 @@ components:
         - duration
         - from
         - to
-        - legGeometry
```
traines-source, line 4084: Have we had this case before? @felixguendling  Converting required response fields to optional ones when the value will always be present as long as the client doesn't use the newly introduced flag? – Meaning the API itself is backwards compatible, but generated clients are not, because the field is suddenly nullable (hence the UI error in our own client).
Not sure what is less ugly, converting it to optional or filling it with an empty polyline.

felixguendling: In general, we can break generated clients because a client generated for the previous version will still continue to work in this case (we will keep 100% API compatibility for the actual MOTIS API).
However, I would say that it's ok to have an empty `legGeometry` in order to ease the transition to new versions of the generated client (to also keep 100% compatibility of the generated client we publish).

PR #1341, opened 2026-03-18
railviz precision

traines-source:
oops


src/railviz.cc
```diff
@@ -373,49 +374,63 @@ api::trips_response get_trains(tag_lookup const& tags,
...
+  };
+
+  switch (static_cast<int>(query.precision_)) {
+    case 0: return make_response(std::integral_constant<int, 0>{});
+    case 1: return make_response(std::integral_constant<int, 1>{});
+    case 2: return make_response(std::integral_constant<int, 2>{});
+    case 3: return make_response(std::integral_constant<int, 3>{});
+    case 4: return make_response(std::integral_constant<int, 4>{});
+    case 5: return make_response(std::integral_constant<int, 5>{});
+    case 6: return make_response(std::integral_constant<int, 6>{});
```
felixguendling: this would be useful to have in geo

src/railviz.cc
```diff
@@ -7,6 +7,7 @@
 #include "cista/containers/rtree.h"
 #include "cista/reflection/comparable.h"
 
+#include <net/bad_request_exception.h>
```
felixguendling, line 10:
```suggestion
#include "net/bad_request_exception.h"
```

PR #1342, opened 2026-03-18
Refactor stop_times

Kiarash-Parvizi:
`get_stop_times_runs` will be needed in #1286

### DIFF
--- a/src/endpoints/stop_times.cc
+++ b/src/endpoints/stop_times.cc
@@ -351,13 +351,15 @@ std::vector<api::Place> other_stops_impl(n::rt::frun fr,
   }
 }
 
-api::stoptimes_response stop_times::operator()(
-    boost::urls::url_view const& url) const {
-  auto const query = api::stoptimes_params{url.params()};
-  auto const& lang = query.language_;
-  auto const api_version = get_api_version(url);
+std::vector<n::rt::run> get_stop_times_runs(
+    stop_times const& self,
+    api::stoptimes_params const& query,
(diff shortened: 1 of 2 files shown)


src/endpoints/stop_times.cc
```diff
@@ -472,16 +467,56 @@ api::stoptimes_response stop_times::operator()(
                              return to_tuple(a) == to_tuple(b);
                            }),
                end(events));
+  return events;
+}
+
+std::vector<nigiri::rt::run> get_stop_times_runs(
+    stop_times const& self, api::stoptimes_params const& query) {
```
felixguendling: this is not Python, C++ has member functions

src/endpoints/stop_times.cc
```diff
@@ -472,6 +463,41 @@ api::stoptimes_response stop_times::operator()(
                              return to_tuple(a) == to_tuple(b);
                            }),
                end(events));
+  return events;
+}
+
+std::vector<nigiri::rt::run> stop_times::get_runs(
+    api::stoptimes_params const& query) const {
+  auto const rt = std::atomic_load(&rt_);
```
felixguendling: bug

felixguendling:
I think a function like in `direct.cc` or an adaption of `get_earliest_alternative` in `pong.cc` is a better choice for the journey id because it finds routes operating at both stops very fast and then we only iterate those routes.

felixguendling (changes requested):

felixguendling:
not needed for itinerary-id anymore, closing

PR #1346, opened 2026-03-20
Add MOTIS version in server config response

pialwis:
The version of the running MOTIS server can be retrieved by looking at the `Server` HTTP header of any response. I find it not reliable as HTTP headers might be changed by reverse proxy configs or anything.

This PR adds in the server config object in the initial map API endpoint response the running version of the MOTIS server. My solution may not be the cleanest way to do it, but it introduces very few lines of codes and just works. What do you think?

### DIFF
--- a/src/endpoints/initial.cc
+++ b/src/endpoints/initial.cc
@@ -62,7 +62,7 @@ api::initial_response initial::operator()(boost::urls::url_view const&) const {
       .lon_ = center.lng_,
       .zoom_ = static_cast<double>(zoom),
       .serverConfig_ = api::ServerConfig{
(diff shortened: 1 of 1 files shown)


src/endpoints/initial.cc
```diff
@@ -62,6 +62,7 @@ api::initial_response initial::operator()(boost::urls::url_view const&) const {
       .lon_ = center.lng_,
       .zoom_ = static_cast<double>(zoom),
       .serverConfig_ = api::ServerConfig{
+          .motisVersion_ = static_cast<std::string>(motis_version_),
```
felixguendling:
```suggestion
          .motisVersion_ = std::string{motis_version_},
```

PR #1353, opened 2026-03-29
Feat/config geocode limits

flowride:
Being able to modify the number of results returned by geocode and reverse geocode allow a more precise sorting of data on the client side

### DIFF
--- a/include/motis/config.h
+++ b/include/motis/config.h
@@ -247,6 +247,8 @@ struct config {
     unsigned gtfsrt_expose_max_trip_updates_{100U};
     unsigned street_routing_max_prepost_transit_seconds_{3600U};
     unsigned street_routing_max_direct_seconds_{21600U};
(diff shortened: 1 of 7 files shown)


flowride:
need to clean

felixguendling:
If you would like to introduce new features in MOTIS it would be nice if you could open an issue before so we can discuss what's the best way to solve the problem. Also, it would make review easier if your PRs are small.

I think result quality of the geocoding should be improved in the server, so all clients and not just the debug UI benefits. The debug UI in MOTIS should display things as close as possible to the actual API result and not do any transformations. If we think that it's clearly better to reorder geocoding results (for whatever reason -  to be discussed in a issue here in MOTIS), this should be done in the `adr` library.

flowride:
Sorry I have isolated the small modification that I need to transmit in another PR

PR #1354, opened 2026-03-29
config: allow to configure geocode and reverse_geocode limits

flowride:
Add two new yaml configs to setup limits in geocode and reverse_geocode 

limits:
  stoptimes_max_results: 16384
  plan_max_results: 256
  plan_max_search_window_minutes: 5760
  stops_max_results: 2048
  onetoall_max_results: 524288
  onetoall_max_travel_minutes: 2880
  routing_max_timeout_seconds: 90
  **geocode_max_suggestions: 25
  reverse_geocode_max_results: 5**

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1829,7 +1829,7 @@ paths:
             For example, this can be used to allow only `ADDRESS` and `STOP` results.
           schema:
             $ref: '#/components/schemas/LocationType'
-        - name: limit
+        - name: numResults
           in: query
           required: false
           description: |
@@ -1923,7 +1923,7 @@ paths:
           schema:
             type: number
             default: 1
-        - name: limit
+        - name: numResults
           in: query
           required: false
           description: |
--- a/src/endpoints/adr/geocode.cc
+++ b/src/endpoints/adr/geocode.cc
@@ -83,7 +83,7 @@ api::geocode_response geocode::operator()(
               })
           .value_or(std::function<bool(adr::place_idx_t)>{});
   auto const config_limit = config_.get_limits().geocode_max_suggestions_;
-  auto const requested_limit = params.limit_.value_or(kDefaultSuggestions);
+  auto const requested_limit = params.numResults_.value_or(kDefaultSuggestions);
   utl::verify<net::bad_request_exception>(requested_limit >= 1,
                                           "limit must be >= 1");
   utl::verify<net::bad_request_exception>(
@@ -91,9 +91,8 @@ api::geocode_response geocode::operator()(
       "limit must be <= geocode_max_suggestions ({})", config_limit);
   auto const token_pos = a::get_suggestions<false>(
       t_, params.text_, static_cast<unsigned>(requested_limit), lang_indices,
-      ctx, place,
-      static_cast<float>(params.placeBias_), to_filter_type(params.type_),
-      place_filter);
+      ctx, place, static_cast<float>(params.placeBias_),
+      to_filter_type(params.type_), place_filter);
   return suggestions_to_response(t_, f_, ae_, tt_, tags_, w_, pl_, matches_,
                                  lang_indices, token_pos, ctx.suggestions_);
 }
--- a/src/endpoints/adr/reverse_geocode.cc
+++ b/src/endpoints/adr/reverse_geocode.cc
@@ -20,7 +20,7 @@ api::reverseGeocode_response reverse_geocode::operator()(
     boost::urls::url_view const& url) const {
   auto const params = api::reverseGeocode_params{url.params()};
   auto const config_limit = config_.get_limits().reverse_geocode_max_results_;
-  auto const requested_limit = params.limit_.value_or(kDefaultResults);
+  auto const requested_limit = params.numResults_.value_or(kDefaultResults);
   utl::verify<net::bad_request_exception>(requested_limit >= 1,
                                           "limit must be >= 1");
   utl::verify<net::bad_request_exception>(


src/config.cc
```diff
@@ -121,6 +121,12 @@ void config::verify() const {
                   nigiri::routing::kMaxSearchIntervalSize.count(),
               "plan_max_search_window_minutes limit cannot be above {}",
               nigiri::routing::kMaxSearchIntervalSize.count());
+  utl::verify(limits_.value().geocode_max_suggestions_ >= 1U &&
+                  limits_.value().geocode_max_suggestions_ <= 256U,
+              "geocode_max_suggestions must be between 1 and 256");
+  utl::verify(limits_.value().reverse_geocode_max_results_ >= 1U &&
+                  limits_.value().reverse_geocode_max_results_ <= 256U,
+              "reverse_geocode_max_results must be between 1 and 256");
```
felixguendling: why limit this? if the server operators wants to return > 256 results, that's fine IMO

traines-source:
much more useful than a server limit would be an API param so that clients can decide how many results they want? The server limit could IMO just be hardcoded to something reasonably high – but of course a configurable server limit can't harm

src/config.cc
flowride: yes you are right, can you integrate and remove the limit ? or do you need for me to make another PR ?

felixguendling: you can just update this PR

flowride: done is it ok ?

flowride:
> much more useful than a server limit would be an API param so that clients can decide how many results they want? The server limit could IMO just be hardcoded to something reasonably high – but of course a configurable server limit can't harm

Have much more implication but definitely a future PR for that

src/config.cc
flowride: Do you think you can integrate this modification ?

felixguendling: We should have the query parameter in this PR so it makes sense.

flowride: You mean both the default config and the http request parameter ?

felixguendling: Yes

felixguendling: The config should contain the upper limit and the request parameter should control the number of results.

flowride: is it ok for you like this ?

src/endpoints/adr/geocode.cc
```diff
@@ -79,10 +80,18 @@ api::geocode_response geocode::operator()(
                     }}};
               })
           .value_or(std::function<bool(adr::place_idx_t)>{});
-  auto const token_pos =
-      a::get_suggestions<false>(t_, params.text_, 10U, lang_indices, ctx, place,
-                                static_cast<float>(params.placeBias_),
-                                to_filter_type(params.type_), place_filter);
+  auto const config_limit = config_.get_limits().geocode_max_suggestions_;
+  auto const requested_limit = params.limit_.value_or(config_limit);
```
felixguendling: the default limit should stay the same (10, always) no matter what the config says

src/config.cc
flowride: Hope this is ok for you now, don't hesitate if you have any other question

src/endpoints/adr/geocode.cc
```diff
@@ -79,10 +82,18 @@ api::geocode_response geocode::operator()(
...
-                                static_cast<float>(params.placeBias_),
-                                to_filter_type(params.type_), place_filter);
+  auto const config_limit = config_.get_limits().geocode_max_suggestions_;
+  auto const requested_limit = params.limit_.value_or(kDefaultSuggestions);
+  utl::verify<net::bad_request_exception>(requested_limit >= 1,
+                                          "limit must be >= 1");
+  utl::verify<net::bad_request_exception>(
+      requested_limit <= config_limit,
+      "limit must be <= geocode_max_suggestions ({})", config_limit);
+  auto const result_limit = requested_limit;
```
felixguendling: what is this variable doing? can be removed?

flowride: The variable is removed

openapi.yaml
```diff
@@ -1829,6 +1829,17 @@ paths:
             For example, this can be used to allow only `ADDRESS` and `STOP` results.
           schema:
             $ref: '#/components/schemas/LocationType'
+        - name: limit
```
felixguendling: I think until now we already have `numItineraries` and `n` as parameter names. I understand that `limit` might also be a good name (since it's common, for example in SQL) but I would rather just call it `n` in order to try to be at least a little bit consistent.

felixguendling:
CI compiler error

```
/__w/motis/motis/src/endpoints/adr/reverse_geocode.cc:32:60: error: implicit conversion changes signedness: 'const value_type' (aka 'const long') to 'std::size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
   32 |       r_.lookup(t_, parse_location((params.place_))->pos_, result_limit,
      |          ~~~~~~                                            ^~~~~~~~~~~~
1 error generated.

/__w/motis/motis/src/endpoints/adr/geocode.cc:93:25: error: implicit conversion loses integer precision: 'const value_type' (aka 'const long') to 'unsigned int' [-Werror,-Wshorten-64-to-32]
   92 |   auto const token_pos = a::get_suggestions<false>(
      |                          ~
   93 |       t_, params.text_, requested_limit, lang_indices, ctx, place,
      |                         ^~~~~~~~~~~~~~~
1 error generated.
```

openapi.yaml
flowride: n seems a little short in that context may be numResults 
I fixed the compilation issue by the way

felixguendling: ok, then let's use `numResults`

PR #1355, opened 2026-03-29
Routing without realtime data

Yams709:


Lach-anonym:
I see it as a fallback function in case RT data is gibberish, especially helpful for issues with blocked trips or too early Flixbuses

felixguendling:
> I see it as a fallback function in case RT data is gibberish, especially helpful for issues with blocked trips or too early Flixbuses

This won't work with this PR unless it's a parameter you can set from outside.

traines-source:
The idea was to evolve this further by
* supporting a hybrid mode where both realtime and non-realtime calculations are done and they are merged in one `pareto_set` (for that, the lambda is useful IMO) (I'm not 100% sure whether `pareto_set::add` or `pareto_set::add_not_optimal` make more sense, since we do not want duplicates, but are probably interested in non-optimal entries? Maybe we also need multiple modes for that, see below)
* exposing this to the API via an enum e.g. rtMode=ONLY (only rt, current default), OFF (only scheduled), HYBRID (see above) and possibly later more variants

felixguendling:
OJP additionally supports annotation only (routing without RT and updating the journey later).

src/endpoints/routing.cc
```diff
@@ -937,54 +937,71 @@ api::plan_response routing::operator()(boost::urls::url_view const& url) const {
...
+            routing(false);
+            for (auto j : r.journeys_->els_) {
+                j_tmp.add(std::move(j));
+            }
+
+            routing(true);
+            for (auto j : r.journeys_->els_) {
+                j_tmp.add(std::move(j));
+            }
+            r.journeys_ = &j_tmp;
```
felixguendling: dangling pointer

PR #1370, opened 2026-04-08
Create test cases for reusable tests

MichaelKutzner:
This implements reusable test cases, that can be used for multiple, potentially unrelated, tests.

The benefit of this change would be the possibility for more fine-grained tests, without duplicating setups and repeating imports. The implementation also prevents unnecessary imports, if only a subset is executed, e.g. when using the  `--gtest_filter=` option.

As there are currently no tests, that use exactly the same data and configuration, I decided to use the same name for test and test case (including some prefix). However, once test cases get reused, it's probably better to use a more descriptive name, if possible. But I'm currently missing some good ideas. Therefore I'd prefer changing the names in the future, when usage is more clear. Maybe some tests can be updated by then to use the same test case, reducing the required work on imports as well.


felixguendling:
Moving test cases away from where they are used decreases readability. I think it would be nice to leave the test case in the same file where possible. Also not all test cases are used in multiple tests. So for those that are used only once, we don't need to make it more complicated than necessary. It's sufficient to focus on those that are used in multiple tests but even then leave them where they are for now.

MichaelKutzner:
Thank you for the feedback. The decreased readability is something I considered as well.
As already mentioned, there are currently no reused test cases, as each differs somehow (some can probably be changed without much afford though). I started splitting tests for One-to-Many, but it still feels more like an example for now.
Moving the data back wouldn't be an issue. But then again it becomes more difficult to find a suitable, already existing test case, if something new is developed (not by much though). And maybe just having the framework available might be enough to avoid some duplicates in the future.

felixguendling:
> And maybe just having the framework available might be enough to avoid some duplicates in the future.

Then maybe it's easier to just add the framework for now. It's probably easier to start from scatch (master) and only add the framework.

MichaelKutzner:
That's fine for me. And also how I would have approached this.

MichaelKutzner:
I restored the `master`s state for most test cases. But I decided to keep at least the test case for One-to-Many as a reference. Splitting these tests also makes it a bit easier to fix a particular issue.

PR #1372, opened 2026-04-10
add aerialway to style

Findus23:
See https://github.com/motis-project/tiles/pull/16, inspired by #1359 

Before:
image
After:
image

The dash-style was picked kind of arbitrary, as I don't think there is a way to make orthogonal lines as a style.

Maybe https://docs.osmand.net/docs/user/map-legend/osmand/#aerialways would be an alternative in case this is not clear as aerialways.

Aerialway routing feels way out of scope though... :)


felixguendling:
For me I would not know what those lines are, and also the Osmand version doesn't tell me much more except the direction. But it's probably better than to display nothing.


> Aerialway routing 

There are a lot in Switzerland's GTFS feed so routing should already work somehow.

Findus23:
> There are a lot in Switzerland's GTFS feed so routing should already work somehow.

Oh wow. I assumed from a quick glance that getting even rough schedules for gondolas in Austria would be completely impossible. 
But indeed looking through Switzerland and in the Gondolas dataset on https://gtfs.geops.ch/ routing seems to be working completely fine.

> For me I would not know what those lines are, and also the Osmand version doesn't tell me much more except the direction. But it's probably better than to display nothing.

Indeed I can't think of a more intuitive display. But I guess this is still a lot better than having the markers hover over nothing.

felixguendling:
I merged the PR in tiles. Can you please update the reference in `.pkg` and run cmake or `pkg -l` so `.pkg.lock` updates as well and then update this PR?

Findus23:
@felixguendling I rebased to the latest master branch and updated pkg, so unless I made a mistake this should be ready to merge.

felixguendling:
Thanks!

PR #1374, opened 2026-04-15
ui: don't initialize state tracking vars from props

KristjanESPERANTO:
This PR simplifies tracking of the last applied map layer values.

In Layer and ZoneLayer, those tracking variables were set from props too early. That made the first comparison less reliable. They now start as undefined and are set during the first real update.

As part of that cleanup, the `@ts-ignore` and `$state.snapshot()` workarounds in `Layer` are not needed anymore and have been removed.

No user-facing behavior is intended to change. This is a small robustness cleanup for more predictable internal updates.

There may be a few more similar cleanup opportunities in this area. But I am still new here, so I wanted to keep PRs intentionally small and easy to review.

### DIFF
--- a/ui/src/lib/map/Layer.svelte
+++ b/ui/src/lib/map/Layer.svelte
@@ -141,12 +141,10 @@
 	let source: { id: string | null } = getContext('source'); // from GeoJSON component
 
 	let initialized = false;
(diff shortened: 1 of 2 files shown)


felixguendling:
I'll deploy it on test.motis-project.org for testing.

PR #1376, opened 2026-04-16
ui: fix UI test setup and warnings

KristjanESPERANTO:
When I tried running `pnpm run test` in the UI, it just didn't work - the scripts were referencing `npm` (but `packageManager` is `pnpm` in the `package.json`) and `vitest` was starting in watch mode instead of doing a single run. I fixed those so the test command actually works with the pnpm setup.

After that, the tests did run but came back with a bunch of Svelte 5 warnings about non-reactive prop reads, incorrect type imports, and one test still checking for the old SvelteKit boilerplate title. I cleaned all of that up.

Two warnings remain. They're not critical and fixing them properly would probably touch quite a bit more code - I'd rather do that separately.


felixguendling (approved):

felixguendling:
same, I'll deploy it on test.motis-project.org for testing
thank you!

felixguendling:
I had to revert it for now because the debug UI didn't work anymore. It should look like this:

Screenshot 2026-04-17 at 22 21 07

KristjanESPERANTO:
Okay, sorry. I'll have a look what went wrong.

PR #1379, opened 2026-04-17
api: Wheelchair Accessibility

jbruechert:
depends on https://github.com/motis-project/nigiri/pull/348


jbruechert:
hmm, the changed test results vaguely make sense judging from the comments, but I'm not sure why they changed at all.

jbruechert:
Disabling trip wheelchair filter restores the previous test results

jbruechert:
should work properly now

felixguendling:
nigiri PR is merged -> you can reference nigiri master here

jbruechert:
> nigiri PR is merged -> you can reference nigiri master here

done

openapi.yaml
```diff
@@ -4410,6 +4410,10 @@ components:
           description: |
             Whether bikes can be carried on this leg.
           type: boolean
+        wheelchairAccessible:
```
felixguendling, line 4419: I think we should make it future-proof by using an enum like for in/out allowed.
SKI+ uses those values:
https://github.com/openTdataCH/ojp-demo-app-src/blob/291549102a1775d6787cf4745bfbe2ceae8a0855/src/app/shared/models/stop-point-call.ts#L67-L87

In other places, other policies might be used, which we probably also want to be able to support.

jbruechert: I didn't use the same values because this applies primarily to the vehicle and not the platform, but I switched this to an enum that can be extended.
We might want to add a separate enum to places later on though.

openapi.yaml
```diff
@@ -3805,6 +3805,12 @@ components:
           type: integer
           description: decline in meters across this path segment
 
+    WheelchairAccessibility:
+      type: string
+      enum:
+        - WHEELCHAIR_SPACE_AVAILABLE
```
felixguendling: this sounds like we checked that it's actually not booked, which is not the case
maybe something more generic like just `ACCESSIBLE`?

PR #1382, opened 2026-04-22
Improve matching when no level is set

MichaelKutzner:
This improves routing results, when no level is specified.

Currently routing results can differ, if no level is set. For example the `/one-to-many` endpoints will allow matching to any nearby way, independent of their level. For the `/plan` endpoint however, only ways with level `0` are considered. This will often lead to different results.
This change will fix that behavior, as now no level restrictions will be applied for any endpoint.

About the implementation and backward compatibility:
I considered two possible solutions.

1. Change the code to show the same results for all versions. For `v1` to `v5` the level will always be set.
    This is the most consistent implementation, as the `level` field is only set after the results are computed. But as a consequence, the patch will now lead to different results for `v1` to `v5`.
1. Keep the same results for `v1` to `v5` and apply changes to the new versions only.
    For me this seemed slightly more complicated, as this requires slightly more changes. It also could be surprising, if upgrading to a new version leads to different results.

This Pull Requests implements the first approach.
However, if there are good reasons to go for the second approach, I can still change it.


MichaelKutzner:
Please check, if that change is fine for you. Or if it would break something you rely on, so the second approach would be preferred (see above for details).
@traines-source @jbruechert @vkrause

MichaelKutzner:
As there are no comments or objections until now, I assume, that first proposed approach (same results for all API versions) is fine for everyone.

src/place.cc
```diff
@@ -214,4 +220,19 @@ place_t get_place(n::timetable const* tt,
   return tt_location{tags->get_location(*tt, input)};
 }
 
+api::Place bwc_adjust(api::Place&& pl, unsigned const api_version) {
+  if (!pl.level_.has_value() && 0 < api_version && api_version < 6) {
```
felixguendling, line 224: why handle api_version=0 differently?

src/place.cc
```diff
@@ -57,7 +58,12 @@ double get_level(osr::ways const* w,
 }
 
 osr::location get_location(api::Place const& p) {
-  return {{p.lat_, p.lon_}, osr::level_t{static_cast<float>(p.level_)}};
+  return {{p.lat_, p.lon_},
+          p.level_
+              .transform([](auto const lvl) {
+                return osr::level_t{static_cast<float>(lvl)};
+              })
+              .value_or(osr::level_t{osr::kNoLevel})};
```
felixguendling: kNoLevel is already of type `level_t`?

openapi.yaml
```diff
@@ -3428,7 +3428,12 @@ components:
           description: longitude
           type: number
         level:
-          description: level according to OpenStreetMap
+          description: |
+            level according to OpenStreetMap
+            If no level is given, the field will be unset.
+
+            For older versions (v1-v5), this field is mandatory and therefore set to 0.
+            Affected endpoints: plan,trip,stoptimes,one-to-all,map/stops,map/trips
```
```suggestion
            Affected endpoints: plan, trip, stoptimes, one-to-all, map/stops, map/trips
```

include/motis/place.h
```diff
@@ -80,4 +82,9 @@ place_t get_place(nigiri::timetable const*,
                   tag_lookup const*,
                   std::string_view user_input);
 
+// backward compatibility function for place
```
felixguendling:
```suggestion
```

comment doesn't say anything what it really does -> remove

src/place.cc
MichaelKutzner: Is there a real usage for `api_version == 0`?
As `get_api_version()` will return `0`, if no API version is contained in the URL (mostly relevant for tests, I think?), I would assume, that we would like to use the most recent behavior for these as well.
https://github.com/motis-project/motis/blob/0bbaf30e08ad1037f4ff3ca0a9fb5473fb857d98/src/server.cc#L81-L87
I thought, that all other cases had `api_version==`, but I missed cases with `>` or `>=`. So I'm fine with whatever seems better.

src/place.cc
MichaelKutzner: It is. Fixed it.

src/place.cc
felixguendling: Tests should could also be explicit about which version they use? Having production code with behavior that's only intended for tests is kind of an anti-pattern.

include/motis/place.h
MichaelKutzner: The comment was primary to explain the name, as `bwc_` doesn't seem intuitive. At least for me. And naming the function `backward_compatibility_adjust` seemed a bit long.

felixguendling: that would mean I can't understand code without jumping back and forth to the comment?

then I think it would make more sense to call the function in a way that it's easy to understand - something like `bwd_compat_lvl_adjust`?

src/place.cc
MichaelKutzner: I agree, if it's code, that's only relevant for tests.
But given that out of 49 tests, only 1 sets an explicit version (and this test is part of this change), I assumed, that tests should usually use the latest API version? (If we set the version now, we need to update them later, every time the version changes.)
So maybe it would be better to change `get_api_version` to either always return the newest version or simply `std::numeric_limits<unsigned>::max()`, when the version is missing?

include/motis/place.h
MichaelKutzner: That name sounds good to me. I've updated the code.

src/place.cc
felixguendling: > that tests should usually use the latest API version?

So users using an older API version use untested code that might crash?

I think ideally we would test all supported API versions?

It doesn't help if only the latest version works if a request to older versions either crash the server or lead to wrong results.

A supported API version means ideally a tested API version.
If we don't test it, we should declare it unsupported.

MichaelKutzner: Good point. I assumed, that the code for older versions usually should not change. But if we don't test it, we cannot make any assumptions.
So ideally we need some method to run parameterize tests? To make updates not too complicated, I would like a centralized range of supported versions. And if some tests only work for a subset of that range, it shouldn't be too difficult to skip these cases, I think.
Not sure how this can be implemented though, yet. A quick search hints, that [`TEST_P`](https://google.github.io/googletest/reference/testing.html#TEST_P) might be an option?

felixguendling: Not sure if TEST_P is the simplest solution.
I think a simple range-based for loop will do the job.

MichaelKutzner: The setup certainly requires some overhead, especially if we don't want to import data more than once. But I think, that this might still be the easier part.
I have changed one test as an example. The overhead for using `TEST_P` seems very simple, as I didn't optimize anything at all. So the import currently needs to run for each parameter (== API version).
https://github.com/motis-project/motis/compare/master...MichaelKutzner:api_version_tests
Updating the assertions, however, feels more brittle. We either need a different assertion, each time something output related changes. Or create the expected output, depending on the API version, which I would like to avoid. This also makes reading the code more challenging.
For comparison, I also changed a different test to use a simple range based loop. And also the code looks simple, I cannot see for which API version an error occurred. That would be much easier when using something like `TEST_P`.

Example output for parameterized tests:
```
[==========] Running 5 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 5 tests from motis_endpoint
[ RUN      ] motis_endpoint.routing/APIVersion1
...
[       OK ] motis_endpoint.routing/APIVersion1 (3215 ms)
[ RUN      ] motis_endpoint.routing/APIVersion2
...
[       OK ] motis_endpoint.routing/APIVersion2 (3004 ms)
...
```

So I'm not sure what would be the best solution. We also could consider different test (files) for each version.

MichaelKutzner: Quick update on identifying the correct loop (`api_version`) during testing:
I just noticed, that this can be done using [`SCOPED_TRACE`](https://google.github.io/googletest/reference/testing.html#SCOPED_TRACE). I have updated the branch (previous message) to use it as well.
I think, then using a simple loop can be sufficient.

MichaelKutzner: As the Pull Request has been merged and adding support to test all API versions is out of scope of this feature, the discussion will continue at #1419

PR #1384, opened 2026-04-23
Fix healthep seg fault

aff3npirat:


aff3npirat:
So there was a problem with sanitizer PR check in `run_gbfs_update` - which I am having difficulties fixing. I would propose we remove gbfs/rt updates from `health_test.cc` and I will add seperate tests for them - since in essence health endpoint only relies on setting the respective metrics. What do you think? @felixguendling

felixguendling:
> So there was a problem with sanitizer PR check in run_gbfs_update

CI seems to be green. Where is the problem?
We should fix this before merge.

aff3npirat:
@felixguendling 

> CI seems to be green. Where is the problem?

The Problem was in [cd3092f](https://github.com/motis-project/motis/pull/1384/commits/cd3092f60aefb17705f06d4526addf508e555a16). It occurs in `run_gbfs_update`. Im pretty sure it is caused by uninitialized config, but could not pin point it yet - so I dont know if allowed configs (outside of tests) could cause it.

> We should fix this before merge.

Health ep logic only relies on setting specific metrics, not really on update methods. So I would change `health_test.cc` as in [90ea035](https://github.com/motis-project/motis/pull/1384/commits/90ea03587d006630811dfb801b2c69b6db862730) (ready for merge) and add new test cases that check `run_gbfs_update` for different configs/scenarios.

src/metrics_registry.cc
```diff
@@ -127,7 +128,14 @@ metrics_registry::metrics_registry(
           prometheus::BuildGauge()
               .Name("nigiri_timetable_transports_x_days_count")
               .Help("The number of transports x service days in the timetable")
-              .Register(registry_)} {}
+              .Register(registry_)},
+
```
felixguendling:
```suggestion
```

PR #1386, opened 2026-04-24
Add /stop endpoint

Jouca:
This PR proposes a new endpoint `/v6/stop`, designed to retrieve complete information about a stop including its place details and all routes serving it. This replaces the previous `/v1/stop/routes` proposal with a richer, unified endpoint that avoids relying on `/stoptimes` or other endpoints to reconstruct stop-level data.

It follows the review discussion suggesting that rather than a narrow routes-only endpoint, a proper `/stop` endpoint should return the full stop context in one call.

Location grouping uses the shared `collect_locations` utility for consistency with other endpoints: a stop is expanded to its parent root, its equivalences (same name), and their child stops, with optional radius-based search and ADR/address-engine integration for cross-feed equivalences.

### Parameters for `/v6/stop`

| Name | Type | Required | Default | Description |
|------|------|----------|---------|-------------|
| `stopId` | `string` | no¹ | - | Stop ID as returned by `/api/v1/geocode` or `/api/v1/map/stops` |
| `center` | `string` | no¹ | - | `latitude,longitude` pair used as center for a radius search |
| `radius` | `integer` | no | - | Search radius in meters. Required if `stopId` is not given; if provided with `stopId`, expands the result to nearby stops |
| `exactRadius` | `boolean` | no | `false` | If `true`, only physically-in-radius stations are included; if `false`, equivalences with the same name and their children are also considered |
| `language` | `string[]` | no | - | Language preference for translated names (BCP-47, e.g. `en`, `fr`, `de`) |

¹ Either `stopId` or (`center` + `radius`) must be provided.

### Response

The response is a JSON object with two fields:
- `place` : full stop information (name, coordinates, modes, timezone, parent station, stop code, etc.)
- `routes` : list of routes serving the stop, sorted by `routeShortName` then `routeId`, each with agency, color, mode, and route type

### Examples

```http
GET /api/v6/stop?stopId=IDFM-gtfs_IDFM:71264
GET /api/v6/stop?stopId=IDFM-gtfs_IDFM:71264&language=fr
GET /api/v6/stop?center=48.8583,2.3488&radius=200
GET /api/v6/stop?stopId=IDFM-gtfs_IDFM:71264&radius=300
```


Jouca:
Fixed formatting & errors, ready for a workflow run

Jouca:
Should be fixed again, ready for a workflow run

src/endpoints/stop_routes.cc
```diff
@@ -0,0 +1,119 @@
...
+      if (range.empty()) {
+        continue;
+      }
+
+      auto const t_idx = range.from_;
+      auto const fr = n::rt::frun{
+          tt_, nullptr,
+          n::rt::run{.t_ = n::transport{t_idx, first_day},
+                     .stop_range_ = {n::stop_idx_t{0}, n::stop_idx_t{1}}}};
+      auto const rs = fr[0];
```
felixguendling: just picking the first one isn't guaranteed to yield all route_id_idx

- route_id_ix => route from timetable
- route_idx_t => route for RAPTOR algorithm (grouped by stop sequence)

you might have:
- the same stop sequence with a different route_id (missing route in this code)
- same route_id with different stop sequence (duplicate route in this code)

both cases should be handled properly here

Jouca:
Hello @felixguendling,
These cases are now took correctly, you can take a look here: https://github.com/motis-project/motis/commit/0afce55750c7cb04e09f06863c16104a8d94574d#diff-9f1033ef7611cd5bb5eca1866ac3b9c1bdc3bf10db8b4ed9c2573671a7c4e0c2

src/endpoints/stop_routes.cc
```diff
@@ -0,0 +1,135 @@
...
+        auto const trip = tt_.merged_trips_[sections.front()].front();
+        auto const rid_idx = tt_.trip_route_id_[trip];
+        if (!seen_rid.emplace(rid_idx).second) {
+          continue;
+        }
+
+        auto const fr = n::rt::frun{
+            tt_, nullptr,
+            n::rt::run{.t_ = n::transport{t_idx, first_day},
+                       .stop_range_ = {n::stop_idx_t{0}, n::stop_idx_t{1}}}};
```
felixguendling: checking at stop_range=0 gives the route_id at the first stop which might differ from the route_id when the trip reaches this stop (different trips with same block_id will end up in the same transport_idx_t and therefore the route along a trip can change)

we should probably also have test cases for everything you fix now

Jouca:
Fixed by iterating all transports in the route range and using the actual stop position of the queried location instead of always stop 0. For the last stop, kArr is used instead of kDep to avoid an out-of-bounds section access on block-merged transports. 

I also added `test/endpoints/stop_routes_test.cc` covering: 
- same stop sequence with different route_ids
- deduplication of the same route_id across multiple stop sequences
- block-merged transports (first/last stop).

Jouca:
Fixed formatting

src/endpoints/stop_routes.cc
```diff
@@ -0,0 +1,144 @@
...
+  auto const query = api::stopRoutes_params{url.params()};
+  auto const& lang = query.language_;
+
+  utl::verify<net::bad_request_exception>(
+      query.stopId_.has_value() || query.center_.has_value(),
+      "either stopId or center must be provided");
+
+  auto locations = std::vector<n::location_idx_t>{};
+
+  auto const add_with_children = [&](n::location_idx_t const l) {
```
felixguendling: I think we should have 100% consistency with how the stop times endpoint handles grouping.

I think the best way to enforce consistency is if both endpoints would use the same utility. This can be derived from the stop times endpoint where already the place_idx grouping is used when available.

https://github.com/motis-project/motis/blob/b5f4a38629509b1b3d467f8212eba85de5e8df0a/src/endpoints/stop_times.cc#L452-L481

Jouca:
Ok normally this should respond to last review, I extracted the grouping logic into a shared utility (`include/motis/collect_locations.h`) used by both endpoints. 

Also added the missing `adr::typeahead const* t_` to stop_routes and aligned the stop-query path to match stop_times.

@traines-source Could you take a look on this review?

openapi.yaml
```diff
@@ -2843,6 +2843,82 @@ paths:
...
+                $ref: '#/components/schemas/Error'
+        '200':
+          description: array of routes/lines serving the stop(s)
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Route'
+
```
traines-source: I discourage the use of an array as the top level element, because if ever we want to add some metadata to the response, we can't do so without a breaking change -> should be an object containing an array, cf. stoptimes

src/endpoints/stop_routes.cc
```diff
@@ -0,0 +1,127 @@
...
+
+          auto const route_id = tags_.route_id(rs, ev);
+          if (!seen.emplace(route_id).second) {
+            continue;
+          }
+
+          auto const& agency = rs.get_provider(ev);
+          auto const color = rs.get_route_color(ev);
+
+          result.push_back(
```
traines-source: shouldn't this be sorted somehow? or is the default order somehow sensible? I would think it's completely random

traines-source (commented):
As discussed vaguely in #1427 thinking about it I tend to not wanting to introduce a new endpoint for that. Instead, we could just add a parameter and new field to the stoptimes endpoint, because 1. the API is very similar anyways 2. it will actually reduce duplicated code 3. a common use case will be to display all routes above the departure board 4. it saves us from thinking about the inconsistencies between REST-style and RPC-style URLs (stoptimes vs stop/routes).
Then in a next step we could also build the same into /map/stops.
What do you think?

mlundblad:
> As discussed vaguely in #1427 thinking about it I tend to not wanting to introduce a new endpoint for that. Instead, we could just add a parameter and new field to the stoptimes endpoint, because 1. the API is very similar anyways 2. it will actually reduce duplicated code 3. a common use case will be to display all routes above the departure board 4. it saves us from thinking about the inconsistencies between REST-style and RPC-style URLs (stoptimes vs stop/routes). Then in a next step we could also build the same into /map/stops. What do you think?

Yeah, I think integrating it into the stoptimes response makes sense.

I'm currently implementing stoptimes support in GNOME Maps, so it would make sense to also be able to "collect" the routes in the (first) call to stoptimes for a place, and then show the lines (we would probably group them by transport mode if there's multiple modes in a place). Rather than having to make another call and update the information asyncly afterwards.

Jouca:
From what we've seen together with @traines-source, the best outcome for this PR would be to focus exactly on current endpoints (such as `/geocode`, `/stoptimes`, `/map/stops`, ...), but on this specific case we will create a `/stop` endpoint and providing routes inside it instead of creating a ungeneric `/stop/routes` endpoint.

@mlundblad After I'm finishing creating the /stop endpoint, I will focus on your case about /stoptimes and working on other endpoints to provide routes information

src/endpoints/stop.cc
```diff
@@ -121,6 +129,92 @@ api::stopRoutes_response stop_routes::operator()(
...
+    if (parent != n::location_idx_t::invalid()) {
+      for (auto const a : al.location_[parent]) {
+        add_alert(a);
+      }
+    }
+
+    if (!alerts.empty()) {
+      result.place_.alerts_ = std::move(alerts);
+    }
+  }
```
traines-source, line 141: is it complicated to deduplicate this code with stoptimes?

traines-source (commented):
* We stick to "routes" vs "lines"? Maybe better since we already have a /map/routes (for the shapes) which is a similar concept
* I think this should be under /v6/ since it uses `routeShortName` and `mode` in a way compatible to >=v5 and we somehow wanted to introduce the policy that all endpoints have the same version number (even though that is not yet the case).
* while I think that location grouping in the same way as stoptimes would be useful, maybe we can leave this to the stoptimes endpoint. Just note that for the routes on Transitous this endpoint, which only takes the actual location id and its parent (?), may be of limited usefulness since 1) you usually want to get routes for all related stops from all feeds 2) stopIds change so it's useful if you can use center as a fallback to get the routes 3) you might on the other hand really only want to get routes for a single platform/bay = location (possible with `radius=0` `exactRadius=true` over in stoptimes)

other than that, LGTM

Jouca:
Thanks for the feedback @traines-source! 👍

1. Agreed, keeping `routes` is consistent with `/map/routes` and the existing `Route` type in the API.
2. Done, moved to `/v6/stop`.
3. Good points. I've simplified the location grouping to stop + direct children only (no ADR cross-feed expansion), so it won't pull in routes from unrelated stops across feeds. I've also added `center` + `radius` back as a fallback for when `stopId` is not provided, which should help with the stopId stability concern. For the single platform/bay use case, agreed that `/stoptimes` with `exactRadius=true` remains the right tool for that.

openapi.yaml
```diff
@@ -2809,6 +2809,14 @@ paths:
...
           explode: false
 
+        - name: withAlerts
+          in: query
+          required: false
+          schema:
+            type: boolean
+            default: true
+          description: include active alerts for the stop
+
```
traines-source: is there an implementation for this?

traines-source (commented):
Regarding location grouping, maybe we slightly misunderstood eachother, and I might have misunderstood your previous code, sorry. I meant that from my POV either we just have a very simple root-child location grouping and no other params (radius, center) or we have the exact same behavior as the stoptimes endpoint with adr grouping, including all parameters (radius, exactRadius, center). 

IMO for a train station you want to get both trains and local buses departing there, if there is an obviously associated bus stop. With the current state, on Transitous (and already in France with its hundreds of feeds) you will not be able to obtain that other than with a guessed radius flag. That's what the adr grouping is for and why IMO it may also be useful here.

But if you think the current state is a good compromise, I guess we can also leave it like that and still change it later. LMK :)

openapi.yaml
Jouca: Actually no, when checking back we don't need to keep this since most of the alerts are filtered with lines mostly

Jouca:
No worries, thanks for clarifying!

I went with full parity with `/stoptimes`: same `add_location` (ADR grouping) plus all the same parameters `radius`, `exactRadius`, and `center` as a fallback when `stopId` isn't provided or doesn't resolve. The query logic now mirrors `stop_times::get_runs` exactly (same root + radius/ADR expansion semantics), so a train station's associated bus stop from another feed should now show up, addressing the Transitous use case you mentioned.

I also removed `withAlerts` parameter since we don't need it anymore

traines-source:
Regarding `withAlerts` I was just referring to the map/stops endpoint, I think for the stop endpoint they could be useful. But that can still be added back again later.
Thanks for your contribution :) I have deployed it to https://test.motis-project.org/ if you want to have a look how it will behave on Transitous.

PR #1387, opened 2026-04-25
docs(setup): optimize local dev setup docs

KristjanESPERANTO:
I kept running into the same issues while trying to get the local dev setup running, and to be fair, part of that was me not reading the existing docs carefully enough. But even with that, I still spent too much time breaking my head over setup friction, so the goal of this PR is to refine the docs to make the path clearer and less error-prone.

### DIFF
--- a/docs/dev-setup-server.md
+++ b/docs/dev-setup-server.md
@@ -1,5 +1,7 @@
 # Setting up a server from a development build
 
+## Build and run MOTIS with bundled UI
+
 1. Build `motis`. Refer to the respective documentation if necessary:
    - [for Linux](linux-dev-setup.md)
    - [for Windows](windows-dev-setup.md)
@@ -12,23 +14,23 @@
     motis/ui$ pnpm install
     motis/ui$ pnpm -r build
     ```
-   
+
 3. Move the UI build into the build folder of `motis`:
     ```shell
     motis$ mv ui/build build/ui
     ```
-   
+
 4. Copy the tiles profiles to the `motis` build folder:
     ```shell
     motis$ cp -r deps/tiles/profile build/tiles-profiles
     ```
-   
+
 5. Download OpenStreetMap and timetable datasets and place them in the build folder of `motis`:
     ```shell
     motis/build$ wget https://github.com/motis-project/test-data/raw/aachen/aachen.osm.pbf
     motis/build$ wget https://opendata.avv.de/current_GTFS/AVV_GTFS_Masten_mit_SPNV.zip
     ```
-   
+
 6. Run `motis config` on the downloaded datasets to create a config file:
     ```shell
     motis/build$ ./motis config aachen.osm.pbf AVV_GTFS_Masten_mit_SPNV.zip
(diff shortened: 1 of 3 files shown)


docs/linux-dev-setup.md
```diff
@@ -37,10 +39,31 @@ ninja
...
+mkdir build && cd build
+CXX=clang++-21 CC=clang-21 cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -GNinja ..
+ninja
+```
+
+## Troubleshooting
+
+### Switching compilers does not take effect
+
+CMake caches the selected compiler in the build directory. When switching compilers,
```
felixguendling: that's only the case because we use environment variables here

when using cmake flags like `-DCMAKE_CXX_COMPILER` directly, this would not be necessary

docs/linux-dev-setup.md
```diff
@@ -37,10 +39,31 @@ ninja
...
+
 ```sh
 git clone git@github.com:motis-project/motis.git
 cd motis
 mkdir build && cd build
 CXX=clang++-21 CC=clang-21 CXXFLAGS=-stdlib=libc++ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -GNinja ..
 ninja
 ```
+
+### Local build without `libc++`
```
felixguendling: we don't officially support clang-21 without libc++

we support:
- GCC with libstdc++
- Clang with libc++

with other combinations like GCC with libc++ or Clang with libstdc++ you're basically on your own and I would not want to advertise those combinations (which doesn't mean that if someone wants to fix compiler errors for such a not supported combination we would not accept this PR)

docs/dev-setup-server.md
```diff
@@ -41,3 +43,21 @@
...
+    motis/build$ ./motis server
+    ```
+
+2. In a second terminal, start the UI dev server:
+    ```shell
+    motis/ui$ pnpm dev
+    ```
+
+3. Open the UI with an explicit backend target:
+    - `http://localhost:5173/?motis=8080`
```
felixguendling:
```suggestion
    - `http://localhost:5173/?motis=http://localhost:8080`
```

it's probably better to show that you can also put full URLs

so for UI development that don't require backend changes, its usually fine to use `http://localhost:5173/?motis=http://api.transitous.org`

docs/linux-dev-setup.md
KristjanESPERANTO: Ah, looks like I created part of my own problems there 😅 Removed.

docs/linux-dev-setup.md
KristjanESPERANTO: Removed.

docs/dev-setup-server.md
KristjanESPERANTO: Oh wow, that is super practical, thanks! This would have saved me quite a bit of head-scratching around backend setup 🙂   

Using `http://api.transitous.org` caused a CORS issue, but `https://api.transitous.org` works. I updated the docs accordingly and squashed the commits to keep the git history clean.

PR #1391, opened 2026-05-01
add user agent to global variables

### LINKED ISSUE #963: Allow to set User-Agent header for fetching realtime data
I'll probably work in this myself at some point, just opening an issue to remember it.

DIDONAX:
- fixes: https://github.com/motis-project/motis/issues/963
- will not overwrite feed specific user-agent headers if they exist.

### DIFF
--- a/include/motis/motis_instance.h
+++ b/include/motis/motis_instance.h
@@ -3,7 +3,6 @@
 
 #include "boost/asio/io_context.hpp"
 
-#include "motis/config.h"
 #include "net/web_server/query_router.h"
 
 #include "utl/set_thread_name.h"
--- a/include/motis/motis_instance.h
+++ b/include/motis/motis_instance.h
@@ -179,7 +179,7 @@ struct motis_instance {
     }
   }
 
-  void run(data& d, config& c) {
+  void run(data& d, config const& c) {
     if (d.w_ && d.l_ && c.has_gbfs_feeds()) {
       gbfs_ = io_thread{"motis gbfs update", [&](boost::asio::io_context& ioc) {
                           gbfs::run_gbfs_update(ioc, c, *d.w_, *d.l_, d.gbfs_,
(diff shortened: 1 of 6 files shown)


include/motis/motis_instance.h
```diff
@@ -178,13 +178,41 @@ struct motis_instance {
   }
 
   void run(data& d, config const& c) {
+    bool has_user_agent = c.user_agent_.has_value();
+    auto ensure_user_agent = [&](auto& headers) {
+      auto& h = const_cast<std::optional<headers_t>&>(headers);
```
felixguendling: don't use `const_cast`

include/motis/motis_instance.h
```diff
@@ -178,13 +178,41 @@ struct motis_instance {
...
+      if (!h.has_value()) {
+        h = headers_t{};
+      }
+      if (!h->contains("User-Agent")) {
```
felixguendling: `try_emplace`?

include/motis/motis_instance.h
```diff
@@ -177,14 +177,35 @@ struct motis_instance {
     }
   }
 
-  void run(data& d, config const& c) {
+  void run(data& d, config& c) {
+    bool has_user_agent = c.user_agent_.has_value();
+    auto ensure_user_agent = [&](auto& h) {
+      if (!h.has_value()) h = headers_t{};
```
felixguendling:
```suggestion
      if (!h.has_value()) {
        h = headers_t{};
      }
```

always use brackets

include/motis/server.h
```diff
@@ -9,8 +9,8 @@ namespace motis {
 struct data;
 struct config;
 
-int server(data d, config const& c, std::string_view);
+int server(data d, config& c, std::string_view);
```
felixguendling: I think config should stay const.

Modify in `config config::read(std::string const& s)`.

src/config.cc
```diff
@@ -85,7 +85,35 @@ config config::read_simple(std::vector<std::string> const& args) {
...
+          }
+        }
+      }
+    }
+  }
+
+  return c;
 }
 
 config config::read(std::string const& s) {
```
felixguendling: should be here

PR #1396, opened 2026-05-09
Support for HTTP CONNECT Proxies

### LINKED ISSUE #263: Use HTTP CONNECT method for proxy server
Currently, MOTIS provides basic support for proxy servers using the `ris.http_proxy` config option. However, this works only for HTTP and not for HTTPS requests. In order for HTTPS connections to work, it would be necessary to use the [HTTP CONNECT method](https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_method) instead. As most GTFS-RT APIs require HTTPS, the GTFS-RT updates are practically unusable if you need to use MOTIS behind a company proxy server.

DIDONAX:
- fixes https://github.com/motis-project/motis/issues/263
- fixes bug where using tls was always determined by the proxy scheme rather than target scheme.
- fixes same issue with choosing the SNI.

### DIFF
--- a/include/motis/http_req.h
+++ b/include/motis/http_req.h
@@ -1,7 +1,9 @@
 #pragma once
 
+#include <boost/beast/core/stream_traits.hpp>
 #include <chrono>
 #include <map>
+#include <print>
 #include <string>
 
 #include "boost/asio/awaitable.hpp"
@@ -49,6 +51,7 @@ boost::asio::awaitable<void> http_CONNECT(Stream& stream,
   if (!proxy) {
     co_return;
   }
+  std::println("making a connect request");
   auto const target = std::string(url.host()) + ":" +
                       (url.has_port() ? std::string(url.port()) : "443");
 
@@ -56,11 +59,11 @@ boost::asio::awaitable<void> http_CONNECT(Stream& stream,
   req.set(http::field::host, target);
   beast::flat_buffer buf;
 
-  co_await http::async_write(stream.next_layer(), req);
+  co_await http::async_write(beast::get_lowest_layer(stream), req);
 
   http::response_parser<http::empty_body> res;
   res.skip(true);
-  co_await http::async_read_header(stream.next_layer(), buf, res);
+  co_await http::async_read_header(beast::get_lowest_layer(stream), buf, res);
 
   if (res.get().result() != http::status::ok) {
     throw utl::fail("CONNECT failed: target={}, status={}", target,
--- a/include/motis/http_req.h
+++ b/include/motis/http_req.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <boost/beast/core/stream_traits.hpp>
 #include <chrono>
 #include <map>
 #include <string>
--- a/include/motis/http_req.h
+++ b/include/motis/http_req.h
@@ -3,7 +3,6 @@
 #include <boost/beast/core/stream_traits.hpp>
 #include <chrono>
 #include <map>
-#include <print>
 #include <string>
 
 #include "boost/asio/awaitable.hpp"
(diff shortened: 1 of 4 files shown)


DIDONAX:
- @felixguendling i am not so sure of this one:  removed use_tls flag from proxy struct, the flag was intended to support client-proxy encryption, but from what i read its not common especially in conjunction with connect, and if it exists i assume its used to hide Proxy-Authorization headers which we don't currently support. What do you think ?
- also proxy is only used for gbfs updates i think we should consider adding it to the other ep as well.

traines-source:
> also proxy is only used for gbfs updates i think we should consider adding it to the other ep as well.

As long as it is configurable separately. For Transitous we rely on setting a proxy for GBFS but not for GTFS-RT feeds.

src/http_req.cc
```diff
@@ -159,9 +160,7 @@ asio::awaitable<http::response<http::dynamic_body>> http_POST(
   auto n_redirects = 0U;
   auto next_url = url;
   while (n_redirects < 3U) {
-    auto const use_tls =
-        proxy.has_value() ? proxy->use_tls_
-                          : next_url.scheme_id() == boost::urls::scheme::https;
+    auto const use_tls = next_url.scheme_id() == boost::urls::scheme::https;
```
felixguendling, line 168: this will kill support for non-CONNECT proxies?
we still want to support "normal" proxies

include/motis/http_req.h
```diff
@@ -34,6 +40,30 @@ boost::asio::awaitable<http_response> http_POST(
     std::chrono::seconds timeout,
     std::optional<proxy> const& = std::nullopt);
 
+template <typename Stream>
+boost::asio::awaitable<void> http_CONNECT(Stream& stream,
+                                          boost::urls::url const& url,
+                                          std::optional<proxy> const& proxy) {
+  if (!proxy) co_return;
```
felixguendling: always use braces

felixguendling: bump @DIDONAX

src/http_req.cc
felixguendling: bump @DIDONAX

DIDONAX: i tested this with  proxy=https://rt.triptix.tech  and use_connect = false and it works.

src/http_req.cc
```diff
@@ -144,9 +140,7 @@ asio::awaitable<http::response<http::dynamic_body>> http_GET(
   auto n_redirects = 0U;
   auto next_url = url;
   while (n_redirects < 3U) {
-    auto const use_tls =
-        proxy ? proxy->use_tls_
-              : next_url.scheme_id() == boost::urls::scheme::https;
```
traines-source: why don't we need this anymore?

DIDONAX: this was wrong,  the request schema should always be that of the target: for example before for an http proxy a http request will always be forwarded even for an https target. i used mitmproxy to confirm this .

this is before (master branch) the feed endpoint is never reached: 
image

and after:
image

so basically it wasn't working before when proxy and target schema didnt match.

traines-source: For CONNECT yes, but also for non-CONNECT? If the proxy is https but the target is http, I still need to send a https request?

traines-source: Also depends on the config of the proxy I guess. In rt.triptix.tech the schema of the target is just hardcoded, in a dynamic proxy I'm not sure what is supposed to happen. But maybe I'm missing something. I think rt.triptix.tech might not handle redirects correctly in this case either.

DIDONAX: the screenshots are with use_connect=false so this is the normal case.  i also tested rt.triptix.tech with connect disabled and it seems to work. can you test if this is also the case for you?

traines-source: with rt.triptix.tech actually both variants should work because it listens on both 80 and 443 and because it will contact the upstream target with whatever scheme was configured for that target in the proxy.
But I guess in the end you're right that for a dynamic proxy, it is most important to know whether to connect to the upstream via http or https and without any additional headers etc it can only know that from it's own connection to the client – meaning all such proxies would need to listen on both http and https.

Do we know about someone else using the proxy feature already?

Issue #1406, opened 2026-05-15
GTFS Flex: overlapping destination access candidates can shadow faster Flex option in itinerary routing


maiertbi:
In `/api/v5/plan`, overlapping GTFS Flex destination access candidates can cause itinerary routing to select a slower Flex option even though a faster Flex option exists and is routable.

### Generic example

Assume the following topology:

- `A`: trip origin
- `B`: public transit stop reached by train/bus
- `D`: requested destination
- `F1`: Flex service from `B` directly to `D`
- `F2`: Flex service from `B` to nearby location `C`, followed by walking from `C` to `D`

Observed result in full itinerary search:

```text
A -> B by public transport
B -> C by Flex service F2
C -> D by walking
```

Expected/better result:
```text
A -> B by public transport
B -> D by Flex service F1
short final walk if needed
```

A **direct request from B to D can find F1**, so the Flex route exists and is routable. The problem appears when the Flex candidates are prepared as destination access options for the full itinerary search.

If multiple Flex destination access candidates are valid for the same destination search area and overlapping time windows, the selected candidate should not depend on insertion/internal ordering. For each relevant time interval/location, the faster applicable Flex candidate should dominate the slower one, while preserving the selected Flex mode/route ID for itinerary reconstruction.

### Root cause hypothesis
This does not appear to be caused by the row sorting inside of the GTFS itself, but rather in how overlapping Flex candidates are represented before itinerary routing.

MOTIS converts Flex access/egress options into time-dependent offsets for routing. Overlapping Flex candidates appear to be appended as offsets and sorted by validity start time, but not normalized by dominance/travel duration. If multiple candidates are valid in the same time window, one candidate can shadow another depending on internal ordering.

felixguendling:
Yes, overlapping areas are currently not supported. I assumed that there's one service per area. Are there real-world examples where a area is served by >= 2 OnDemand services?

maiertbi:
Yes, we have a real-world scenario where multiple OnDemand and Flex services overlap from a routing perspective.  We’re analysing the efficiency of „Hotelshuttles“ and „Almtaxis“ (hotel-specific OnDemand services) in alpine tourist regions of Austria. And hotels in those areas are often quite close together.

The way we currently model these services in GTFS Flex is roughly:

- each hotel-specific shuttle is represented as its own Flex service/route,
- shared public transport access points, such as railway or bus stations, can be pickup points for several of these services,
- the hotel/accommodation is represented as the destination of the corresponding service,
[...]

PR #1408, opened 2026-05-15
normalize overlapping time-dependent offsets

maiertbi:
Normalize overlapping GTFS Flex destination access candidates before inserting time-dependent offsets. This prevents slower overlapping Flex candidates from shadowing faster candidates during full itinerary routing.

The normalization should build validity breakpoints from all candidates and emits a piecewise offset sequence where each interval contains the shortest active candidate, or an unavailable offset if no candidate is active.

Relates to #1406 

Note: I am not primarily a C++ developer, so feedback on style or a more idiomatic implementation is very welcome.

### DIFF
--- a/include/motis/flex/flex.h
+++ b/include/motis/flex/flex.h
@@ -17,16 +17,6 @@ using flex_routings_t =
     hash_map<std::pair<nigiri::flex_stop_seq_idx_t, nigiri::stop_idx_t>,
              std::vector<mode_id>>;
 
-struct flex_td_candidate {
-  nigiri::unixtime_t valid_from_;
-  nigiri::unixtime_t valid_to_;
-  nigiri::duration_t duration_;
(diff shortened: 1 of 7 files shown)


felixguendling:
I moved it to a more general space because it might also concern wheelchair time-dependent offsets for elevators. Additionally, the design previously didn't enforce FIFO properties. .g. (10:00 - 11:00, 10min), (11:00 - 12:00, 5min) where already at 10:56 it makes sense to wait in order to use the 5min offset starting at 11:00 but the previous solution would choose the the slower option to use the 10min offset where 10:56 + 10min = 11:06 but waiting 10:56 - 11:00 and then doing the 5min offset yields 11:05. Maybe that's theoretic but I felt like if we touch this code anyway, we can as well fix every known issue about it.

PR #1420, opened 2026-05-20
Ensure 'blocked' is initialized when only using direct routing

MichaelKutzner:
Fixes an segmentation fault, when only using direct routing. For example when only running this test: `./motis-test --gtest_filter='motis.routing_osm_only_direct_walk'`

### DIFF
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -496,6 +496,10 @@ std::pair<std::vector<api::Itinerary>, n::duration_t> routing::route_direct(
   if (!w_ || !l_) {
     return {};
   }
+  if (ep::blocked.get() == nullptr) {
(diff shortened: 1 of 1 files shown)


src/endpoints/routing.cc
```diff
@@ -496,6 +496,10 @@ std::pair<std::vector<api::Itinerary>, n::duration_t> routing::route_direct(
   if (!w_ || !l_) {
     return {};
   }
+  if (ep::blocked.get() == nullptr) {
+    ep::blocked.reset(new osr::bitvec<osr::node_idx_t>{w_->n_nodes()});
```
felixguendling:
```suggestion
  if (blocked.get() == nullptr) {
    blocked.reset(new osr::bitvec<osr::node_idx_t>{w_->n_nodes()});
```

PR #1424, opened 2026-05-21
Allow multiple types of locations in geocoding

### LINKED ISSUE #1198: Allow multiple types of locations in geocoding

Kiarash-Parvizi:
resolves #1198

### DIFF
--- a/.pkg
+++ b/.pkg
@@ -15,9 +15,9 @@
   branch=master
   commit=d9930c90e440df761c8c4916c36072de9dd00f49
 [adr]
-  url=git@github.com:triptix-tech/adr.git
-  branch=master
-  commit=64ff88efc22622a79d38c7a7e101d4a86e906a1e
+  url=git@github.com:Kiarash-Parvizi/adr.git
+  branch=multi-filter-type
+  commit=30c37403aeb695988b4ad67e3dcee59c8ffb7d17
 [googletest]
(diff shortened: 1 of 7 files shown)


.pkg
```diff
@@ -15,9 +15,9 @@
   branch=master
   commit=d9930c90e440df761c8c4916c36072de9dd00f49
 [adr]
-  url=git@github.com:triptix-tech/adr.git
-  branch=master
-  commit=64ff88efc22622a79d38c7a7e101d4a86e906a1e
+  url=git@github.com:Kiarash-Parvizi/adr.git
+  branch=multi-filter-type
+  commit=30c37403aeb695988b4ad67e3dcee59c8ffb7d17
```
felixguendling:
```suggestion
  url=git@github.com:triptix-tech/adr.git
  branch=master
  commit=de66c7dd13f63189b9132fa7f4bb1e24d236bf79
```

traines-source:
what is our current philosophy? do we consider a breaking change in the API client a breaking change in the API?

felixguendling:
it's something we would try to avoid - but here, the alternative would be to introduce yet another parameter which also isn't clean

but in terms of compatibility, I would say it's a breaking change if the requests of an older client are rejected by the server or trigger another behavior which is not the case here

PR #1438, opened 2026-05-23
nearest endpoint

DIDONAX:
- requires: https://github.com/triptix-tech/openapi-cpp/pull/5, this will eliminate manual in path parameter parsing.


src/endpoints/nearest.cc
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  auto coord = parse_location(params.coordinates_);
+  utl::verify<net::bad_request_exception>(coord.has_value(),
+                                          "invalid coordinates");
+  utl::verify<net::bad_request_exception>(params.number_ >= 1,
+                                          "invalid number");
+
+  std::swap(coord->pos_.lat_, coord->pos_.lng_);
+  api::NearestResponse res;
```
felixguendling, line 30: AAA style

test/endpoints/nearest_test.cc
```diff
@@ -0,0 +1,85 @@
...
+       ++i) {
+    auto const profile = static_cast<osr::search_profile>(i);
+    for (auto n = 1U; n <= number; ++n) {
+      auto const q = query(profile, loc, n, radius);
+      auto const res = ep(boost::urls::url_view{q});
+      auto const expected = osr::with_profile(profile, [&]<typename P>(P&&) {
+        return d.l_->match<P>(
+            std::get<typename P::parameters>(osr::get_parameters(profile)), loc,
+            false, osr::direction::kForward, radius, nullptr);
+      });
```
felixguendling, line 43: don't run the same code 2x
expect concrete results, no computation

openapi.yaml
```diff
@@ -3077,7 +3077,45 @@ paths:
...
+
+        - name: number
+          in: query
+          required: false
+          description: Number of nearest segments that should be returned.
+          schema:
+            type: integer
+            default: 1
+
+        - name: radiuses
```
felixguendling, line 3105: name implies array?

traines-source:
Is `/nearest/v1/{profile}/{coordinates}` the path required by the "standard"? I think if possible we should then put it to a subpath (`/api/v6/osrm/nearest/v1/{profile}/{coordinates}` or sth), because it doesn't fit our path structure at all. If it's more flexible we should just make it fit our scheme IMO.

PR #1444, opened 2026-05-28
feat(stoptimes): Add eventType=BOTH to merge arrivals and departures in a single query

c22dev:
## Summary

This PR adds a new `both` query parameter to the /stoptimes endpoint that allows clients to retrieve both arrivals and departures in a single request.

## Motivation

Previously, the stop times endpoint required separate queries for arrivals (`arriveBy=true`) and departures (`arriveBy=false`). This was inconvenient for UIs that want to display a unified view of upcoming events at a stop (e.g. a departure board showing UIs both arrivals and departures interleaved by time).

### New `both` parameter

A new optional `both` query parameter (boolean, default: `false`) is available now:
- `both=true`: returns both arrivals and departures in a single merged, time-sorted response, ignoring `arriveBy`.
- `both=false` (default): respects `arriveBy` to return only arrivals (`true`) or only departures (`false`).


felixguendling:
`Place` already contains both event times, not sure what your PR adds?

Example:
```json
{
    "name":"Darmstadt Hauptbahnhof",
    "stopId":"de-DELFI_de:06411:4734:49:49",
    "importance":6.440422683954239E-2,
    "lat":4.9872303E1,
    "lon":8.631478999999999E0,
    "level":0E0,
    "tz":"Europe/Berlin",
    "arrival":"2026-05-28T19:52:00Z",
    "departure":"2026-05-28T19:53:00Z",
    "scheduledArrival":"2026-05-28T19:51:00Z",
    "scheduledDeparture":"2026-05-28T19:51:00Z",
    "scheduledTrack":"19 Ost",
    "track":"4 Ost",
    "description":"Bus Platz 19",
    "vertexType":"TRANSIT",
    "pickupType":"NORMAL",
    "dropoffType":"NORMAL",
    "cancelled":false,
    "modes":[...]
}
``` 

The logic in the PR also seems to be inconsistent, not ordering by event time (sometimes it's departure, sometimes arrival time - depending on whether it's first stop or not) because that's not even possible when one entry always contains two events (departure + arrival).

felixguendling:
Before writing code also please open an issue to discuss

1) is this something that the project needs
2) what's the best way to implement it

c22dev:
Thanks for your answer !

As discussed before in the [Matrix channel](https://matrix.to/#/!kDMWQgLgnsbvxbEJdH:matrix.org/$XI-vsJfvbJF46fk4E3ZWIHgcrGPFnFT_DaSK3ezTqQQ?via=matrix.org&via=mozilla.org&via=gitter.im), there is currently an issue if you fetch departures using stoptime ; you don't get the trips that arrive at the stop.
If a trip has the queried stop for final stop, it won't appear (as it's considered only an arrival). In some cases it's kind of unpractical.

And yeah sorry, if needed I can open one now.

traines-source (commented):
I think this is very useful (would be for TeSpace anyways :D) and LGTM in general.

I'm just not entirely convinced of the new API param. I think it could maybe be possible to change `arriveBy` to not required -> std::optional, but with default false, and then be able to explicitly set `arriveBy=` to set it to empty/both? Not sure if that works. If it doesn't work, I'd still prefer a simple additional boolean flag `both` or something.

c22dev:
Thanks for your feedback!

I tried to implem your second suggestion and make it so it uses a new `both` boolean (default false). If `both=true`, it returns both arrivals and departures ignoring `arriveBy` if that's what you meant.

Tell me if it looks good to you!

openapi.yaml
```diff
@@ -2265,6 +2265,15 @@ paths:
...
             format: date-time
 
+        - name: both
+          in: query
+          required: false
+          schema:
+            type: boolean
+            default: false
+          description: |
+            Optional. Default is `false`. If set to `true`, returns both arrivals and departures, ignoring `arriveBy`.
```
felixguendling, line 2431: `arriveBy` is never ignored in the code? It's always used for sorting if both events are available?

src/odm/meta_router.cc
```diff
@@ -145,7 +145,9 @@ meta_router::meta_router(ep::routing const& r,
                           : query_.ignorePostTransitRentalReturnConstraints_},
       dest_ignore_rental_return_constraints_{
           query.arriveBy_ ? query_.ignorePostTransitRentalReturnConstraints_
-                          : query_.ignorePreTransitRentalReturnConstraints_} {}
+                          : query_.ignorePreTransitRentalReturnConstraints_} {
+  (void)direct_modes_;
```
felixguendling: ?

openapi.yaml
felixguendling: this was not a request to change code, rather a question

felixguendling (approved):

Issue #1465, opened 2026-06-05
Rename "Trips by [...]" in dropdown in top-right to "Vehicles by [...]"


Kladki:
To me at least, I was confused at first when I saw it called "trips by", since I usually imagine "trip" being synonymous with "journey" or "itinerary". I think it makes more sense to have it calls "vehicles by", since the dots on the maps are the positions of vehicles after all.

Lach-anonym:
It's because of GTFS parlance, sometimes it can be weird, but it's better to keep it that way, so it's known it's about the GTFS/NeTEx trips and just keep it consistent
I just could propose changing "Trips by mode" to "Trips by vehicle type" and add "colours" to "Trips by route"

traines-source:
I'm also not too happy with the wording (in other languages it is even weirder). But what is shown is not the actual position of vehicles but the interpolated position according to scheduled or realtime timetable data. IMO a wording including "vehicle" makes this even more likely to be interpreted as "actual position".

felixguendling:
Yes. The main problem I see is that "Vehicle" could make the impression that we are parsing GTFS RT VehicleUpdates which is currently not the case. So this "Vehicle Positions" naming would definitely be more straightforward but also quite misleading.

manfredbrandl:
What about "Vehicles by timetable" and "Vehicles by forecast" or "Vehicles by time forecast"? In German "Fahrzeuge nach Fahrplan" und "Fahrzeuge nach Prognose"?

Kladki:
> But what is shown is not the actual position of vehicles but the interpolated position according to scheduled or realtime timetable data.

Oh, that's really not the impression I got when seeing those dots on the map. I assumped that having dots for a trip with a position too of the direction that it's fscing, that it was the actual positions of vehicles.

manfredbrandl:
> Yes. The main problem I see is that "Vehicle" could make the impression that we are parsing GTFS RT VehicleUpdates which is currently not the case. So this "Vehicle Positions" naming would definitely be more straightforward but also quite misleading.

For me "Vehicle Positions" hints more to GTFS RT VerhicleUpdates then "Vehicles" only. Adding "by timetable/delay" will hint that we are not parsing VehicleUpdates, which are not available everywhere by the way.

traines-source:
> by timetable/delay

This to me would imply that the interpolated position depends on this setting, which is not true (the interpolated position is always according to realtime delay, if available)

[...]

PR #1471, opened 2026-06-08
Update German translations in de.ts

manfredbrandl:
Enthält auch Verbesserungen nach persönlichem Geschmack "Fahrten nach Fahrplan", "Achtung: Mautpflichtige Straße!"


ui/src/lib/i18n/de.ts
```diff
@@ -177,11 +177,11 @@ const translations: Translations = {
 		none: 'Nichts',
 		stops: 'Haltestellen',
 		rt: 'Fahrten nach Echtzeit',
-		route: 'Fahrten nach Linie',
+		route: 'Fahrten nach Fahrplan',
```
traines-source: The labeling of these options surely has room for improvement (cf. https://github.com/motis-project/motis/issues/1465), but I think this is more misleading, since this is about line colors given by the feed providers / agencies. So maybe "Fahrten nach Linienfarbe"?

ui/src/lib/i18n/de.ts
```diff
@@ -145,8 +145,8 @@ const translations: Translations = {
 		information: 'Informationen',
 		more: 'mehr'
 	},
-	RENTAL_BICYCLE: 'Bikesharing',
-	RENTAL_CARGO_BICYCLE: 'Lastenrad Sharing',
+	RENTAL_BICYCLE: 'Fahrradverleih',
+	RENTAL_CARGO_BICYCLE: 'Lastenradverleih',
```
traines-source: Fahrradverleih to me sounds more like a longer term rental scheme for holidays etc, not for a 10 minute bike ride. But maybe that's just me. Maybe "Leihrad", "Leih-Lastenrad"? And then for consistency we should also change "Car Sharing", "Moped Sharing" etc? Or we end up leaving it as it is?

ui/src/lib/i18n/de.ts
manfredbrandl: See my proposal at https://github.com/motis-project/motis/issues/1465#issuecomment-4650548314

ui/src/lib/i18n/de.ts
manfredbrandl: As Rental is Sharing here what about "Fahrrad Sharing" and "Lastenfahrrad Sharing"?

ui/src/lib/i18n/de.ts
felixguendling: > What about "Vehicles by timetable" and "Vehicles by forecast" or "Vehicles by time forecast"? In German "Fahrzeuge nach Fahrplan" und "Fahrzeuge nach Prognose"?

I think it would help to mention that it's about the delay. So maybe `Fahrten nach Verspätung` would describe what the colors mean. Or if it can be longer it could also be `Fahrten eingefärbt nach [...]` (Verspätung, Linienfarbe, Verkehrsmittel).

manfredbrandl: As there is no really good equivalent for "Verfrühugen" in English i think "delay" is the best choice und should be understand to include vehicles or trips ahead of schedule. I would prefer vehicles over trips because on the map there are only vehicle positions and not full trips.

manfredbrandl: I did update my pr according to https://github.com/motis-project/motis/issues/1465#issuecomment-4663777209

ui/src/lib/i18n/de.ts
manfredbrandl: I did update my pr according to https://github.com/motis-project/motis/issues/1465#issuecomment-4663777209

traines-source:
Thanks! I'll follow up with the other languages.

PR #1473, opened 2026-06-09
query without realtime

### LINKED ISSUE #1136: Feature Request: query scheduled data without realtime via api
at first thank you for this amazing project and the public access ❤️ this really opens a lot of possibilities, its hard to get access to german train data

I did not found a way for this project to get scheduled times only (without realtime data) because in my case i want to find delayed or cancelled trains.

If i use the existing api, trains/busses that are delayed, a lot are not beeing shown because the output is sorted by realtime data

it would help me a lot to have a fitting api for station and journey (non realtime data)

thanks for your great work

Kiarash-Parvizi:
resolves #1136


traines-source:
As discussed in https://github.com/motis-project/motis/pull/1355#issuecomment-4154125679 I think we will want to have more different variants of combining RT and non-RT. While we do not have to implement them all in this PR, I think it would be good to use an enum instead of a boolean in the API to be able to easily to extend with additional variants.

Also I'm not sure if there really is a use case for not even annotating with realtime (i.e. searching based on scheduled and not annotating the current realtime situation to the found itineraries) because the scheduled information is contained nonetheless if one annotates with realtime information. The issue you linked to also sounds to me more like even the hybrid mode would actually be useful to them.

TLDR, I propose to change the API param to `realtimeMode` and make it an enum:
* `REALTIME` (current default)
* `SCHEDULED` (if you think it's easy to annotate with realtime data (just always pass rtt to journey_to_response?), let's do it, otherwise, let's introduce a separate variant later that does it and make this really a scheduled-only variant)

and later more variants. Same for stoptimes for consistency IMO, even if there it only influences sorting.

Kiarash-Parvizi:
> if you think it's easy to annotate with realtime data (just always pass rtt to journey_to_response?)

Yeah I think just passing rtt to journey_to_response is sufficient.

> Also I'm not sure if there really is a use case for not even annotating with realtime (i.e. searching based on scheduled and not annotating the current realtime situation to the found itineraries)

So a scheduled-only variant is not needed right?
Also what other variants are you thinking of for the future?

traines-source:
> So a scheduled-only variant is not needed right?

Actually, HAFAS has such a mode: https://pastebin.com/qZ9WS3Cx So I guess someone saw a need. So I changed my mind: let's keep this PR simple as it is (actual scheduled-only variant) and just change the boolean to an enum with `OFF` and `REALTIME`.

Then later we can add:
* `INFOS` - route/sort/window by scheduled, but annotate with RT
* `FULL` - search on both scheduled and realtime state (to still show cancelled and broken connections, but also new alternatives according to RT, naively two completely separate RAPTOR runs are necessary)
* `HYBRID` - search on rt only if itineraries found in scheduled are broken according to realtime info

(Since I actually just reassigned a related topic as a thesis proposal, please wait with the implementation of these since IMO it's a good starter for that person)

PR #1474, opened 2026-06-10
HGV Support

pablohoch:
Depends on https://github.com/motis-project/osr/pull/111 + https://github.com/motis-project/osr/pull/114

Works with the standard MOTIS UI (deselect public transport, select Direct Connection: HGV) or with this [HGV test UI](https://github.com/motis-project/hgv-ui), which also includes a quick comparison between car/hgv and highlights ways with HGV restrictions or conditionals on the map.


felixguendling:
@pablohoch requires merge

include/motis/osr/parameters.h
```diff
@@ -1,9 +1,10 @@
 #pragma once
 
-#include "osr/routing/parameters.h"
-#include "osr/routing/profile.h"
+#include <cstdint>
 
 #include "motis-api/motis-api.h"
+#include "osr/routing/parameters.h"
```
felixguendling, line 5: why change header grouping?

src/endpoints/initial.cc
```diff
@@ -15,6 +17,58 @@ namespace n = nigiri;
 
 namespace motis::ep {
 
+namespace {
```
felixguendling: why namespace? we don't do this anywhere else?

src/endpoints/initial.cc
```diff
@@ -15,6 +17,58 @@ namespace n = nigiri;
...
+
+std::pair<geo::latlng, unsigned> get_center_and_zoom(double const lat_min,
```
felixguendling: take geo::latlng already here -> avoids double conversion at least for osr case

test/endpoints/initial_test.cc
```diff
@@ -26,3 +27,23 @@ TEST(initial, server_config) {
...
+
+  auto d = data{"test/data_osm_only_initial", c};
+  d.init_initial("v1.2.3-test");
+  auto const ep = utl::init_from<motis::ep::initial>(d).value();
+
+  auto const res = ep("");
+
+  EXPECT_NE(0.0, res.lat_);
+  EXPECT_NE(0.0, res.lon_);
+  EXPECT_GT(res.zoom_, 0.0);
```
felixguendling: maybe assert numbers more precisely?

test/odm/prima_test.cc
```diff
@@ -61,7 +61,7 @@ TRANSFERS: 0
      FROM: (START, START) [1970-01-01 09:57]
        TO: (END, END) [1970-01-01 12:00]
 leg 0: (START, START) [1970-01-01 09:57] -> (A, A) [1970-01-01 10:55]
-  MUMO (id=16, duration=58)
+  MUMO (id=17, duration=58)
```
felixguendling, line 64: how come this changes?

.pkg.lock
```diff
@@ -1,25 +1,31 @@
...
-res b759b93316afeb529b6cb5b2548b24c41e382fb0
-date ce88cc33b5551f66655614eeebb7c5b7189025fb
-googletest 9758d168fd1ccff2d76f3bd6ef802943d0f8f7e5
+zlib-ng d68e1d908e789c31c1f2fafe4bd8e09cb91e21c5
+boost 77467bf580d98ea06716e2931cbe3b1f28e0cd37
 docs 75dc89a53e9c2d78574fc0ffda698e69f1682ed2
 fmt 34a5bee1df1f0d144de0e127807d161d08670472
+googletest 9758d168fd1ccff2d76f3bd6ef802943d0f8f7e5
 utl 5ddd9956455b18fc96fd5f8095f7552d5be92c17
+ctx 9b495bdd798520007a4f1c13e51766a26f10ef10
```
felixguendling, line 9: maybe we should sort this file in pkg so diff makes sense :D

CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(motis LANGUAGES C CXX ASM)
 
 set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 
+enable_testing()
```
felixguendling: what does this do? where do we use it?
does it activate IDE features?

include/motis/osr/parameters.h
pablohoch: Not intentional, caused by a different clang-format version. I'll change it back.

CMakeLists.txt
pablohoch: It enables [ctest](https://cmake.org/cmake/help/latest/manual/ctest.1.html) integration, so that `ctest` can run all tests in motis, osr etc.
However, the way it is currently integrated it doesn't allow filtering by specific unit test IIRC (only motis vs. osr etc.).
There is a way to better integrate this ([GoogleTest](https://cmake.org/cmake/help/latest/module/GoogleTest.html)), but I haven't looked into that.
I'll just remove that again for now.

test/odm/prima_test.cc
pablohoch: Because in [transport_mode_ids.h](https://github.com/motis-project/motis/blob/master/include/motis/transport_mode_ids.h#L9):
```cpp
constexpr auto const kOdmTransportModeId =
    static_cast<nigiri::transport_mode_id_t>(osr::kNumProfiles);
 ```
 
 And the osr changes add a new osr profile for HGV routing.

CMakeLists.txt
pablohoch: (There are also cmake test presets, which are quite nice)

PR #1478, opened 2026-06-12
fastestDirectFactor has been superseded by total non-transit duration filter

traines-source:
this is bad because
1) the UI default differs from the actual default so we don't usually test the actual default
2) in the current state, clients can not turn on faster direct modes by default because they will far too often dominate PT results

This reintroduces some weird cases when selecting faster modes for pre/post than for direct, since then the total non-transit duration filter (direct_filter.cc) doesn't apply. I think these are of minor importance.

### DIFF
--- a/src/direct_filter.cc
+++ b/src/direct_filter.cc
@@ -16,8 +16,13 @@ void direct_filter(std::vector<api::Itinerary> const& direct,
                    std::vector<n::routing::journey>& journeys) {
   auto const get_direct_duration = [&](auto const transport_mode_id) {
     auto const m = to_mode(transport_mode_id);
-    auto const i = utl::find_if(
-        direct, [&](auto const& d) { return d.legs_.front().mode_ == m; });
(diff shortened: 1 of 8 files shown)


src/direct_filter.cc
```diff
@@ -13,7 +15,7 @@ namespace n = nigiri;
 void direct_filter(std::vector<api::Itinerary> const& direct,
                    std::vector<n::routing::journey>& journeys) {
   auto const get_direct_duration = [&](auto const transport_mode_id) {
-    auto const m = static_cast<api::ModeEnum>(transport_mode_id);
+    auto const m = to_mode(transport_mode_id);
```
felixguendling, line 18: will still not work with rental/flex since they often have walk first

traines-source:
it's still not very satisfying because we do not return `numItineraries` anymore if some of them are filtered, even though a bit later there may be a reasonable PT itinerary. But oh well...

PR #1479, opened 2026-06-12
UI: itinerary tags

mority:
I thought it would be nice to highlight itineraries in the list that are best at something.

Currently, I would mark the ones with the lowest number of transfers and/or the shortest travel time. 

But maybe shortest walking duration would also be interesting, or longest(safest) transfers, etc.

Very early stage, we should probably use some pill tags or emoji/icons to make it look nicer.

### DIFF
--- a/ui/src/lib/ItineraryList.svelte
+++ b/ui/src/lib/ItineraryList.svelte
@@ -33,6 +33,19 @@
 		updateStartDest: (r: Awaited<RequestResult<PlanResponse, ApiError, false>>) => PlanResponse;
 	} = $props();
 
+	let minTransfers = $state(Number.MAX_SAFE_INTEGER);
+	let minDuration = $state(Number.MAX_SAFE_INTEGER);
+	$effect(() => {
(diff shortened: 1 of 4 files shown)


traines-source:
In the long run we probably even want to return an enum from the backend? With more criteria it will be difficult to tell why an itinerary was included.
There are surely fitting Lucide icons for everything :)

mority:
You're right. I only thought about this as a client-side feature but it could actually be part of the API.

mority:
This is what I came up with. At first I only showed the total walking time for the itinerary with the longest total walking time, but maybe it is interesting to always show it. Although, oftentimes it might not be that interesting.

image

manfredbrandl:
> But maybe shortest walking duration would also be interesting, or _longest(safest) transfers_, etc.

Transfers are save if the departure vehicle can still be reached even if the feeder vehicle is delayed; this can be achieved either by allowing a buffer in the transfer time (longest transfer) or by ensuring a connection, meaning that if the feeder vehicle is delayed, the departure vehicle departs later than scheduled.

PR #1481, opened 2026-06-14
Mapterhorn relative url to avoid privacy issues

traines-source:
for dev purposes you may use in your vite.config.ts: 
```
proxy: {
			'/mapterhorn/': {
				changeOrigin: true,
				target: 'https://api.transitous.org'
			}
}
```

### DIFF
--- a/ui/src/lib/map/style.ts
+++ b/ui/src/lib/map/style.ts
@@ -146,7 +146,11 @@ export const getStyle = (
 		? {
 				hillshadeSource: {
 					type: 'raster-dem',
-					url: 'https://tiles.mapterhorn.com/tilejson.json'
+					tiles: [getAbsoluteUrl(apiBaseUrl, 'mapterhorn/{z}/{x}/{y}.webp')],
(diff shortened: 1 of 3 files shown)


felixguendling:
> for dev purposes you may use in your vite.config.ts:
> 
> ```
> proxy: {
> 			'/mapterhorn/': {
> 				changeOrigin: true,
> 				target: 'https://api.transitous.org'
> 			}
> }
> ```

this should be the default in the vite.config.ts then so the UI can be tested without having to know this special PR id

ui/src/lib/map/style.ts
```diff
@@ -146,7 +146,11 @@ export const getStyle = (
 		? {
 				hillshadeSource: {
 					type: 'raster-dem',
-					url: 'https://tiles.mapterhorn.com/tilejson.json'
+					tiles: [getAbsoluteUrl(apiBaseUrl, 'mapterhorn/{z}/{x}/{y}.webp')],
```
felixguendling, line 149: should still work by default (also add vite config lines)

PR #1482, opened 2026-06-16
Improve provider selection

### LINKED ISSUE #826: GBFS: Provider selection performance
Currently (as of #824), all GBFS providers that are in a radius reachable by car are chosen for routing direct routes. For bike products, this may include providers that are too far away and can't produce routes within the time limit.

To improve routing performance, the radius used to select the products should depend on the form factor (car vs. bike).

Possible solutions include:
- Instead of a provider r-tree, use two product-specific r-trees: One for car products and one for the rest (bike, scooter etc). Then the correct radius can be used for lookup in each r-tree.
- Store a bounding box for each product. After lookup in the single provider r-tree, check the distance to the bounding box based on the product form factor to filter the results.

In both cases, the r-trees or bounding boxes need to be updated during the GBFS feed update whenever a vehicle is added/removed/moved, vehicles become available at a station, or the partitioning changes etc.

Kiarash-Parvizi:
resolves #826

### DIFF
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -281,16 +281,16 @@ std::vector<n::routing::offset> get_offsets(
         return;
       }
 
-      auto const foot_dist =
+      auto const foot_radius =
           get_max_distance(osr::search_profile::kFoot, osr_params, max);
       auto const car_radius =
           dir == osr::direction::kForward
-              ? foot_dist
+              ? foot_radius
               : get_max_distance(osr::search_profile::kCarSharing, osr_params,
                                  max);
       auto const bike_radius =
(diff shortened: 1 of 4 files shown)


pablohoch (approved):

src/endpoints/routing.cc
```diff
@@ -281,14 +281,33 @@ std::vector<n::routing::offset> get_offsets(
         return;
       }
 
-      auto const max_dist_to_departure =
+      auto const foot_dist =
```
felixguendling: why not also foot_radius like for car_radius and bike_radius?
dist vs radius

felixguendling (approved):

PR #1487, opened 2026-06-17
extend realtimeMode

Kiarash-Parvizi:
continuing #1473


felixguendling:
I would prefer to merge a `INFO` only PR first as it's simple (as long as we keep sorting by scheduled departure time) and self-contained.


IMO the `HYBRID` mode is not clearly defined regarding paging?

Kiarash-Parvizi:
> IMO the `HYBRID` mode is not clearly defined regarding paging?

Yes, I still don't know how to solve this problem. Maybe we can treat it as FULL whenever pageCursor is present?

felixguendling:
> Maybe we can treat it as FULL whenever pageCursor is present?

It can still create inconsistencies with the first search (first search `HYBRID`, no pageCursor, second search with pageCursor). `INFO` should hopefully be a quick win without any problems as long as we keep scheduled time sorting from the search.

Kiarash-Parvizi:
Is it ok for delayed trips to appear twice in the FULL mode?

felixguendling:
> Is it ok for delayed trips to appear twice in the FULL mode?

We will have to filter duplicate journeys server side, otherwise it's not better than doing two parallel requests from the client.

PR #1489, opened 2026-06-18
set page title state

### LINKED ISSUE #1402: Set the page title based on the current page
E.g. "`origin` to `destination`"
It would help if you have multiple MOTIS tabs open, are looking through your browser history, or just had the tab open for a while and forgot what you were searching there (without having to load/switch to the tab).

DIDONAX:
fixes : https://github.com/motis-project/motis/issues/1402

### DIFF
--- a/ui/src/lib/pageTitle.ts
+++ b/ui/src/lib/pageTitle.ts
@@ -19,9 +19,7 @@ export const getPageTitle = (state: PageTitleState, translations: Translations):
 	if (state.activeTab === 'departures' && state.selectedStop) {
 		const stopName = state.stopName || state.selectedStop.name;
 		if (stopName) {
-			return state.stopArriveBy
-				? pageTitle.arrivalsAt(stopName)
-				: pageTitle.departuresAt(stopName);
+			return state.stopArriveBy ? pageTitle.arrivalsAt(stopName) : pageTitle.departuresAt(stopName);
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


ui/src/lib/pageTitle.ts
```diff
@@ -0,0 +1,42 @@
...
+	activeTab: 'connections' | 'departures' | 'isochrones';
+	from: Location;
+	to: Location;
+	one: Location;
+	selectedStop?: { name: string };
+	stopArriveBy?: boolean;
+	stopName?: string;
+};
+
+const locationLabel = (location: Location): string => location.label || location.match?.name || '';
```
krkk, line 14: I'd probably prioritize `location.match?.name` over `location.label` here imo. From what I understand, `location.label` tries to be very unambiguous (with City, Province, Country suffixes), whereas match name is usually just the place/stop name. The location label might be less ambiguous, but I'd prioritise shorter titles, especially in the Departures tab where we have to put two of them and arguably the latter seems more important.

krkk (commented):
idk if github notifies about opening pull requests to other pull requests, so you might also take a look at DIDONAX/motis/pull/1 for additional departure/arrival times. :)

ui/src/lib/pageTitle.ts
traines-source: I have decided to merge this first. I agree with you that too long titles are also not that helpful, that's why frankly I'm not convinced I would put the times there as you propose. But I'm not completely opposed either, so if you want you can redirect your PR to this repo then we can merge it. Also feel free to change the location name precedence if you feel it's appropriate.

krkk: I'd like to show times because that way they will also be remembered in browser's history, and I think that's cool. ;)
It might also help slightly when juggling between tabs of the same route.

I'll open a PR in a moment.

PR #1490, opened 2026-06-18
make advanced options a dialog

### LINKED ISSUE #1451: Move advanced options to a dialog
On mobile, it's possible to accidentially change it while scrolling, so it would be nice if there was a button to reset it to the default, as I can't imagine most people would remember what it was in that situation.

DIDONAX:
fixes: https://github.com/motis-project/motis/issues/1451


traines-source:
actually I think we should do it like that everywhere, even on desktop the remaining space for the results has become so tiny that you barely see it.
also it would be good if the plan requests are only fired when the dialog is closed to avoid sending bazillions of requests with every change.

PR #1494, opened 2026-06-22
realtimeMode::INFOS added

Kiarash-Parvizi:
Only `realtimeMode::INFOS` from #1487


Kiarash-Parvizi:
should I rename it to `INFO` from `INFOS`?

felixguendling:
There is no plural of information in English.

However, the word `INFO` doesn't say anything anyway.

In the code variables and the comment in the doc the word "annotate" is used more often, so maybe `ANNOTATE_REAL_TIME` would tell a bit more about what it actually does?

PR #1496, opened 2026-06-23
N-nearest offsets

### LINKED ISSUE #1477: N-nearest stops for pre/post/transfers
To avoid [weird routing results](https://api.transitous.org/?itineraryId=ClQhfJCvTt%2F8QkApiwrm%2BR28N0AyE2dyLUF0aGVucy1CdXNfNjAwMTU5%2FgjDgCX9QkBBoQ3ABkS8N0BI0Pet0QZQuPqt0QZaBFdBTEtxAAAAAAAAAAAKqQEKATESMTIwMjYwNjEyXzA2OjAwX2dyLUF0aGVucy1CdXNfMTc2MF9kYXlfNV80MDc0XzA2MDAaE2dyLUF0aGVucy1CdXNfNjAwMTUh%2FgjDgCX9QkApoQ3ABkS8N0AyE2dyLUF0aGVucy1CdXNfNjEzNjg5W3wKgPH%2BQkBB4uR%2Bh6K4N0BIuPqt0QZQzP%2Bt0QZaA0JVU2ABaQAAAAAAAAAAcQAAAAAAAAAACokBGhNnci1BdGhlbnMtQnVzXzYxMzY4IVt8CoDx%2FkJAKeLkfoeiuDdAMipnci1IZWxsZW5pYy1UcmFpbl82OTI4YzNjMTUwYWI1YmM3MmIzOWQwNzY58oRefxL%2FQkBB1zcwuVG4N0BI0Iau0QZQyIeu0QZaBFdBTEtpAAAAAAAAAABxAAAAAAAAAAAK5gEKBElDNTASMzIwMjYwNjEyXzA2OjU4X2dyLUhlbGxlbmljLVRyYWluX0lDNTAtMjAyNjA2MTItMDM1OBoqZ3ItSGVsbGVuaWMtVHJhaW5fNjkyOGMzYzE1MGFiNWJjNzJiMzlkMDc2IfKEXn8S%2F0JAKdc3MLlRuDdAMipnci1IZWxsZW5pYy1UcmFpbl82OTI4YzNjMTUwYWI1YmM3MmIzOWNlN2I5Syno9pJSREBBq%2BY5It%2FtNkBIyIeu0QZQ6Jmv0QZaDVJFR0lPTkFMX1JBSUxgAWkAAAAAAAAAAHEAAAAAAAAAAAqlARoqZ3ItSGVsbGVuaWMtVHJhaW5fNjkyOGMzYzE1MGFiNWJjNzJiMzljZTdiIUsp6PaSUkRAKavmOSLf7TZAMi9ldS1mbGl4YnVzXzliNjhkZjc0LTNlY2ItMTFlYS04MDE3LTAyNDM3MDc1Mzk1ZTmKWS%2BGclJEQEEmj6flB%2B42QEjwrrDRBlDor7DRBloEV0FMS2kAAAAAAAAAAHEAAAAAAAAAAAryAQoMRmxpeEJ1cyBOOTgwEjUyMDI2MDYxMl8xNDozMF9ldS1mbGl4YnVzX045ODAtOS0xNDMwMDYyMDI2LVROSyNCTS0wMBovZXUtZmxpeGJ1c185YjY4ZGY3NC0zZWNiLTExZWEtODAxNy0wMjQzNzA3NTM5NWUhilkvhnJSREApJo%2Bn5QfuNkAyL2V1LWZsaXhidXNfYmQ5Yzk4MDYtMzcxOS00ZTdkLTliNDUtYjE3YTgyYzA5M2I1OXHl7J3RxkRAQbAl5IOeRTdASOivsNEGUOj6sNEGWgVDT0FDSGABaQAAAAAAAAAAcQAAAAAAAAAACqoBGi9ldS1mbGl4YnVzX2JkOWM5ODA2LTM3MTktNGU3ZC05YjQ1LWIxN2E4MmMwOTNiNSFx5eyd0cZEQCmwJeSDnkU3QDIvZXUtZmxpeGJ1c19iZDljOTgwNi0zNzE5LTRlN2QtOWI0NS1iMTdhODJjMDkzYjU5ceXsndHGREBBsCXkg55FN0BI6Pqw0QZQ4Puw0QZaBFdBTEtpAAAAAAAAAABxAAAAAAAAAAAK8gEKDEZsaXhCdXMgTjk4MBI1MjAyNjA2MTJfMjA6NDVfZXUtZmxpeGJ1c19OOTgwLTgtMjA0NTA2MjAyNi1CTSNUTkstMDAaL2V1LWZsaXhidXNfYmQ5Yzk4MDYtMzcxOS00ZTdkLTliNDUtYjE3YTgyYzA5M2I1IXHl7J3RxkRAKbAl5IOeRTdAMi9ldS1mbGl4YnVzXzliNjhkZjc0LTNlY2ItMTFlYS04MDE3LTAyNDM3MDc1Mzk1ZTmKWS%2BGclJEQEEmj6flB%2B42QEjs%2BLPRBlDsw7TRBloFQ09BQ0hgAWkAAAAAAAAAAHEAAAAAAAAAAApwGi9ldS1mbGl4YnVzXzliNjhkZjc0LTNlY2ItMTFlYS04MDE3LTAyNDM3MDc1Mzk1ZSGKWS%2BGclJEQCkmj6flB%2B42QDmUysrl9VFEQEGq4Kb1be82QEjsw7TRBlC0yrTRBloEV0FMS2kAAAAAAAAAAA%3D%3D&joinInterlinedLegs=false&detailedTransfers=true&detailedLegs=true&withFares=true&numLegAlternatives=3&language=en&transitModes=TRANSIT&pedestrianProfile=FOOT&useRoutedTransfers=false&requireBikeTransport=false&requireCarTransport=false&debug=yes) and reduce the number of relations where we return "no itineraries found", in addition to the fixed maximum duration for pre/post/transfers (from API parameters or config) we want to optionally include the n nearest stops. I.e. if within the set duration less than n locations are reached, extend it to the n nearest locations.

First step: Just for pre/post. There should be api params for `n` similar to the duration params.
Second step: transfers. There should be a config option next to `max_footpath_length`. Additional footpaths are added in a fixed way during import.

Since OSR doesn't discover timetable locations on its own, we still need to get the surrounding locations from the rtree in routing.cc. I think exponentially increasing the rtree bbox until a sufficient number of locations is found should be good enough for the time being. Some factor more than `n` locations need to be discovered since there might be other locations that are physically closer than the locations found according to the beeline/bbox. This is a heuristic but `n` is a heuristic anyways. 

In OSR there is astar which supports early stopping based on the number of discovered locations I think. But since we still want to discover all locations within the configured walking duration, maybe it's easier to stick to normal dijkstra and pick an upper bound for the walking duration needed to reach the `n` locations, based on the bbox of the rtree (another heuristic, i.e. `max(configured_walking_duration, upper_bound_walking_duration_to_find_n)`). Then after OSR routing filter the obtained offsets by configured walking duration and `n`.

~~The maximum discovered distance should still be limited by the server limits given in the config.~~ Since the main reason for the server limits is GBFS in dense areas and/or long range car journeys, maybe we should have a separate higher limit (or separately limit GBFS)? I would like to find an itinerary with 5 hours of walking if the next stop is 5 hours away.

Testing needed how often this creates unnecessarily long walking connections depending on `n`. I would hope a low default value like 5 would work nicely. Once we have multi-criteria routing that optimizes walking duration, we can increase the default value to more like 100+.

aff3npirat:
Change routing behaviour to always find a mimium number of start/end stations or transfer options.

Closes #1477

### DIFF
--- a/include/motis/endpoints/routing.h
+++ b/include/motis/endpoints/routing.h
@@ -70,8 +70,7 @@ struct routing {
       double max_matching_distance,
       gbfs::gbfs_routing_data&,
       stats_map_t& stats,
-      unsigned const min_near_stations = 0U,
-      unsigned const max_bbox_increases = 5U) const;
+      unsigned const min_near_stations = 0U) const;
 
   nigiri::hash_map<nigiri::location_idx_t,
                    std::vector<nigiri::routing::td_offset>>
--- a/include/motis/endpoints/routing.h
+++ b/include/motis/endpoints/routing.h
@@ -69,7 +69,9 @@ struct routing {
       std::chrono::seconds max,
       double max_matching_distance,
       gbfs::gbfs_routing_data&,
-      stats_map_t& stats) const;
+      stats_map_t& stats,
+      unsigned const min_near_stations = 0U,
+      unsigned const max_bbox_increases = 5U) const;
 
   nigiri::hash_map<nigiri::location_idx_t,
                    std::vector<nigiri::routing::td_offset>>
--- a/include/motis/get_stops_with_traffic.h
+++ b/include/motis/get_stops_with_traffic.h
@@ -22,6 +22,5 @@ std::vector<nigiri::location_idx_t> get_stops_with_unique_routes(
     nigiri::rt_timetable const*,
     point_rtree<nigiri::location_idx_t> const&,
     osr::location const&,
-    double const,
-    hash_set<nigiri::route_idx_t>&);
+    double const);
 }  // namespace motis
--- a/include/motis/get_stops_with_traffic.h
+++ b/include/motis/get_stops_with_traffic.h
@@ -4,6 +4,7 @@
 
 #include "motis/fwd.h"
 #include "motis/point_rtree.h"
+#include "motis/types.h"
 
 namespace motis {
 
@@ -16,4 +17,11 @@ std::vector<nigiri::location_idx_t> get_stops_with_traffic(
     nigiri::location_idx_t const not_equal_to =
         nigiri::location_idx_t::invalid());
 
+std::vector<nigiri::location_idx_t> get_stops_with_unique_routes(
+    nigiri::timetable const&,
+    nigiri::rt_timetable const*,
+    point_rtree<nigiri::location_idx_t> const&,
+    osr::location const&,
+    double const,
+    hash_set<nigiri::route_idx_t>&);
 }  // namespace motis
--- a/include/motis/get_stops_with_traffic.h
+++ b/include/motis/get_stops_with_traffic.h
@@ -16,11 +16,4 @@ std::vector<nigiri::location_idx_t> get_stops_with_traffic(
     double const distance,
     nigiri::location_idx_t const not_equal_to =
         nigiri::location_idx_t::invalid());
-
-std::vector<nigiri::location_idx_t> get_stops_with_unique_routes(
-    nigiri::timetable const&,
-    nigiri::rt_timetable const*,
-    point_rtree<nigiri::location_idx_t> const&,
-    osr::location const&,
-    double const);
 }  // namespace motis
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -763,6 +763,18 @@ paths:
             default: 900
             minimum: 0
 
+        - name: minNearStations
+          in: query
+          required: false
+          description: |
+            Optional. Default is `1`.
+            Minimum number of stations to route from / to. Increase
+            `maxPostTransitTime` / `maxPreTransitTime` if less stops are found. Pass `0` to disable.
+          schema:
+            type: integer
+            default: 1
+            minimum: 0
+
         - name: maxDirectTime
           in: query
           required: false
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -272,15 +272,20 @@ std::vector<n::routing::offset> get_offsets(
         max,
         std::chrono::seconds{
             r.config_.get_limits().street_routing_max_near_stops_seconds_});
+    auto const max_beeline_meters =
+        get_max_distance(profile, osr_params, max_beeline_seconds);
 
-    while (expanded_time <= max_beeline_seconds &&
+    while (expanded_time < max_beeline_seconds &&
            near_stops.size() < min_near_stations * kBeelineNearStationsFactor) {
+
       near_stops =
           get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, expanded_dist);
 
-      expanded_dist *= 2;
+      expanded_dist = std::min(expanded_dist * 2, max_beeline_meters);
       expanded_time *= 2;
     }
+    expanded_time = std::min(expanded_time, max_beeline_seconds);
+
     auto const near_stop_locations = utl::to_vec(
         near_stops,
         [&](n::location_idx_t const l) { return stop_to_osr_location(r, l); });
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -261,27 +261,19 @@ std::vector<n::routing::offset> get_offsets(
       profile = osr::search_profile::kCarSharing;
     }
 
-    constexpr unsigned const FACTOR = 2U;
+    constexpr unsigned const kBeelineNearStationsFactor = 4U;
 
     auto near_stops = std::vector<n::location_idx_t>{};
     auto expanded_dist = get_max_distance(profile, osr_params, max);
     auto expanded_time = max;
-    auto prev_distance = 0U;
     for (unsigned i = 0U; i < max_bbox_increases; ++i) {
-      auto const stops =
+      near_stops =
           get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, expanded_dist);
-      std::copy_if(stops.begin(), stops.end(), std::back_inserter(near_stops),
-                   [&](auto const& l) {
-                     return geo::distance(pos.pos_,
-                                          r.tt_->locations_.coordinates_[l]) >
-                            prev_distance;
-                   });
 
-      if (near_stops.size() >= min_near_stations * FACTOR) {
+      if (near_stops.size() >= min_near_stations * kBeelineNearStationsFactor) {
         break;
       }
 
-      prev_distance = expanded_dist;
       expanded_dist *= 2;
       expanded_time *= 2;
     }
@@ -378,15 +370,13 @@ std::vector<n::routing::offset> get_offsets(
     utl::sort(mode_offsets, [](auto const& a, auto const& b) {
       return a.duration_ < b.duration_;
     });
-    auto N = min_near_stations;
-    for (auto const& a : mode_offsets) {
-      if (a.duration_.count() * 60 < max.count() + 60 || N > 0) {
-        --N;
-        offsets.push_back(a);
-      } else {
-        break;
-      }
-    }
+    auto const it = utl::find_if(mode_offsets, [&](auto const& a) {
+      return a.duration_.count() * 60 > max.count() + 60;
+    });
+    auto const n_found = std::max(
+        static_cast<unsigned>(std::distance(mode_offsets.begin(), it)) + 1U,
+        min_near_stations);
+    offsets.append_range(mode_offsets | std::views::take(n_found));
 
     stats.emplace(fmt::format("prepare_{}_{}", to_str(dir), fmt::streamed(m)),
                   UTL_GET_TIMING_MS(timer));
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -1,6 +1,7 @@
 #include "motis/endpoints/routing.h"
 
 #include <algorithm>
+#include <chrono>
 #include <cmath>
 #include <iterator>
 #include <utility>
@@ -15,6 +16,7 @@
 
 #include "utl/erase_duplicates.h"
 #include "utl/helpers/algorithm.h"
+#include "utl/logging.h"
 #include "utl/timing.h"
 
 #include "osr/lookup.h"
@@ -227,8 +229,7 @@ std::vector<n::routing::offset> get_offsets(
     double const max_matching_distance,
     gbfs::gbfs_routing_data& gbfs_rd,
     stats_map_t& stats,
-    unsigned const min_near_stations,
-    unsigned const max_bbox_increases) {
+    unsigned const min_near_stations) {
   if (!r.is_osr_loaded()) {
     return {};
   }
@@ -266,13 +267,15 @@ std::vector<n::routing::offset> get_offsets(
     auto near_stops = std::vector<n::location_idx_t>{};
     auto expanded_dist = get_max_distance(profile, osr_params, max);
     auto expanded_time = max;
-    for (unsigned i = 0U; i < max_bbox_increases; ++i) {
+
+    auto const max_beeline_seconds = std::max(
+        max, std::chrono::seconds{
+                 r.config_.limits_->street_routing_max_near_stops_seconds});
+
+    while (expanded_time <= max_beeline_seconds &&
+           near_stops.size() < min_near_stations * kBeelineNearStationsFactor) {
       near_stops =
           get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, expanded_dist);
-
-      if (near_stops.size() >= min_near_stations * kBeelineNearStationsFactor) {
-        break;
-      }
 
       expanded_dist *= 2;
       expanded_time *= 2;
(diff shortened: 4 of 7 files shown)


aff3npirat:
Why are we not searching for stops with traffic when using [radius](https://github.com/aff3npirat/motis/blob/ec6bc195fb327ef2f36ea5efea5572b0934db67c/src/endpoints/routing.cc#L898) compared to [here](https://github.com/aff3npirat/motis/blob/ec6bc195fb327ef2f36ea5efea5572b0934db67c/src/endpoints/routing.cc#L271), wouldnt it make sense to always look at stops with traffic?

And should I also increase the search radius when using `radius_offsets`?

src/endpoints/routing.cc
```diff
@@ -256,8 +261,23 @@ std::vector<n::routing::offset> get_offsets(
...
+    for (unsigned i = 0U; i < max_bbox_increases; ++i) {
+      utl::concat(near_stops, get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_,
+                                                     pos, expanded_dist));
+
+      if (near_stops.size() >= min_near_stations * FACTOR) {
+        break;
+      }
+
+      expanded_dist *= max_dist;
+      expanded_time *= max.count();
```
traines-source: this might be a bit too fast, I was thinking more like pow(2, i)*max_dist ? Unless there is insight that justifies a quicker increase...

And expanded_dist and expanded_time need to grow somehow in sync (actually expanded_time faster?) because otherwise we will find n stops in the rtree radius, but none/less in the expanded_time?

include/motis/endpoints/routing.h
```diff
@@ -69,7 +69,9 @@ struct routing {
       std::chrono::seconds max,
       double max_matching_distance,
       gbfs::gbfs_routing_data&,
-      stats_map_t& stats) const;
+      stats_map_t& stats,
+      unsigned const min_near_stations = 0U,
+      unsigned const max_bbox_increases = 5U) const;
```
traines-source: If we need to make this configurable, I think it should be based on (walking) time as well (even though I'm not sure which distance/config limit exactly, cf. issue).

src/endpoints/routing.cc
```diff
@@ -256,8 +261,23 @@ std::vector<n::routing::offset> get_offsets(
...
-    auto const near_stops =
-        get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, max_dist);
+
+    constexpr unsigned const FACTOR = 2U;
+
+    auto near_stops = std::vector<n::location_idx_t>{};
+    auto expanded_dist = max_dist;
+    auto expanded_time = max;
+    for (unsigned i = 0U; i < max_bbox_increases; ++i) {
+      utl::concat(near_stops, get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_,
```
traines-source: do we need to concat this? rtree will return all the same locations all over again already? maybe I'm missing sth

src/endpoints/routing.cc
```diff
@@ -337,14 +360,20 @@ std::vector<n::routing::offset> get_offsets(
...
               static_cast<n::transport_mode_id_t>(profile));
         }
       }
     }
 
+    utl::sort(mode_offsets, [&](auto const& a, auto const& b) {
+      // return trip or location based?
+      return a.duration_ < b.duration_;
+    });
+    utl::concat(offsets, mode_offsets | std::views::take(min_near_stations));
```
traines-source: we also want to take at least all that are contained in the max time?

openapi.yaml
```diff
@@ -763,6 +763,18 @@ paths:
             default: 900
             minimum: 0
 
+        - name: minNearStations
```
traines-source, line 766: I think you're right that a setting for both pre and post together suffices.

traines-source (commented):
I think the general approach makes sense, but there are some things I noticed

src/endpoints/routing.cc
traines-source: put differently, rn the rtree radius is an upper bound for the distance that could be covered within the max time, and now we need to find an upper bound for the max time needed to cover the rtree radius?

traines-source:
> Why are we not searching for stops with traffic when using [radius](https://github.com/aff3npirat/motis/blob/ec6bc195fb327ef2f36ea5efea5572b0934db67c/src/endpoints/routing.cc#L898) compared to [here](https://github.com/aff3npirat/motis/blob/ec6bc195fb327ef2f36ea5efea5572b0934db67c/src/endpoints/routing.cc#L271), wouldnt it make sense to always look at stops with traffic?

> And should I also increase the search radius when using radius_offsets?

This is a niche feature that some people wanted where we skip the OSR calculation entirely and just use beeline distances to discover nearby locations. Maybe it's not worth the effort to change this, although I think this could also benefit from the n-nearest feature.

aff3npirat:
> This is a niche feature that some people wanted where we skip the OSR calculation entirely and just use beeline distances to discover nearby locations. Maybe it's not worth the effort to change this, although I think this could also benefit from the n-nearest feature.

So I still dont get why we return locations that have no routes in `radius_offsets`?

And either way I think we could merge `radius_offsets` and `get_stops_with_traffic` into a single function (maybe provide a flag `traffic_only`) and implement the expanding logic there.

src/endpoints/routing.cc
aff3npirat: > And expanded_dist and expanded_time need to grow somehow in sync 

Yes, actually my implementation is wrong and distance outgrows time. 

> actually expanded_time faster?

So currently (before I touched anything) distance was simply `time * speed`, but I think it would make sense that time corresponds to a larger radius than distance since it represents walking time and not beeline. I could add a config option for that and do some research / testing later for a good default value.

include/motis/endpoints/routing.h
aff3npirat: Do you mean provide a config to set the maximum walking time it can expand to (replacing `max_bbox_increases` by time dependent config)?

src/endpoints/routing.cc
traines-source: Thinking about it again (and changing my mind) we actually want to use this exact same correlation between distance and time, because this way we can guarantee that we will find all stops that are actually in closest walking distance. (There is no upper bound for the walking time given the beeline distance, because a place that is 100 meter across a river may be 1 hour (or days) of walking away. But if our rtree was only 1km, we may miss a station that is just 2 kilometers/20 minutes away on the same side of the river. I think it is worse to find further away stations when there exist closer stations than it is to not always reach `n`.) In turn we may choose a slightly larger `FACTOR` (which we by the way normally would call `kFactor` or better `kBeelineNearStationsFactor` or sth like this). I think 4 would be a nicely justifiable value (because assuming the stations are distributed uniformely, doubling the radius should yield 4 times more stations, and so if we see `4n` stations and acceptable walking distances are not more than twice as long as beeline distances, we will find at least `n` stations within a walking time of `expanded_dist/speed`. What do you think?

include/motis/endpoints/routing.h
traines-source: I think we could abuse the existing config.yml `street_routing_max_direct_seconds` for that or introduce a separate `street_routing_max_prepost_transit_min_near_seconds`. EDIT: since we will need the same for transfers, "..._prepost_..." is not really good. So maybe `street_routing_max_near_stops_seconds`?
Even better would probably be to conversely limit GBFS and car routing etc. to a maximum number of locations (or GBFS elements) found, because that tends to make things expensive. But that is out of scope of this PR.

traines-source:
> So I still dont get why we return locations that have no routes in radius_offsets?

I don't think there is a good reason other than that in most cases it probably doesn't really make a difference performance wise or otherwise.

> And either way I think we could merge radius_offsets and get_stops_with_traffic into a single function (maybe provide a flag traffic_only) and implement the expanding logic there.

They are not really returning the same things (offsets vs location)? But feel free to refactor if you see fit.

src/endpoints/routing.cc
```diff
@@ -260,23 +261,29 @@ std::vector<n::routing::offset> get_offsets(
...
-      utl::concat(near_stops, get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_,
-                                                     pos, expanded_dist));
+      auto const stops =
+          get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, expanded_dist);
+      std::copy_if(stops.begin(), stops.end(), std::back_inserter(near_stops),
+                   [&](auto const& l) {
+                     return geo::distance(pos.pos_,
+                                          r.tt_->locations_.coordinates_[l]) >
+                            prev_distance;
+                   });
```
traines-source: Isn't this needlessly complicated? Is there any drawback to just overwrite the vector?

src/endpoints/routing.cc
```diff
@@ -368,11 +375,18 @@ std::vector<n::routing::offset> get_offsets(
...
-    utl::concat(offsets, mode_offsets | std::views::take(min_near_stations));
+    auto N = min_near_stations;
+    for (auto const& a : mode_offsets) {
+      if (a.duration_.count() * 60 < max.count() + 60 || N > 0) {
+        --N;
+        offsets.push_back(a);
+      } else {
+        break;
+      }
+    }
```
traines-source: couldn't we do this easier and in situ with a loop starting at `n` that continues as long as `mode_offsets[i].duration_.count() * 60 < max.count() + 60` and then does a `mode_offsets.resize(i)`?

mority:
btw, this is the notorious repertoire filter idea: https://github.com/motis-project/nigiri/compare/master...mority:nigiri:repertoire-filter

include/motis/get_stops_with_traffic.h
```diff
@@ -15,5 +16,4 @@ std::vector<nigiri::location_idx_t> get_stops_with_traffic(
     double const distance,
     nigiri::location_idx_t const not_equal_to =
         nigiri::location_idx_t::invalid());
-
```
felixguendling: keep

src/endpoints/routing.cc
```diff
@@ -255,9 +263,31 @@ std::vector<n::routing::offset> get_offsets(
       profile = osr::search_profile::kCarSharing;
     }
 
-    auto const max_dist = get_max_distance(profile, osr_params, max);
-    auto const near_stops =
-        get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, max_dist);
+    constexpr unsigned const kBeelineNearStationsFactor = 4U;
+
+    auto const max_beeline_seconds = std::max(
```
felixguendling, line 268: max? this will go beyond `street_routing_max_near_stops_seconds_`

shouldn't we throw 400 so the user knows that parameters were bad?

src/endpoints/routing.cc
```diff
@@ -255,9 +263,31 @@ std::vector<n::routing::offset> get_offsets(
...
+        get_max_distance(profile, osr_params, max_beeline_seconds);
+
+    auto expanded_time = max;
+    auto expanded_dist = get_max_distance(profile, osr_params, expanded_time);
+
+    auto near_stops =
+        get_stops_with_traffic(*r.tt_, rtt, *r.loc_tree_, pos, expanded_dist);
+    while (expanded_time * 2 <= max_beeline_seconds &&
+           near_stops.size() < min_near_stations * kBeelineNearStationsFactor) {
+
```
felixguendling, line 282: no new line loop start
```suggestion
```

openapi.yaml
felixguendling, line 766: inconsistent - sometimes stops, sometimes stations
-> always stops

openapi.yaml
```diff
@@ -763,6 +763,18 @@ paths:
...
+          in: query
+          required: false
+          description: |
```
felixguendling, line 769: description doesn't help me to understand what's actually happening

src/endpoints/routing.cc
```diff
@@ -222,7 +229,8 @@ std::vector<n::routing::offset> get_offsets(
     std::chrono::seconds const max,
     double const max_matching_distance,
     gbfs::gbfs_routing_data& gbfs_rd,
-    stats_map_t& stats) {
+    stats_map_t& stats,
+    unsigned const min_near_stations) {
```
felixguendling, line 233: keep stats last, add close to `max` param

felixguendling (commented):
implementation doesn't feel clean to me
it seems to also not guarantee the same behavior as before if the parameter is not set?

not sure if inside get_offsets is the right place to implement this

src/get_stops_with_traffic.cc
```diff
@@ -1,9 +1,12 @@
 #include "motis/get_stops_with_traffic.h"
 
 #include "osr/location.h"
+#include "utl/helpers/algorithm.h"
```
felixguendling, line 4: header grouping

PR #1498, opened 2026-06-23
use gzip instead of deflate for MOTIS tiles

Findus23:
requires https://github.com/motis-project/tiles/pull/19

Also adds Not Implemented response if client doesn't accept gzip.


felixguendling:
can you please update the tiles commit in .pkg and run 1x pkg -l so it updates .pkg.lock?

PR #1503, opened 2026-06-25
ui: Show times in page title and use shorter place names

krkk:
Continuation of/context: https://github.com/motis-project/motis/pull/1489#discussion_r3472754842


traines-source:
Thanks!

PR #1506, opened 2026-07-02
Remove confusing instructions

thomas-zahner:
Since ca3e3f7ced18e7f49229b53a0999fc2a782aa4bd these files are symlinked in the build process already. There is no need to do such manual modifications any longer.

Noticed this while trying to set up motis for the first time and following these docs.


traines-source:
Thank you! I hope these symlinks work on Windows as well...

PR #1515, opened 2026-07-08
ui: Don't flip the order of earlier alternatives after selecting one

krkk:
The server returns a list in which the first alternative connection is the closest to the original time.

However, after selecting an earlier alternative connection, the list would get sorted lexicographically, reversing the list.

[recording of a bug](https://github.com/user-attachments/assets/743a4d4d-8e2b-4b18-a9b9-01c16c3e49e1)


felixguendling:
Nice!

Issue #1403, opened 2026-05-13
Transfer and Checkin Times


traines-source:
An attempt to distill all the needs we have that are remotely related to transfer and checkin times.

### Requirements
1. support for {stop_id,trip_id,route_id,\*}x{stop_id,trip_id,route_id,*} transfers as per the [GTFS spec](https://gtfs.org/documentation/schedule/reference/#transferstxt)
2. support for Netex transfers modeling?
3. support for checkin+checkout times for flights by route_type and/or agency and/or feed, possibly including shorter transfers between flights
4. support for checkin times for Eurostar etc. by agency and/or feed and/or route_id
5. support for longer transfer/checkin/checkout times before, after, but also between e.g. Flixbuses, by agency and/or route_type
6. support for 2 minute compulsory checkin time for French trains by agency and/or feed
7. R3, R4, R6 and possibly R5 also need to be applied for pre/post transit and should not be overriden by routed transfers
8. support for the Paris use case: despite local transport available, enforce minimum transfer time between two Paris train stations by stop_id,  but not when coming from pretransit/local transport
9. operators want to enforce minimum transfer times even with routed transfers (cf. https://github.com/motis-project/nigiri/issues/339), also for same-platform transfers (currently not profile-specific)
10. in some cases on the other hand one might want to overwrite too long transfers given in the feed with routed transfers (cf. https://github.com/motis-project/motis/issues/1324), also for same-platform transfers (currently not profile-specific) 
11. in any case routed transfers should fall back to feed-provided transfers if for a certain relation no routed transfer was found due to mapping errors or the distance limit (cf. https://github.com/motis-project/motis/issues/1401)
12. operators want to define guaranteed transfers up to a maximum waiting time (ensuring the connecting trip will wait in the case of delays, without propagating the delay, because in the end the connecting trip might not wait if the waiting time is exceeded because the incoming trip is delayed too much) by trip_id and/or route_id (cf. transfer_type=1)
13. all of this should also be possible to set via Lua scripts
14. (additional transfer time for agency changes that are likely to entail a break in passenger rights)
15. (additional transfer time for high frequency -> low frequency transfers)
16. (inter-feed transfers)

### Implications
[...]

PR #1526, opened 2026-07-16
make routed transfers the default

mority:
"When calculating an itinerary, GTFS-consuming applications interpolate transfers based on allowable time and stop proximity."
https://gtfs.org/documentation/schedule/reference/#transferstxt

Since routing on OSM data is a much better approximation than beeline distance, we should make it the default and use it if its available?


traines-source:
I agree but I think we should maybe improve the fallback mechanisms first? (esp. `min(routed_transfer, feed_transfer)`, also cf. the table i proposed in https://github.com/motis-project/motis/issues/1403)

openapi.yaml
```diff
@@ -270,12 +270,12 @@ paths:
...
           required: false
           description: |
-            Optional. Default is `false`.
+            Optional. Default is `true`.
             
             Whether to use transfers routed on OpenStreetMap data.
           schema:
             type: boolean
-            default: false
+            default: true
```
felixguendling, line 355: defaults also have to be reflected in the UI code

felixguendling:
> I agree but I think we should maybe improve the fallback mechanisms first?

Yes, we should probably only fall back to transfers that really came from a `transfers.txt` and not to our own beeline transfers. Especially the 300m inter-dataset distance is quite high and already leads to problems. I've not seen any problems with 100m, so maybe those (or maybe < 50m?) can still be used in case osr has no route because of missing OSM data (probably not needed in Germany, but not all countries have proper OSM data).

PR #1527, opened 2026-07-16
parallel query generation

mority:


exe/generate.cc
```diff
@@ -304,162 +305,194 @@ int generate(int ac, char** av) {
...
+  auto mutex = std::mutex{};
+  utl::parallel_for(
+      ranks,
+      [&](auto const r) {
+        thread_local auto p = master_params;
+        thread_local auto stops = master_stops;
+        thread_local auto ss = std::optional<n::routing::search_state>{};
+        thread_local auto rs = std::optional<n::routing::raptor_state>{};
+        thread_local auto geo_distance =
+            std::unordered_map<n::location_idx_t, double>{};
```
felixguendling:
```suggestion
            hash_map<n::location_idx_t, double>{};
```

std::unordered_set is a no-go

even if this is not performance sensitive, it's better to never use it ever

exe/generate.cc
```diff
@@ -304,162 +305,194 @@ int generate(int ac, char** av) {
...
+  auto out = std::ofstream{"queries.txt"};
+  auto const progress_tracker =
+      utl::activate_progress_tracker(fmt::format("generating {} queries", n));
+  progress_tracker->in_high(n);
+  auto const silencer = utl::global_progress_bars{false};
+  auto mutex = std::mutex{};
+  utl::parallel_for(
+      ranks,
+      [&](auto const r) {
+        thread_local auto p = master_params;
```
felixguendling: `thread_local` can have unpredictable perf on different hardware

-> utl::parallel_for_thread_local

exe/generate.cc
```diff
@@ -304,162 +305,186 @@ int generate(int ac, char** av) {
...
-      random_from_to(r);
-      random_time();
-      out << p.to_url("/api/v1/plan") << "\n";
-      progress_tracker->increment();
-    }
-  }
+    random_from_to();
+    random_time();
+
+    auto guard = std::lock_guard{mutex};
```
felixguendling, line 486: this looks like a job for parallel for with ordered collect?

mority: hm, yeah. I guess, we could just return generated `plan_params` objects for the collector to write to the file in batches.

mority: Ordered collect seems to be slower (second measurement). Maybe because we don't need order? I mean, we do not really care on which line of the output file each randomized query ends up, do we?
We could add an unordered collect to utl, though.
```
mority@turing:~/code/motis/cmake-build-release-gcc-13$ ./motis generate -n 1000 -m WALK
Timetable ---
n_locations: 496176
n_routes: 281580
n_trips: 2035888
---
date range: [2026-03-09 00:00:00, 2026-03-23 00:00:00], tt=[2026-03-09 00:00, 2027-03-10 00:00[
modes: WALK
from and to pairings by lower bounds rank
2026-07-28T15:22:51Z [info] [timer.cc:9] [scoped_timer] [generate queries] starting
generating 1000 queries: [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ ] 100%
2026-07-28T15:22:53Z [info] [timer.cc:18] [scoped_timer] [generate queries] finished (1888.107ms)
mority@turing:~/code/motis/cmake-build-release-gcc-13$ ./motis generate -n 1000 -m WALK
Timetable ---
n_locations: 496176
n_routes: 281580
n_trips: 2035888
---
date range: [2026-03-09 00:00:00, 2026-03-23 00:00:00], tt=[2026-03-09 00:00, 2027-03-10 00:00[
modes: WALK
from and to pairings by lower bounds rank
2026-07-28T16:38:10Z [info] [timer.cc:9] [scoped_timer] [generate queries] starting
generating 1000 queries: [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ ] 100%
2026-07-28T16:38:13Z [info] [timer.cc:18] [scoped_timer] [generate queries] finished (2097.238ms)
```

felixguendling: yes. I think we are not able to produce the same file with repeated calls anyway, right? nigiri-benchmark has this concept but here it won't work like this (2x motis generate -> 2x same query set) but probably it's not needed anyway because we write the queries to a file unlike nigiri-benchmark where they are executed immediately

mority: For single-threaded, we produce the same queries for the same timetable, OSM data & program options. But for multi-threaded, we would have to make sure that the calls to rand_in are in the same order every time. Otherwise the threads will race for the rand_in calls and it will not be reproducible. 

I agree that it is not really a problem since we can just store the query file when reproducibility is required.

mority:
`./motis generate -n 1000 -m WALK` on DELFI GTFS and Germany OSM
|    [ms]     | mutex    | ordered_collect  |
| ------- | -------- | ---------------- |
|         | 3862     | 3889             |
|         | 3810     | 3929             |
|         | 3879     | 3931             |
|         | 3983     | 3914             |
|         | 3841     | 3883             |
|         | 3830     | 3936             |
|         | 3869     | 4004             |
|         | 3865     | 3919             |
|         | 3836     | 3890             |
|         | 3836     | 3934             |
| **AVG** | **3861** | **3923** (+1.6%) |

Issue #1484, opened 2026-06-16
Abort HTTP requests


traines-source:
* UI: abort in-flight plan requests when a new one is being sent due to changed options (currently multiple plan requests stay in flight concurrently -> unnecessary server resource consumption)
* server: abort calculations when client has closed HTTP connection/aborted the request

(not sure how complicated it is, but the benefit would not only be lower resource consumption, but also enable better quota handling per client, because then concurrent connections per client, which usually hint at abusive behavior, can be limited in a reverse proxy)

PR #1531, opened 2026-07-17
UI: search button

mority:
played some tetris with the search mask, button does not do anything yet

looks huge on desktop
image

but this way it fits into one row on mobile
Screenshot_20260717-222833

### DIFF
--- a/ui/src/lib/AdvancedOptions.svelte
+++ b/ui/src/lib/AdvancedOptions.svelte
@@ -28,6 +28,7 @@
 	import { generateTimes } from './generateTimes';
 	import ViaStopOptions from './ViaStopOptions.svelte';
 	import Slider from './components/ui/slider/Slider.svelte';
+	import { Settings } from '@lucide/svelte';
 	let {
 		advancedOptionsOpen = $bindable(false),
 		useRoutedTransfers = $bindable(),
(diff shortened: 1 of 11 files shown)


felixguendling:
Maybe it could have its own row? Not sure how that would look like.

mority:
> Maybe it could have its own row? Not sure how that would look like.

I tried that at first, but then we would have to choose between a huge 'Options' button or some empty space in the row above. Or make it more responsive and move some options out of the dialog directly into the search mask for large viewports.

My solution was to replace 'Options' with the settings wheel to make it fit into one row.

mority:
I put everything into one flex row again and just let the responsive design handle wrapping and growing

image

image

image

traines-source:
Frankly I'm still in favor of just not having a search button. The time picker is not really causing that many search triggers usually and that would IMO be the only reason why one would want a search button now.

mority:
yeah, but why waste resources at all?

traines-source:
If we want to avoid wasting resources principally we should properly [abort requests](https://github.com/motis-project/motis/issues/1484) I think. This is something that is relevant for all clients and not just ours.

felixguendling:
> time picker is not really causing that many search triggers usually and that would IMO be the only reason why one would want a search button

I think it's quite standard to fill the form from top to bottom and once you have filled start and destination it starts searching, even if you wanted to still fill the time. So basically every user searching not "Now" will trigger an additional search. That's why I like how DB Navigator does it: the first time you have to press the search button but after that it will search when changing any input.

Aborting requests can be done additionally.

PR #1538, opened 2026-07-24
sort offsets

traines-source:
see https://github.com/motis-project/nigiri/pull/387


felixguendling:
I moved it to `nigiri::routing::query::sanitize()` (merged to nigiri master + motis master).

PR #1547, opened 2026-07-28
default crow-fly duration for radius offsets

orhazal:
Hello !

Follow-up to #1360 and this [comment](https://github.com/motis-project/motis/pull/1360#issuecomment-4182494340) : offsets created by the radius parameter previously had a zero duration, so a stop 50 m from the coordinate and one 20 km away were equally attractive to the router : journeys could start/end at arbitrarily distant stops and the reported start/end times didn't account for reaching them. This PR gives each radius offset a duration derived from the crow-fly distance to the stop, at walking speed (1.5 m/s, kWalkSpeed) by default. Since the appropriate access speed depends on the deployment (walking vs. some faster feeder mode further out), it can optionally be configured per distance band with a new timetable.radius_speeds config option (e.g. 2 m/s up to 5 km, faster beyond); when unset, behavior is plain walking speed. The config is validated on startup (ascending bands, positive speeds).

Since the offset legs now have a non-zero duration, they are no longer removed as zero-duration legs in journey_to_response, which would dereference the OSR data unconditionally. With no OSM data loaded (the main use case for radius), this crashed. Offset legs now fall back to a plain crow-fly leg when street routing data is unavailable, mirroring the existing footpath fallback. Includes tests for the speed bands, config parsing/validation, and an end-to-end routing test without OSM data, plus documentation updates in docs/setup.md and openapi.yaml.

The speed bands are one of many possible ways to configure this. I chose this shape to match our use case (legacy matching oblige), but I'm open to other options. Thank you :)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -77,7 +77,9 @@ paths:
           description: |
             Experimental. Search radius in meters around the `fromPlace` / `toPlace` coordinates.
             When set and the place is given as coordinates, all transit stops within
-            this radius are used as start/end points with zero pre-transit/post-transit time.
+            this radius are used as start/end points. The pre-transit/post-transit time
+            for each of them is estimated from the crow-fly distance to the coordinate
+            at walking speed (1.5 m/s), rounded down to whole minutes.
             Works without OSM/street routing data loaded.
           schema:
             type: number
--- a/src/endpoints/routing.cc
+++ b/src/endpoints/routing.cc
@@ -26,6 +26,7 @@
 #include "osr/types.h"
 
 #include "nigiri/common/interval.h"
+#include "nigiri/constants.h"
 #include "nigiri/location_match_mode.h"
 #include "nigiri/routing/limits.h"
 #include "nigiri/routing/pareto_set.h"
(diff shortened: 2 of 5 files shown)


felixguendling:
This logic looks extremely specific to your use case and not like something anyone else would use. Our team is too small and not well funded to maintain this kind of code for you in the long run.

But I would be open for a more generic version where you can pass any set of offsets (separate start + dest) you want. This would not just cover your specific use case but would also allow users to retrieve offsets from their favorite specialized street routing engine and use them for first+last mile in MOTIS' public transit routing. Also in this case, changes to your business logic (next you need not stepwise but linear or other functions) will not require changes to MOTIS (advantage: you don't have to wait for a new release). It would require a POST endpoint as `[(stop, offset), ...]` lists can become very long. So in order not to copy the next endpoint we want to have as GET+POST we would need to find a way in openapi-cpp to reuse properties for both.

orhazal:
Thanks for the feedback, makes sense ! 

Client-supplied offsets are better for us too: our access-time model changes with business rules, and computing offsets on our side removes MOTIS releases from that loop entirely indeed. I'd be happy to work on this. I'll start by looking at what openapi-cpp needs to share the parameter definitions between the GET and a POST variant of /plan, and open a draft PR / discussion once I understand the scope.

For the existing radius, the crow-fly at walk speed default is welcome as a fix for offsets cost or it's not worth it?

felixguendling:
> For the existing radius, the crow-fly at walk speed default is welcome as a fix for offsets cost or it's not worth it?

With one hardcoded parameter, it might be a helpful tool to quickly do a evaluation without having to load OSM.

> I'll start by looking at what openapi-cpp needs to share the parameter definitions between the GET and a POST variant of /plan, and open a draft PR / discussion once I understand the scope.

Sounds great! You could also start with `OneToManyParams` which was duplicated for the same reason (GET+POST).

In general, our goal is that everyone uses `osr` for offsets. If it's not fast enough, the best option would be to make it faster with something like BucketCH, PHAST, etc. Then there's no real reason to use radius.

orhazal:
> With one hardcoded parameter, it might be a helpful tool to quickly do a evaluation without having to load OSM.

I updated with this change.

orhazal:
There is a failure on GPU Validation, _apparently, it comes from exe/generate.cc_

felixguendling:
I fixed it on master yesterday I think.

PR #1548, opened 2026-07-29
reservation

traines-source:


src/journey_to_response.cc
```diff
@@ -713,12 +718,16 @@ api::Itinerary journey_to_response(
                             ? std::optional{tt.src_end_date_[src]}
                             : std::nullopt,
                     .bikesAllowed_ =
-                        enter_stop.bikes_allowed(nigiri::event_type::kDep),
+                        check_flag_enter_exit(nigiri::kBikesAllowed),
                     .wheelchairAccessible_ =
-                        enter_stop.wheelchair_accessible(
-                            nigiri::event_type::kDep)
```
traines-source, line 719: I think it's better to at least check whether it's allowed at both enter and exit stop?

PR #1551, opened 2026-07-30
tweaks and cleanup

traines-source:


src/endpoints/stop_times.cc
```diff
@@ -634,6 +644,20 @@ api::stoptimes_response stop_times::operator()(
                                    : api::PickupDropoffTypeEnum::NOT_ALLOWED,
                 .cancelled_ = stop_cancelled,
                 .tripCancelled_ = run_cancelled,
+                .loopedCalendarSince_ =
+                    (fr.is_scheduled() && src != n::source_idx_t::invalid() &&
```
felixguendling: this is 2x checking the same?
`fr.is_scheduled() <=> src != n::source_idx_t::invalid()`

PR #1554, opened 2026-07-31
limit max matching distance

felixguendling:


src/odm/prima.cc
```diff
@@ -1,5 +1,6 @@
 #include "motis/odm/prima.h"
 
+#include <algorithm>
```
traines-source, line 3: why these imports everywhere?

traines-source (approved):

src/odm/prima.cc
felixguendling: std::min comes from `<algorithm>`
https://en.cppreference.com/cpp/algorithm/min

I guess it's included somewhere else already but not sure if that's the best strategy because if those places stop including it, the build will break in all kinds of unrelated places.

PR #1556, opened 2026-08-01
feat(rt): add shadow GPS departure prediction diagnostics

piotrski:
## What changed

- publishes bounded vehicle observation history with realtime snapshots
- projects matched vehicles onto trip shapes and builds whole-trip ETA candidates in raw seconds
- compares provider, GPS, and schedule candidates with stateful hysteresis while remaining shadow-only
- exposes an opt-in `includePredictionComparison` stop-times contract with sparse `predictionDebug` entries
- publishes immutable diagnostics atomically with the realtime snapshot and supports stop, station, and nearby query shapes
- fixes Nigiri boundary-event handling for first/last stops and sliced runs

## Why

Provider trip updates can materially disagree with vehicle GPS progress. This provides the measurement and debug path needed to compare both sources without changing production departure timings or the default API response.

## Behavior

The default response is unchanged. Diagnostics are returned only when `includePredictionComparison=true`. Until R3.6 calibration supplies selector policies, selection reports `POLICY_UNAVAILABLE` while exposing provider/GPS candidates and the effective provider timing; operational timings are not mutated.

## Validation

- focused prediction/store/stoptimes suite: 21/21 passed
- complete MOTIS suite: 181/181 passed
- deterministic estimator crash reproducer: 3/3 passed after boundary fix
- `git diff --check` passed


felixguendling:
Please discuss features, bugfixes, etc. **before** opening a PR (Issue or Discussion).

We can neither review nor maintain 10k LoC patches (even if it would be a feature we need).

If you think specific changes from your PR are still useful even if the main functionality can't be merged, you can open an issue to discuss if it's needed.

PR #1557, opened 2026-08-03
fix 500s

traines-source:


ton-An:
@traines-source I was about to look into the "departure is last stop in trip", also noticed that it was still there in the newest release. Thanks! : )

PR #1568, opened 2026-08-07
ODM pong

mority:


mority:
**Correctness**
Results between rRAPTOR and Pong do not match 100%, but the differences stem from the heuristic shortening which reduces ODM time for the price of longer PT time. rRAPTOR and Pong can find different journeys that are equivalent regarding the optimization criteria. However, the heuristic shortening then gives different results on these journeys. It mostly results 1-2 min differences in travel time but no meaningful alternatives are lost.

**Performance**
Pong is ~15% faster end-to-end. The ODM-PT routing part alone gets ~1.6x speedup by switching to Pong.

Caveat: only tested with arriveBy=false

PR #1575, opened 2026-08-16
Add /route-search endpoint

magdhelena:
This PR adds a new endpoint, `experimental/map/route-search`, to search for routes (transit lines) by their name. Currently, the existing `experimental/map/routes` endpoint returns routes in a given area frame but one cannot just simply search for a line given its name.

The endpoint takes both `place` and `placeBias` (akin to `v1/geocode`), in order to give more weight to closer lines.
It's aimed at looking for a route when one does not know the name of a given stop or the exact shape to pinpoint it via a map. It implies the subsequent usage of `experimental/map/route-details` to look up any of the route indexes for a given line. For this reason the endpoint doesn't return geometry data.

### Parameters
| Name | Type | Required | Default | Description |
|------|------|----------|-----|------------|
| `text` | `string` | yes | - | The (potentially partially typed) route name to search for |
| `place` | `string` | no | - |`latitude,longitude` pair used for biasing results towards the coordinate |
| `placeBias` | `number` | no | 1 |  Used for biasing results towards the coordinate. Higher number = higher bias. 0 disables the bias |
| `numResults` | `integer` | no | 10 | Number of transit routes to return. Capped by `route_search_max_results` |
| `language` | `string[]` | no | - | Language preference for translated names (BCP-47, e.g. `en`, `fr`, `de`) |

### Response
A JSON with a single field, `routes`, consisting of an array of matches, ordered by score. Each match has the following structure:

- mode: `Mode`
- transitRoute: `TransitRouteInfo`
- agencyName: `string` (optional)
- routeIndexes: `integer[]`
- score: `number`

### Example
```http
GET /api/experimental/map/route-search?text=v23
GET /api/experimental/map/route-search?text=L9&place=41.3874,2.1686&placeBias=2
GET /api/experimental/map/route-search?text=catalunya&place=41.3874,2.1686&placeBias=2&numResults=5
```

### Config 
Adds a route_search_max_results limit (default 512), bounding numResults, following the existing geocode_max_suggestions pattern.

The reason to add it to experimental is that it's related and implies the usage of two other experimental endpoints. Moreover, the weights for the search score should probably be refined and the existing ones are just what "felt right".

### DIFF
--- a/src/endpoints/map/route_search.cc
+++ b/src/endpoints/map/route_search.cc
@@ -1,0 +1,265 @@
+#include "motis/endpoints/map/route_search.h"
+
+#include <algorithm>
+#include <cmath>
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "net/bad_request_exception.h"
+
+#include "utl/enumerate.h"
+#include "utl/erase_duplicates.h"
+#include "utl/to_vec.h"
+#include "utl/verify.h"
+
+#include "geo/box.h"
+#include "geo/latlng.h"
+
+#include "adr/normalize.h"
+
+#include "nigiri/shapes_storage.h"
+#include "nigiri/timetable.h"
+#include "nigiri/types.h"
+
+#include "motis/config.h"
+#include "motis/parse_location.h"
+#include "motis/timetable/clasz_to_mode.h"
+
+namespace n = nigiri;
+
+namespace motis::ep {
+
+namespace {
+
+constexpr auto const kDefaultNumResults = std::int64_t{10};
+constexpr auto const kApiVersion = 5U;
+
+constexpr auto const kExactScore = 300.F;
+constexpr auto const kPrefixScore = 200.F;
+constexpr auto const kSubstringScore = 100.F;
+constexpr auto const kLengthBonus = 50.F;
+constexpr auto const kLongNamePenalty = 25.F;
+constexpr auto const kDistancePenalty = 20.F;
+
+std::optional<float> match(std::string_view const name,
+                           std::string_view const text) {
+  if (name.empty()) {
+    return std::nullopt;
+  }
+
+  auto const base = [&]() -> std::optional<float> {
+    if (name == text) {
+      return kExactScore;
+    } else if (name.starts_with(text)) {
+      return kPrefixScore;
+    } else if (name.find(text) != std::string_view::npos) {
+      return kSubstringScore;
+    }
+    return std::nullopt;
+  }();
+
+  return base.transform([&](float const b) {
+    auto const coverage =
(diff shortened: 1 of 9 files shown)


src/endpoints/map/route_search.cc
```diff
@@ -0,0 +1,265 @@
...
+      candidates, begin(candidates) + static_cast<std::ptrdiff_t>(n), better);
+  candidates.resize(n);
+}
+
+std::vector<candidate> collect_candidates(n::timetable const& tt,
+                                          std::string_view const text,
+                                          n::lang_t const& lang,
+                                          adr::utf8_normalize_buf_t& buf) {
+  auto candidates = std::vector<candidate>{};
+  for (auto const [s, rids] : utl::enumerate(tt.route_ids_)) {
```
felixguendling, line 165: For a dataset like Transitous, going over all `route_ids` unfiltered and scoring them sounds like an expensive operation?

felixguendling:
@HerrLevin @MrKrisKrisu you were looking for a similar functionality for Träwelling but based on `trip.displayName`, correct? I think this would be the better solution because it would also cover long distance transport where route names are meaningless for users and you would rather search for train numbers like ICE 1234?

src/endpoints/map/route_search.cc
magdhelena: I considered this might be a problem but i'm not sure how to solve it. I expected this issue would arise; but I considered the people more familiar with MOTIS and Transitous would come up with better workarounds than me.
For example, I don't know how feasable generating an index of normalized route names at import time is.

felixguendling: I'm not sure what kind of matching we need. Usually line names are short, so except for whitespace (e.g. "S 6" vs "S6") which can be eliminated with a normalization that strips it (for query phase + preprocessing), I would say exact matches are required? This would enable us to use a [`string_store<trip_name_idx_t>`](https://github.com/motis-project/nigiri/blob/master/include/nigiri/string_store.h) and a `vecvec<trip_name_idx_t, trip_id_idx_t>` or similar for lookup.

HerrLevin:
Yes, this is almost how I've imagined it. What I'm kinda missing is the ability to search for the trip identifier rather than the route identifier (e.g. 19007 for the RE1)

Good for personnel operating the train, wanting to track it without supplying/filtering through departure times with one route identifier.

But filtering through long distance trains is already a huge plus! :)


image

src/endpoints/map/route_search.cc
magdhelena: I guess dropping the ability to search for `longName` is no big deal. At least having a quick look at Barcelona's data, that field tends to be origin/destination of the line (I don't know how it is for other places) and no one conceives of a  line through that, just the short name. Also, pattern matching can be useful for `shortName` (e.g L9N and L9S of Barcelona's metro) but is not indispensable for this endpoint to be useful and the permormance gains are well worth it.

PR #1576, opened 2026-08-17
ui: Don't steal Home and End keypresses from the address inputs

### LINKED ISSUE #1263: Webinterface: Pressing pos1 in the from / to field does nothing

krkk:
a hacky solution, bit it works. ¯\\_(ツ)_/¯

Fixes: #1263


felixguendling:
Yes, indeed very hacky :D
Maybe `bits/select` was not the best starting point in the first place.

PR #1577, opened 2026-08-17
Left-to-right text support in UI

orhazal:
While experimenting with MOTIS on Morocco, I noticed Arabic map labels render with reversed, disconnected letters. This is MapLibre's default behavior without an Right-To-Left (RTL) text plugin. This PR registers @mapbox/mapbox-gl-rtl-text (bundled locally, no CDN; lazy-loaded only when RTL text appears). Pinned to 0.2.3 since 0.4.0's exports blocks importing the worker script.

Before :
{3A3A5CE2-0310-496E-BD2A-BCAD23B8A43D}

After :
{AAB9B890-2603-4ACF-A599-BF23ED1D96F9}

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -64,6 +64,7 @@
 		"@deck.gl/mapbox": "~9.2.11",
 		"@hey-api/client-fetch": "^0.4.4",
 		"@lucide/svelte": "^0.548.0",
+		"@mapbox/mapbox-gl-rtl-text": "0.2.3",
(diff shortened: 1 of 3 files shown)


felixguendling:
Thanks! :)

Do we need more changes to properly support RTL? Feel free to open another PR if you find more problems.

orhazal:
From what i tested, Arabic is OK. I will check other RTL languages (japanese, hebrew... ) in Transitous when it will be released and adapt if necessary.

PR #1579, opened 2026-08-19
gbfs: read num_docks_available, the conditionally required free dock count

alexis-morain:
## Problem

`load_station_status()` only reads `vehicle_docks_available` (free docks broken down per vehicle type). Feeds that publish only the plain `num_docks_available` total lose their free dock count entirely.

Per the GBFS spec this is backwards. `num_docks_available` is conditionally REQUIRED ("REQUIRED except for stations that have unlimited docking capacity"), while `vehicle_docks_available` is only REQUIRED "in feeds where vehicle_types.json is defined and certain docks accept only specific vehicle types". A docked network that does not restrict docks by vehicle type is conformant when it publishes the total alone, so the parser is currently reading the optional refinement and ignoring the required field.

This is not an edge case. I could not find a single deployed feed on Transitous that publishes the typed variant. Checked directly against the live feeds:

| feed | stations | `vehicle_docks_available` | `num_docks_available` |
|---|---|---|---|
| Velib Paris | 1519 | 0 | 1519 |
| WienMobil Rad (`nextbike_wr`) | 267 | 0 | 267 |
| nextbike `bn` | 1050 | 0 | 1050 |

`/api/v1/rentals` accordingly returns `vehicleDocksAvailable: {}` for every station of these providers, in Paris, Vienna and Berlin alike.

Two consequences:

1. **Routing**: the return checks in `station_relevant_for_product()` (update.cc) and `osr_mapping::map_stations()` fall through to the bare `is_returning` flag, so a full station still advertising `is_returning: true` is proposed as a drop-off target. Measured on the live feeds: 24 such stations in Paris, 62 in Vienna (e.g. Julius-Raab-Platz, capacity 17, 32 bikes parked, 0 free docks).
2. **Display**: clients cannot show free dock counts for any of these feeds.

## Change

* `station_status` gains `std::optional<unsigned> num_docks_available_`, parsed from `num_docks_available` (same field name in GBFS 2.x and 3.x).
* The two return checks use it as a fallback **only when `vehicle_docks_available` is absent**. Typed counts keep precedence.
* `station_information` gains `std::optional<unsigned> capacity_`, parsed from `capacity`.
* Both are exposed on `RentalStation` as optional `numDocksAvailable` and `capacity`.

## Why the capacity guard matters

Virtual and free-floating stations legitimately report `num_docks_available: 0` while still accepting returns, so treating 0 as "full" would wrongly block them. `capacity` separates the two cases: in WienMobil Rad, 68 of the 130 stations reporting 0 free docks declare no capacity at all, and stay returnable. The remaining 62 declare physical docks and are genuinely over capacity.

One deliberate limitation: `capacity` is OPTIONAL in `station_information.json`, so a docked station that omits it will not get the new return check. That keeps the change strictly non-regressive, at the cost of not covering feeds that publish dock counts without declaring capacity.

`std::optional` likewise keeps "feed does not publish dock counts" distinct from "0 free docks", so providers publishing neither field keep the current behavior.

## Testing

* Parser test `gbfs_parser_station_status_total_docks_fallback` covers: total-only full station, total-only with free docks, no dock info at all (`nullopt`), and a capacity-less station reporting 0 docks.
* Full suite green locally, 170/170 (macOS arm64, Apple clang 21, Release).
* clang-format clean on all touched files.

Happy to adjust the guard if you would rather key it on something other than `capacity`, or to split the API additions into a separate PR.

Disclosure: this change was written with AI assistance (Claude). The measurements above were taken against the live feeds, and the code was built and tested locally before submitting.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -5086,14 +5086,13 @@ components:
         capacity:
           type: integer
           description: |
-            Number of physical docks at this station. Absent for virtual or
-            free-floating stations that do not have docks.
+            Number of docks or parking spots at this station.
         numDocksAvailable:
           type: integer
           description: |
             Total number of free docks at this station, for feeds that publish
             the total instead of per vehicle type counts. Absent if the feed
-            publishes neither.
+            does not publish num_docks_available.
         stationArea:
           $ref: '#/components/schemas/MultiPolygon'
         bbox:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -5083,6 +5083,17 @@ components:
           description: List of vehicle docks currently available at this station (vehicle type ID -> count)
           additionalProperties:
             type: integer
+        capacity:
+          type: integer
+          description: |
+            Number of physical docks at this station. Absent for virtual or
+            free-floating stations that do not have docks.
+        numDocksAvailable:
+          type: integer
+          description: |
+            Total number of free docks at this station, for feeds that publish
+            the total instead of per vehicle type counts. Absent if the feed
+            publishes neither.
         stationArea:
           $ref: '#/components/schemas/MultiPolygon'
         bbox:
--- a/include/motis/gbfs/data.h
+++ b/include/motis/gbfs/data.h
@@ -124,6 +124,9 @@ struct station_information {
   std::string address_{};
   std::string cross_street_{};
   rental_uris rental_uris_{};
+  // Number of physical docks. Absent for virtual/free-floating stations,
+  // where a free dock count of 0 does not mean the station is full.
+  std::optional<unsigned> capacity_{};
 
   std::shared_ptr<tg_geom> station_area_{};
 
@@ -142,6 +145,9 @@ struct station_status {
   unsigned num_vehicles_available_{};
   hash_map<vehicle_type_idx_t, unsigned> vehicle_types_available_{};
   hash_map<vehicle_type_idx_t, unsigned> vehicle_docks_available_{};
+  // Total number of free docks for feeds that publish the total instead of
+  // per vehicle type counts. std::nullopt if the feed publishes neither.
+  std::optional<unsigned> num_docks_available_{};
   bool is_renting_{true};
   bool is_returning_{true};
 };
--- a/include/motis/gbfs/data.h
+++ b/include/motis/gbfs/data.h
@@ -124,8 +124,9 @@ struct station_information {
   std::string address_{};
   std::string cross_street_{};
   rental_uris rental_uris_{};
-  // Number of physical docks. Absent for virtual/free-floating stations,
-  // where a free dock count of 0 does not mean the station is full.
+  // Number of docks or parking spots, if published by the feed. Gates the
+  // num_docks_available fallback: stations that declare no capacity report
+  // 0 free docks without being full.
   std::optional<unsigned> capacity_{};
 
   std::shared_ptr<tg_geom> station_area_{};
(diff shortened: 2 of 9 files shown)


alexis-morain:
Heads-up: CI has never actually run on this PR. Both workflow runs on `9fbc9b9` (`CI` and `GPU Validation`) are sitting at `action_required`, the first-time contributor gate, so there is no green check to look at and the PR shows up as `unstable`.

Could a maintainer approve the workflow runs? Happy to rebase if that helps. Locally the full suite is green, 170/170 on macOS arm64.

I also retitled the PR to reflect what it actually is: `num_docks_available` is conditionally REQUIRED in GBFS, `vehicle_docks_available` is only required for type-restricted docks, so this is a conformance fix rather than a workaround for odd feeds.

And the offer at the bottom of the description still stands: if the `openapi.yaml` and `RentalStation` additions make this heavier to review, I am happy to split them out and keep this PR to the parser and the two return checks.

openapi.yaml
```diff
@@ -5131,6 +5083,17 @@ components:
           description: List of vehicle docks currently available at this station (vehicle type ID -> count)
           additionalProperties:
             type: integer
+        capacity:
+          type: integer
+          description: |
+            Number of physical docks at this station. Absent for virtual or
+            free-floating stations that do not have docks.
```
pablohoch: Not sure if the "absent" part is true. According to the [GBFS 3.0 Spec](https://gbfs.org/documentation/reference/#station_informationjson) for `stations[].capacity`:

> If this is a virtual station defined using the is_virtual_station field, this number represents the total number of vehicles of all types that can be parked at the virtual station.

I would just remove the second sentence.

openapi.yaml
```diff
@@ -5131,6 +5083,17 @@ components:
...
+        numDocksAvailable:
+          type: integer
+          description: |
+            Total number of free docks at this station, for feeds that publish
+            the total instead of per vehicle type counts. Absent if the feed
+            publishes neither.
```
pablohoch: "Absent if the feed publishes neither" sounds like it should be available if the feed publishes only vehicle type counts. Doesn't look like the current code calculates this field from the vehicle type counts in that case.

I suggest either removing that sentence, or calculating `numDocksAvailable` as the sum of `vehicleDocksAvailable` if `vehicle_docks_available` isn't provided in the GBFS feed. I would probably prefer just adjusting the description and leaving the code as is.

(The comment in `data.h` should be adjusted as well - `std::nullopt if the feed publishes neither.`).

pablohoch (approved):
LGTM except for some small documentation issues (see comments).

Could you please run `pnpm update-api` in the `ui` directory as well for the CI ui check?

openapi.yaml
alexis-morain: Fixed in 7b41607, the second sentence is gone and the description is now just `Number of docks or parking spots at this station.`

You are right that the claim was wrong, it came from the feeds I sampled rather than from the spec. It does not affect the guard though: the fallback keys on "declares a capacity", not on "is a docked station". A virtual station that declares a capacity and reports 0 free docks is genuinely full, so declining returns there is the intended behavior. I reworded the `data.h` comment along those lines so the rationale no longer rests on the virtual/docked distinction.

openapi.yaml
alexis-morain: Went with your preferred option in 7b41607: description adjusted, code unchanged. It now reads `Absent if the feed does not publish num_docks_available.`, and the `data.h` comment says the same, plus a note that the value is not derived from `vehicle_docks_available`.

For what it is worth, deriving it from the typed counts would have been lossy with the current parser anyway. `load_station_status` flattens `vehicle_docks_available` per vehicle type, so an entry like `{"vehicle_type_ids": ["bike", "ebike"], "count": 5}` stores 5 under both types, and summing the map would report 10 free docks where the feed means 5.

alexis-morain:
Pushed 7b41607: both documentation issues addressed (replies in the threads), and `pnpm update-api` run in `ui`, which updated `ui/api/openapi/schemas.gen.ts` and `types.gen.ts`. That was exactly what the `ui` check was failing on. Verified locally with pnpm 10 and node 24 as in CI: `pnpm -r build` leaves the tree clean, `pnpm run lint` and `pnpm run check` pass, and `clang-format-21` reports no change on `data.h`.

Note on `gpu-validate`: it also fails on `master` (push of 2026-08-22) and on the other open PRs, so it looks unrelated to this branch.

The new commit put both workflow runs back into `action_required` (first-time contributor gate), so they will need approving again before anything turns green.

felixguendling:
@alexis-morain if you want to communicate with us in the future, please do so with your own words. If I want to talk to AI, I can do so on my own.

PR #1588, opened 2026-09-03
feature: resolve all codes before throwing not_found_exception with unknownCodes array

orhazal:
Hello,
Like discussed last weekend, here is the PR for unknown codes resolution.

`/plan` currently fails fast on the first unknown timetable location ID, so clients fix`fromPlace`/`toPlace`/`via` one at a time and have to parse the ID out of the message string. This PR validates all location IDs of a request up front (also POST offset `stopId`s and `one`/`many` in POST one-to-many) and reports every unknown one at once in a new optional `unknownCodes` array on the `Error` schema : not a breaking change, existing error responses stay identical. 

`/stoptimes` and `/stop` with an unknown `stopId` now return 404 with `unknownCodes` instead of silently ignoring the parameter. **would you be ok with this?**

Requires motis-project/net#35, which lets `net::not_found_exception` carry an optional JSON payload that the query router merges into the error body.

### DIFF
--- a/include/motis/place.h
+++ b/include/motis/place.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <optional>
+#include <string_view>
+#include <vector>
 
 #include "osr/location.h"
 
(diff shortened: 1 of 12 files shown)


orhazal:
image

A test on /plan with unknown from, to and POST offsets

Issue #635, opened 2024-10-31
TODOs for v2 Release


felixguendling:
- i18n web UI: https://github.com/motis-project/motis/issues/531
- docs: https://github.com/motis-project/motis/issues/562
- cleanup API:
  - https://github.com/motis-project/motis/issues/613
  - https://github.com/motis-project/motis/issues/612
  - https://github.com/motis-project/motis/issues/610
  - https://github.com/motis-project/motis/issues/630
- pass through nigiri parameters:
  - via stations
  - max travel time
  - max transfers
  - min transfer time
  - transfer time factor
- monitoring:
  - OTEL + Prometheus
- GBFS
  - #633: parallel network requests + caching
  - #632: support malformed GeoJSON
- osr: deal with no level <-> specified level transitions not marked with ramp/entry/etc.
  - problem: https://github.com/motis-project/motis/issues/626
  - wip solution: https://github.com/motis-project/osr/pull/22

DomeQdev:
@felixguendling any update on trip modifications and added/new/duplicated?

traines-source:
ADDED/NEW are supported, DUPLICATED/REPLACEMENT are work in progress, trip modifications is more long term. Do you have any example feeds that use these?

DomeQdev:
Not really, but I'm planning to implement it for my [Warsaw GTFS-RT](https://cdn.zbiorkom.live/gtfs-rt/warsaw.pb) feed. If you would like, I can provide one random trip modification in this feed.

traines-source:
Ah nice. Any reason you would do that with trip modifications instead of NEW/REPLACEMENT? As far as I recall the strengths of trip modifications were that one could modify many trips at once (i.e. being less complicated for producers, but more complicated for consumers). So I would lean towards limiting ourselves to REPLACEMENT/DUPLICATED in MOTIS for the time being.

Jouca:
Hello @felixguendling,
Do you have any update about DUPLICATED/REPLACEMENT trip modifications?

I need it specifically for IDFM GTFS-RT for RATP lines.

felixguendling:
Why would cancel + new not work in your case?

Jouca:
It does, but based on the [GTFS-RT documentation](https://gtfs.org/fr/documentation/realtime/reference/#enum-schedulerelationship_1), it seems that it would be much better to do it like that.

Because currently, I need to cancel RATP lines due to how they works with realtime data and replace with new trips based on theoretical trips.

[...]

PR #1598, opened 2026-09-11
Implement static timetable hot-reload feature with cron scheduling

Jouca:
MOTIS currently loads its static timetable once at startup and that's it. `update_interval` only refreshes GTFS-RT/SIRI/VDV data, not the static schedule itself. If your static feed gets republished (in my case IDFM, which changes its `trip_id` scheme on every export), the only way to pick it up is a full restart, and until you do, real-time matching for that dataset quietly stops working too, since the RT feed's `trip_id`s no longer resolve against whatever's loaded. I hit this in production and got tired of restarting the process by hand.

This adds a `reload_cron` option per dataset: on schedule, it (optionally) re-downloads the feed, rebuilds the timetable, and swaps it into the running server, no restart, no dropped requests. There's already a TODO for something like this (#635, "incremental timetable updates"), but nobody had picked it up. I went with a full rebuild rather than an incremental merge, mainly because it lets me reuse the existing `import` pipeline as-is instead of writing a second code path for partial updates.

The actual swap happens above the endpoint layer, not inside it. Building `data`/`motis_instance` is unchanged, it's the same thing `motis import` + startup already do. A small dispatcher in `server.cc` hands each request a reference to whatever generation was current when it came in, so a request that's mid-flight when a reload finishes just keeps running against the timetable it started with. This is basically the same pattern already used for `data::rt_` (`shared_ptr` + `atomic_store`/`atomic_exchange`), just moved up a level so it covers the whole generation, static timetable, osr/adr indices, matches, shapes, etc. instead of only the RT layer. No endpoint code needed to change.

### Config keys added to `timetable.datasets.<tag>`

| Key | Type | Required | Default | Description |
|-----|------|----------|---------|-------------|
| `url` | `string` | no | – | If set, the dataset's static feed is (re-)downloaded from this URL before every scheduled reload. Downloaded to a temp file, then atomically renamed over `path`, so a failed download never touches the previously working file |
| `download_headers` | `map[string,string]` | no | – | Optional HTTP headers sent with the `url` download (e.g. `Authorization`) |
| `reload_cron` | `string` | no¹ | – | croncpp 6-field expression (`sec min hour dom month dow`, e.g. `"0 30 3 * * *"` = daily at 03:30) controlling when this dataset is reloaded. Validated at config load; an invalid expression fails startup rather than silently never firing |

¹ You can set `reload_cron` without `url` too, it'll just re-read whatever's currently at `path`, which is handy if something else (an external cron job, say) already refreshes the file for you.

### What happens on a reload

1. If `url` is set, download to a temp file and rename it over `path` atomically.
2. Rebuild the whole static timetable through the existing `import` pipeline. Since nigiri merges every configured dataset into a single `nigiri::timetable`, changing one means rebuilding all of them, but `import` already skips datasets whose files haven't changed, so this isn't as wasteful as it sounds.
3. Once the rebuild actually succeeds, build a fresh `data` + `motis_instance` from the new artifacts and swap it in. Anything already in flight keeps running against the old generation until it's done; new requests get the new one. The old generation's RT thread gets stopped and the whole thing is freed once the last request using it finishes.
4. Real-time updates for that dataset pick back up on the next `update_interval` tick after the swap.

If anything fails along the way, download, parsing, whatever it just logs the error and keeps serving the old timetable, then retries at the next scheduled time (with at least 5 minutes between attempts, so a broken feed doesn't get hammered every minute).

One thing worth knowing: you'll briefly have two full generations of the static timetable in memory while old requests drain, so expect a temporary bump roughly the size of one dataset's runtime footprint. Scheduling `reload_cron` for a quiet hour is a good idea for that reason.

### Example

```yaml
timetable:
  datasets:
    idfm:
      path: idfm.gtfs.zip
      url: https://.../idfm-gtfs.zip
      download_headers:
        Authorization: MY_API_KEY
      reload_cron: "0 30 3 * * *"   # daily at 03:30
      rt:
        - url: https://.../gtfs-rt-trip-updates
        - url: https://.../gtfs-rt-alerts
```

### Testing

- Added unit tests for the new config fields and for the cron due-check logic on its own (no network/import involved).
- Ran the full `motis-test` suite under the project's `docker-cpp-build` image, including `linux-sanitizer` (ASan + UBSan) clean, nothing flagged.
- Manually ran `motis server` against a fixture with two versions of a feed (different `trip_id` scheme between them, same idea as the IDFM case) and confirmed the reload picks it up with the same PID and no failed requests. Also ran it against a real dataset (IDFM, ~55k stops / 400k trips) end to end download, rebuild, swap, live GTFS-RT trip updates and alerts, all working.


traines-source:
Our approach to that is have a cronjob that calls:

```
wget -N https://example.com/your/feed.zip
docker compose up motis-import
docker restart motis-server
```

With e.g. the following docker compose config:
```
motis-import:
    image: ghcr.io/motis-project/motis:latest
    container_name: motis-import
    volumes:
      - ./feed.zip:/input/feed.zip
      - ./config.yml:/config.yml
      - motis-data:/data:rw
    command: /bin/sh -c "./motis import"
  motis-server:
    image: ghcr.io/motis-project/motis:latest
    container_name: motis-server
    restart: always
    volumes:
      - ./LICENSE.txt:/ui/LICENSE.txt
      - motis-data:/data:rw
    depends_on:
      - motis-import
    ports:
      - 8080:8080
    command: /bin/sh -c "cd / && /motis server"
```

This is not perfect since it may lead to issues when the data directory of a running instance is replaced under it. An even better approach (for small deployments) is to do the import at image build time and build the data directory into the image. Then you can just start a new container based on a new image whenever you have an update (even with rolling upgrades etc).
On Transitous we do the same without Docker and instead a bit of copying around.

In order to ensure gapless realtime, we finally want to implement this idea: https://github.com/public-transport/transitous/issues/1540#issuecomment-3637402734 Because if you don't know when the feeds are switched, you will want to load the new static feed before the realtime feed switches over to it, but still match with the old realtime feed at the same time.

So currently I don't really see what your approach would solve that can't be solved by an external cronjob, even more so with the "perpetual RT" plan. I think in the future it would be helpful to discuss bigger things together before opening a PR.

Jouca:
Thanks for the context, especially the pointer to the perpetual RT plan, I hadn't seen that comment. Annotating the new static version with the previous version's IDs is a cleaner fix for the actual problem (RT matching across a feed transition) than anything a reload mechanism on its own can do, since it doesn't depend on guessing when the RT feed itself switches over.

Fair point on discussing bigger changes before opening a PR, I'll do that going forward. I'll close this one rather than have you review something that's likely to be superseded.

**One thing I'll leave here in case it's useful later:** the atomic swap this PR uses to replace `data`/`motis_instance` without dropping in-flight requests (a small dispatcher in `server.cc` pins each request to whatever generation was current when it arrived) is decoupled from the reload trigger itself. If the annotate-with-previous-IDs work ends up needing a way to apply a rebuilt timetable to a running server without a restart, that part of this PR should still be usable as-is. Happy to help with that piece if it's useful once the design is further along.

PR #1605, opened 2026-09-18
ticketing error on joined trips

traines-source:


src/journey_to_response.cc
```diff
@@ -223,21 +260,31 @@ std::optional<api::TicketUrls> get_ticketing_urls(
...
 
-    auto const from =
-        from_id.has_value() ? std::string{*from_id} : seq_num(enter_stop);
+    auto const from = from_id.has_value()
+                          ? std::string{*from_id}
+                          : seq_num(enter_stop, n::event_type::kDep);
 
-    auto const to =
-        to_id.has_value() ? std::string{*to_id} : seq_num(exit_stop);
+    auto const to = to_id.has_value() ? std::string{*to_id}
```
felixguendling, line 286: `value_or` (same for `from`)

src/journey_to_response.cc
```diff
@@ -223,21 +260,31 @@ std::optional<api::TicketUrls> get_ticketing_urls(
...
     auto const to_id = location_ticketing_id(exit_stop);
 
-    auto const seq_nums = nigiri::loader::gtfs::stop_seq_number_range{
-        {tt.trip_stop_seq_numbers_[enter_stop.get_trip_idx(
-            n::event_type::kDep)]},
-        static_cast<nigiri::stop_idx_t>(trip_range.size())};
-    auto const seq_num = [&](n::rt::run_stop const& s) {
+    auto const seq_num = [&](n::rt::run_stop const& s,
+                             n::event_type const ev) -> std::string {
+      auto const trip = s.get_trip_idx(ev);
```
felixguendling, line 265: this will always throw for unscheduled trips?

will this then crash the trip response in those cases?

I don't really understand what was the problem and how this is fixed now

https://github.com/motis-project/nigiri/blob/8268521628ff9dc56378fd26fd9706f25681c1db/src/rt/frun.cc#L266-L267

traines-source: I figured that `!enter_stop.fr_->is_scheduled()` above should guard against that.

the problems:
1) our favorite problem, using `kDep` unconditionally which breaks for stay-seated joined trips
2) to my understanding we cannot use `trip_range` which is the representation of the leg. In order to calculate the sequence number we need to calculate the offset relative to the start of the trip, not the leg. For that we need to find the trip range in the transport. Then the + overload hopefully handles finding the correct seq_number as before. But it is all very confusing

https://github.com/motis-project/motis/pull/1605/changes#diff-7cb523c340dadef5b7be19d5b39224e3a7ab89440f3f75faf1cc6be20f119575R277-R279

traines-source: Mmh although I think there is still something wrong: When the leg spans two stay-seated trips, we will only make the first trip id part of the URL and the seq_number of the exit stop then silently refers to the second trip, which will likely confuse the provider website. Not sure whether it's worth to fix that (in particular because that is only the case for the legacy mode `joinInterlinedLegs=true` I think)

felixguendling: > legacy mode joinInterlinedLegs=true I think

Which is also the default mode?

Then we should bump the API version and make `joinInterlinedLegs=false` the default?

traines-source: Idk it's just not very consequential if this is wrong because I have a feeling that no website supports it anyways, and even if. But we can of course fix it, not sure how much more complicated it will make it

felixguendling: I think there's no way to really fix it on our side because the Google Ticketing URL format does not support stay seated? This case is just not covered. Or did I miss something @jbruechert ?

traines-source: Well it would just be two separate trips (just like with `joinInterlinedLegs=false`)? And all the params are arrays so maybe one can put multiple trips in there idk.

traines-source: For  `joinInterlinedLegs=false` we actually have the opposite problem: I suppose we only show the ticketing link of the first trip in the UI.

src/journey_to_response.cc
traines-source: that would need
```
auto const to_owned = [](std::string_view v) { return std::string{v}; };
auto const from = from_id.transform(to_owned)
                          .or_else([&] {
                            return std::optional{seq_num(enter_stop, n::event_type::kDep)};
                          })
                          .value();
```
at least if we want to keep lazyness. So I'd keep it as it is.

felixguendling (approved):
probably should reference the latest nigiri master

PR #1608, opened 2026-09-20
gtfs-rt service alerts : period fix

### LINKED ISSUE #415: Alerts with a trip_id but no start_date are only attached to today's run
I'm feeding MOTIS the SNCF service alerts from http://vps-b11dbea5.vps.ovh.net/service-alerts. It's the official GTFS-RT feed with the trip ids rewritten to the static GTFS ones (SNCF publishes internal train ids that don't match anything in the GTFS, so the feed is unusable as is). Everything else is untouched.

With it, the alerts look like this:

```
active_period { start: 2026-09-16T15:29Z  end: 2026-09-26T21:59Z }
informed_entity { trip { trip_id: "OCESN68130R1187_R:CRE:FR:Line::fd1e9a05-0b73-4c3d-83bf-77a10cde0240::87192039:87142109:2:623:20261212" } }
...one entity per trip of the route, none of them with a start_date
```

Image

Trip 68130 is a daily bus (Metz -> Lorraine TGV). If I plan it for the day of the feed, the leg on 68130 has the alert. Same query for the next day, still well inside the period: no `alerts` on the leg. Any other day of the period, same thing. Also, on the day where it works the leg comes back with `realTime: true` even though nothing about it changed.

| Feed day (2026-09-17) | Next day (2026-09-18) |
|---|---|
| Image | Image |

From what I can see this is how `handle_alert` works: the trip selector goes through `gtfsrt_resolve_run(today, ...)`, which picks exactly one day (`start_date` if present, otherwise the UTC date of the feed header), materialises that run with `add_rt_transport` and stores the alert on it (`gtfsrt_alert.cc:110-122`, `gtfsrt_resolve_run.h:48`). Runs on other days never get the alert, and if the trip happens not to run on that particular day the entity is dropped as "trip not found". Before I trimmed my feed down to the trips running on the header's day, that was 77% of the trip entities (`alert_total_informed_entities=27117, alert_total_resolve_success=6296, alert_trip_not_found=20821`), all of them trips that run inside their alert's period, just not that day.

My reading of the spec is that a `trip_id` on its own identifies the trip on all its service days and `active_period` is what bounds the alert in time. `start_date` is only there to disambiguate frequency-based trips or trips colliding across days (https://gtfs.org/documentation/realtime/reference/#message-tripdescriptor).

Route, agency and stop selectors are already stored day-independently (`alerts.route_id_`, `agency_`, `location_`) and matched at query time, the trip selector is the only one tied to a day. So one option: when the descriptor has no `start_date`/`start_time`, store the entity against the static `trip_idx` rather than a run, and have `get_alerts` check the run's time against the impact period. `get_alerts` would need the time from the caller, which in MOTIS is just `journey_to_response.cc`. That would also get rid of the artificial rt transport, so `realTime` stays false for an untouched run.

I have a patch doing this, with a test, and can open a PR if you agree with the direction. Related: route/agency/stop alerts aren't filtered by period at query time either. Should the same check apply there?

To reproduce: static feed https://github.com/orhazal/sncf-gtfs-toolkit/releases/latest/download/sncf_patched.zip (or official one), alerts http://vps-b11dbea5.vps.ovh.net/service-alerts. nigiri 7cd333c, motis 8d295707.

orhazal:
Fixes https://github.com/motis-project/nigiri/issues/415
See https://github.com/motis-project/nigiri/pull/418


orhazal:
i updated the UI so we can have an alert icon for intermediate stops too.
I noticed that a lot of SNCF service alerts reference intermediate stop entities, so it would be a shame not to surface that information in the UI.

image

